PR #21, opened 2023-07-21
initial storage examples attempt

logan-markewich:
Added a test for the storage context, as well as an initial (broken) example that tries to load an index


yisding (approved):
Looks good. Thanks for doing this.

PR #22, opened 2023-07-24
fix persistance, align fromDocuments

logan-markewich:
As per the title, this PR does a few things
- ensures the indexStore is persisted
- ensures the proper data is persisted in the indexStore
- aligns fromDocuments across both indexes

I updated the tests, docs, and examples. You may need to update the GUI example.

Also, maybe double check that this isn't too janky  😅 was tricky to make this work


packages/core/src/indices/vectorStore/VectorStoreIndex.ts
```diff
@@ -50,15 +50,27 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
...
+    const indexStore = storageContext.indexStore;
+    
+    // Setup IndexStruct from storage
+    let indexStructs = await indexStore.getIndexStructs() as IndexDict[];
+    let indexStruct: IndexDict | null;
+    if (indexStructs.length == 1) {
+      indexStruct = indexStructs[0];
+    } else if (indexStructs.length > 1 && options.indexID) {
+      indexStruct = await indexStore.getIndexStruct(options.indexID) as IndexDict;
+    } else if (options.indexStruct) {
```
yisding, line 62: OK so if I look at the Python code, load_index_from_storage is outside the default index initialization code path.

I think having it like this would work, although we probably want to make it clear this is what's going on.

For example, here if someone passes in an indexStruct in the initialization options it will be ignored if the store already has an indexStruct from persistence, which feels a bit not right.

Maybe we should at least move this else if option to the top, before we even check the store for existing index structs?

packages/core/src/indices/list/ListIndex.ts
```diff
@@ -30,6 +30,7 @@ export enum ListRetrieverMode {
 export interface ListIndexOptions {
   nodes?: BaseNode[];
   indexStruct?: IndexList;
+  indexID?: string;
```
yisding, line 33: Small thing, but we camelcase this as indexId in other parts of the code. There's no incorrect option: i've written both Id and ID in code before and I've seen it both ways in APIs, but since we're using Id in the rest of the code base let's keep it as Id.

yisding (commented):
Thanks for doing this Logan! Left a couple of small comments. If you have time we can make the changes in this PR or we can make a new PR for them.

yisding (approved):

PR #33, opened 2023-07-26
openai v4

yisding:


packages/core/src/callbacks/utility/handleOpenAIStream.ts
```diff
@@ -1,64 +1,36 @@
...
     type: "llmPredict",
   });
-  const stream = __astreamCompletion(response.data as any);
   let index = 0;
   let cumulativeText = "";
-  for await (const message of stream) {
-    const token: StreamToken = JSON.parse(message);
-    const { content = "", role = "assistant" } = token?.choices[0]?.delta ?? {};
+  for await (const part of response) {
+    const { content = "", role = "assistant" } = part.choices[0].delta;
```
logan-markewich, line 23: Nit: Shouldn't we be parsing the role out of the response? It could be a function or assistant no? 👀

logan-markewich (approved):
Looks good to me, nice to clean up that extra code!

packages/core/src/callbacks/utility/handleOpenAIStream.ts
yisding: Yeah good point. Right now I think it always returns assistant (even for function calls, where it just uses the function_call payload) but I might as well not hard code assistant in places and just use the OpenAI role instead.

yisding: FYI, this particular line is just like a default option here, so it'll still destructure the role from .delta if it exists.

PR #34, opened 2023-07-26
set typedoc revision to main

yisding:
In theory this will stop the API docs thrashing on every build.

Of course, there is a drawback in that if main changes but we haven't released a new version of the docs yet the links will go out of date.

So longer term we might want to investigate some kind of variable where we can continue to have up to date revs but keep the rev in a single variable somewhere:

https://github.com/facebook/docusaurus/issues/395

In the meantime much smaller commits will be a relief

### DIFF
--- a/apps/docs/docs/api/classes/BaseEmbedding.md
+++ b/apps/docs/docs/api/classes/BaseEmbedding.md
@@ -36,7 +36,7 @@ custom_edit_url: null
 
 #### Defined in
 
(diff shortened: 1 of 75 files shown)


yisding:
(not so) hilarious thing: I just checked the deployments and for all the thrashing we were doing to the git commits the links weren't even making it into the final API docs. Even though they're in the .md files something strips the github links out at the end.

Not going to investigate because it's probably for the best anyways.

yisding:
Going to merge now (huge QoL issue)

PR #47, opened 2023-07-28
add meta delimitation strategy for llama2

yisding:


logan-markewich (approved):
This looks good to me! Did it improve results for you?

yisding:
> This looks good to me! Did it improve results for you?

I think it's better, although also a bit more verbose. Planning to add it to playground so we can really see what kind of difference it makes.

PR #54, opened 2023-07-31
make VectorStore function signatures unique, export VectorStore types

tyre:
**Changes**
+ Make VectorStore interface function signatures `async` to support API-based stores, whose operations are all async
+ export VectorStore types so they can be implemented by custom integrations

Ran lint and all tests pass.


yisding (approved):
Looks great. Thank you!

PR #55, opened 2023-08-01
Repo maintenance

yisding:


logan-markewich:
I'm guessing we are removing the docs here, but they are still generated for each deploy?

logan-markewich (approved):

yisding:
> ~We are removing the api docs here, but I'm guessing they are still generated for each deploy?~
> 
> They are! Nice!

I didn't mean to actually...

For some reason pnpm run build isn't generating the docs/api folder

I'm checking it back in.

yisding:
I'm not 100% sure what happened, tbh. I have to periodically remove that folder because otherwise some old classes that no longer exist will linger in the docs. But it should have been regenerated when I run pnpm run build. Will need to do some research on the right workflow on this. Thanks for catching it!

PR #57, opened 2023-08-01
add a filter type to exact match filters

tyre:
Since types only exist at compile time, it's very difficult to switch off of an `interface` without a key.

This makes it much easier to `switch` over `MetadataFilter`s and map to other structures, like Pinecone filters.


yisding (approved):
Makes sense. Thanks!

PR #58, opened 2023-08-02
add constructor to index node so it keeps passed-in indexId

tyre:
Without its own constructor, one is created that calls `super` and then sets `indexId` to its default value (`""`). That ends up blanking out any passed-in value :(


yisding (approved):
Looks good. Thanks for the bug fix.

Issue #68, opened 2023-08-14
Sweep: OpenAIEmbedding Fails to Retrieve Session from Parameters When Not Using Azure


swk777:
### Details
OpenAIEmbedding Fails to Retrieve Session from Parameters When Not Using Azure
![SCR-20230814-j4y](https://github.com/run-llama/LlamaIndexTS/assets/1302797/6aab74d2-a804-4f71-a7d4-fb2c9384e921)

<details>
<summary>Checklist</summary>

- [X] `packages/core/src/Embedding.ts`
> * In the OpenAIEmbedding class constructor, add a condition to check if a session is provided in the parameters when not using Azure.
> • If a session is provided, use that session instead of creating a new one.
> • If a session is not provided, continue with the existing logic of creating a new session.

- [X] `apps/docs/docs/api/classes/OpenAIEmbedding.md`
> * Update the documentation for the OpenAIEmbedding class constructor to mention that it now retrieves the session from the parameters even when not using Azure.
> • Specify that if a session is not provided in the parameters, a new session will be created.

</details>

PR #70, opened 2023-08-14
Fix OpenAIEmbedding Fails to Retrieve Session from Parameters When No…

swk777:
fixes: #68


yisding (approved):
Looks good, thank you!

PR #80, opened 2023-08-16
feat: add CSV loader

swk777:
use [papaparse](https://github.com/mholt/PapaParse) for parsing


yisding (approved):
Thank you for doing this! Updated the demo a bit, but code looks good!

PR #92, opened 2023-08-29
support keyword index

swk777:


yisding:
Hi Larry, thanks for doing this! One thing: we can't use rake-js because it's LGPL. Could we switch to a different library? Unfortunately most rake libraries for JS are unpopular, unmaintained, or both. This one looks like it may be relatively maintained: https://www.npmjs.com/package/rake-pos

swk777:
> Hi Larry, thanks for doing this! One thing: we can't use rake-js because it's LGPL. Could we switch to a different library? Unfortunately most rake libraries for JS are unpopular, unmaintained, or both. This one looks like it may be relatively maintained: https://www.npmjs.com/package/rake-pos

I've experimented with rake-pos and encountered an issue related to its dependency on brill, which is ESM only. It's somewhat complex, and I believe resolving this issue isn't worth making more extensive changes.However, I found an alternative library, [rake-modified](https://www.npmjs.com/package/rake-modified?activeTab=readme), which I've tested ok. Check it out.

yisding (approved):
Thank you Larry!

PR #93, opened 2023-08-30
Update README.md for apps example

andfk:
Hi, just a minimal thing.

The example commands should be inverted otherwise the 1º throws an issue due to no modules yet installed.


yisding (approved):
Thanks for the catch!

PR #98, opened 2023-09-04
Add Notion database support to NotionReader

TomPenguin:
## Summary
I/F is backward compatible. The core functionality to be crawled was carved out as a separate library, and the logic to crawl the db was added to that library.

## Feature

- Add Notion DB Suport
- Add method `loadPages`, `toDocuments`
  - To allow users to handle the serialization process
  - They are also used inside `loadData`


TomPenguin:
@yisding 
As I previously informed you, I have added support for Notion's database. Please check it out.

yisding (approved):
Thanks Tom!

PR #99, opened 2023-09-05
Add MongoReader [Needs merge]

kkang2097:
## Summary
We need a simple document reader to handle MongoDB documents. 
- wrote a small class that implements existing interfaces
    - package/core/src/readers/SimpleMongoReader.ts
- wrote a mini-app  to demonstrate how to integrate SimpleMongoReader into QueryEngine
    - apps/simple/mongo.ts

##Feature
- add basic MongoDB support

##To-do list:

- [X] Test logic locally
- [X] Check package runs locally
- [X] Write demo
- [X] Ask for merge


kkang2097:
Okay I think it's worth looking at by now

yisding:
Thanks Elliot! 2 small notes: 1. can you run prettier on the code? I should have it set up to do on commit, but not sure why it doesn't run all the time. Need to look into that. 2. I'm not sure .limit(Infinity) is officially supported by the API: https://stackoverflow.com/questions/43788801/does-mongodb-support-passing-infinity-into-limit 0 looks like it does what you want by default.

kkang2097:
Yup

yisding (approved):
Looks good. Thank you!

PR #100, opened 2023-09-05
feat: added draft for a summarizer of the chat history

marcusschiesser:
Currently, the `ChatEngine` implementations always send all messages to the LLM. If the chat history is too large, this approach quickly exceeds the maximum token size. 

This PR proposes to extract the chat history (the array of `ChatMessage`) into its own `ChatHistory` interface. 
There are two implementations: `SimpleChatHistory` which behaves like the existing chat history and
`SummaryChatHistory` which summarizes the chat history.

>Note: This is just a draft PR to discuss the design changes, `SummaryChatHistory` is currently triggering the summary process each time a new message is added. If the design changes are okay, the next step is to define what are good conditions to trigger the summary process.

### DIFF
--- a/packages/core/src/ChatEngine.ts
+++ b/packages/core/src/ChatEngine.ts
@@ -1,4 +1,7 @@
 import { v4 as uuidv4 } from "uuid";
+import { Event } from "./callbacks/CallbackManager";
+import { ChatHistory, SimpleChatHistory } from "./ChatHistory";
+import { ChatMessage, LLM, OpenAI } from "./llm/LLM";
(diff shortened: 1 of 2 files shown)


yisding:
Thanks for the PR. This looks good. We probably want to call this Memory to align with Python but we can do it in another PR.

yisding (approved):
Thanks!

marcusschiesser:
@yisding Thanks for merging! What about the open TODOs in this PR? Wanna discuss them privately? You can drop me an email mail@marcusschiesser.de

PR #110, opened 2023-09-09
Enhancing Type Safety for metadata

TomPenguin:
I had a little difficulty dealing with `metadata`, so I made some modifications to make it type safe.

```ts
// before
const document = new Document({ text, metadata: { foo: "bar" }});
document.metadata.foo // any

// after
const document = new Document({ text, metadata: { foo: "bar" }});
document.metadata.foo // "bar"
```

### TODO

- [x] Type Implements
- [x] Check build successful

### DIFF
--- a/packages/core/src/Node.ts
+++ b/packages/core/src/Node.ts
@@ -23,19 +23,23 @@ export enum MetadataMode {
   NONE = "NONE",
 }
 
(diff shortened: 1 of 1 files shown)


yisding (approved):
Looks good. Thank you.

PR #112, opened 2023-09-10
Create OutputParser.test.ts [Needs Merge]

kkang2097:
## Issue
SubQuestionOutputParser is not working as expected, writing some tests to check it out.
QuestionGenerator outputs a list of objects in string format, which is unexpected.
In particular:
"[{prompt: "smth", response: "nothing"}]"

## Why we have this issue
LLMs are not guaranteed to return outputs in an exact format, even when prompted to.

## To-dos
- [x] Write tests
- [x] Fix bug
- [x] Confirm the tests pass/Review both tests and code

## Files to change
- src/tests/OutputParser.test.ts
- src/OutputParser.ts

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-pnpm lint
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+pnpm lint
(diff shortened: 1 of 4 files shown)


yisding:
@sourabhdesai @logan-markewich there is a question here about what we should be doing if no subquestion tool (or OpenAI function) is the correct one to select. I think maybe we should handle that explicitly?

kkang2097:
For the parser:  if no subQuestion tool is the right one to select, we still get a String of List[JSON object] representation back (as this is the output from SubQuestion generator).

Solution:
- Doesn't matter if we have ```json specified from PromptTemplate or not, we just need a list of JSON objects wrapped by square brackets.
- Trimming the string before the first "[" and after the last "]" fixes this, is pretty general, and is fast.

kkang2097:
Okay fixed.

yisding (approved):
Looks good. Thank you!

PR #121, opened 2023-09-13
Create Single/Multi Selectors [on hold]

kkang2097:
Porting over Selectors from Python side.

## Feature
Given a number of QueryEngineTools, select 1 or more candidate query engines to query results.
- SingleSelectPrompt and MultiSelectPrompt are the same as the Python side but they are more specific.
  - Handles "no QueryEngines are relevant" case better
  - specified output to be strictly JSON or list of JSON.

##TODO
- [x] Update prompts
- [x] Single selector
- [x] Multi Selector
- [x] Make example in apps/simple


EmanuelCampos:
Closing PR, selectors was already added, thanks for the contribution

PR #129, opened 2023-09-24
Fixed Streaming Support for OpenAI LLM [Needs Review]

kkang2097:
The previous OpenAI LLM streaming capability needed a tune-up.

## Feature
Previously: connect LLM Stream -> Accumulate everything -> Do 1 Callback-> LLM.complete function finishes
Now: connect LLM Stream -> AsyncGenerator yields results iteratively, while updating Callbacks

## Scenarios
- Can we use one LLM for multiple query engines, with and without streaming? **Yes**
- Can you do both simple and complex configs? [4-5 line demo vs. logging + callbacks] **Yes**
## Changes
- Separation of chat and stream_chat, streams just behave too differently from batch responses
- need to upgrade to ES2018 at least, otherwise we can't really support streams
  - as both Thanos and ES2018 like to say: "I am inevitable"
- removed a utility file, it was only a single handler function
- made StreamCallbackResponse require a token, when it used to be optional

## To-do
- [x] Write code
- [x] Write barebones demo
- [x] Write demo with token usage logging (fancy LLM callbacks)
- [x] Make better Generics for StreamToken/Callback interfaces
- [x] Ask for merge


packages/core/src/llm/LLM.ts
```diff
@@ -185,37 +201,90 @@ export class OpenAI implements LLM {
...
+        role: this.mapMessageType(message.role),
+        content: message.content,
+      })),
+      top_p: this.topP,
+      ...this.additionalChatOptions,
+    };
+
+    //Now let's wrap our stream in a callback
+    const onLLMStream = this.callbackManager?.onLLMStream
+      ? this.callbackManager.onLLMStream
```
yisding, line 242: So I'd rather take the onLLMStream out of the callbackManager entirely, or rework it.

What we want is for something like the QueryEngine to be able to return responses async just by using a flag (just like OpenAI or Python)

yisding:
Let's fix the test case and then I'll merge this. But we need to work on making the async functionality working on the QueryEngines and ChatEngines (where most people will be using them)

kkang2097:
Was worried about making the PR too big but I'll add that in as well, was gonna do that next

yisding (approved):
I'm going to merge this and fix up the test case so I can release for the Windows fix.

PR #130, opened 2023-09-26
VectorStore - Add Method "VectorStoreIndex.fromVectorStore" + Prefilters + Pinecone Demo

Einsenhorn:
### Description
This PR adds severals thing at the same time listed below:

- It adds support for a method available in Python "VectorStoreIndex.fromVectorStore" allowing you to do what's below.
  ```ts
  const vectorStoreIndex = await VectorStoreIndex.fromVectorStore(vectorStore, serviceContext);
  ```
  This allows you to have a vector store index without being forced to insert nodes during it's creation.
- It adds support for pre filters (filters on metadata) - only tested with Pinecone!
- It adds a simple showcase using an already existing pinecone index (free tier) (so we can call `VectorStoreIndex.fromVectorStore` with pre-filters).

### Concerns
- I'm neither fluent in LLM nor pinecone, so I might be doing weird stuff!

### Test

Please follow what is described in apps/simple/README.md.
Then, you'll need to add your OpenAI and Pinecone credentials to the scripts (check the beginning of the main for that).
You might need to update the pre-filter in this show case to reflect whatever the metadata you've in your pinecone index.

### DIFF
--- a/apps/simple/package.json
+++ b/apps/simple/package.json
@@ -4,6 +4,7 @@
   "name": "simple",
   "dependencies": {
     "@notionhq/client": "^2.2.12",
+    "@pinecone-database/pinecone": "^1.0.1",
     "commander": "^11.0.0",
     "llamaindex": "workspace:*"
   },
--- a/apps/simple/vectorIndexFromVectorStore.ts
+++ b/apps/simple/vectorIndexFromVectorStore.ts
@@ -1,0 +1,197 @@
+import {
+  OpenAI,
+  ResponseSynthesizer,
+  RetrieverQueryEngine,
+  serviceContextFromDefaults,
+  TextNode,
+  TreeSummarize,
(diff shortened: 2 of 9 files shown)


Einsenhorn:
Hey @kkang2097 , I'd rather keep it in the demo as I'm considering it as a code that people should not use in production. They can copy paste it if they want but it's not "supported" by this library (at least with the current version).

Also, there are already 2 repositories doing that:
- https://github.com/yisding/llamaindex-pinecone
- https://github.com/tyre/llamaindex-pinecone

Let me know what you think :)

kkang2097:
If you fleshed this out it could probably be good for production

yisding (approved):
Thanks, merging.

Einsenhorn:
Thanks a lot @yisding !

PR #131, opened 2023-09-27
ChatEngine streaming [needs merge]

kkang2097:
This is more of an extension to the streaming stuff from v0.0.29

Edit: We can make streaming way more elegant. I'll add some refactoring stuff in later tonight.
Edit 2: Made streaming way better, now the return types auto-set based on streaming flag :)

## Feature
Previously, ChatEngines didn't support streaming. Now, some of them do. 
Note:
For QueryEngines, streaming is a bit more complicated:
- ResponseSynthesizers need to be streamed
- For complex QueryEngines, we'll never have true streaming because the logic goes off in too many directions

## Issues
We have lots of repeated logic being written. There are way too many ChatEngines and QueryEngines. What if you want to use the same engine but don't want to reload your data? Intuitively, we should just be able to switch "QueryStrategies" or something like that.

## Miscellaneous Points
- Engines are heavy, so we should just have a few powerful ones.

Something to think about: Modular QueryEngine, but with different search strategies. One engine, one time data loading, many strategies

### DIFF
--- a/apps/simple/llmStream.ts
+++ b/apps/simple/llmStream.ts
@@ -1,5 +1,5 @@
-import {OpenAI, Anthropic, ChatMessage, SimpleChatEngine } from "llamaindex";
-// import {Anthropic} from '@anthropic-ai/sdk';
+import { ChatMessage, OpenAI, SimpleChatEngine } from "llamaindex";
+import {Anthropic} from "../../packages/core/src/llm/LLM";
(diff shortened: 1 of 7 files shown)


kkang2097:
Okay this is way better, I'm confident this works now

yisding (approved):
Sorry I spent way too much time trying to set up a new GH code review flow. This looks good and am going to merge. Thank you!

PR #140, opened 2023-10-09
Feat: Use tokenizer for chat history summarizer

marcusschiesser:


packages/core/src/ChatHistory.ts
```diff
@@ -47,67 +53,112 @@ export class SimpleChatHistory implements ChatHistory {
...
+    return this.messages.length - 1 - index;
+  }
+
+  private get systemMessages() {
+    // get array of all system messages
+    return this.messages.filter((message) => message.role === "system");
+  }
+
+  private get nonSystemMessages() {
+    // get array of all system messages
```
marcusschiesser, line 147:
```suggestion
    // get array of all non-system messages
```

yisding:
Let's add a metadata to the LLM interface:

```
export interface LLMMetadata {
  contextWindow: number;
  tokenizer: string;
}
```

yisding (approved):
Otherwise looks good.

PR #145, opened 2023-10-13
Feature: Extract ContextGenerator and make HistoryChatEngine pluggable

marcusschiesser:
Prevents context window overflow by including the messages from the ContextGenerator in the token calculation

### DIFF
--- a/packages/core/src/ChatEngine.ts
+++ b/packages/core/src/ChatEngine.ts
@@ -317,15 +317,17 @@ export class ContextChatEngine implements ChatEngine {
 }
 
 /**
- * StatelessChatEngine is a ChatEngine that doesn't keep a state of the chat history.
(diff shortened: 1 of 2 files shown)


packages/core/src/ChatEngine.ts
```diff
@@ -166,29 +166,78 @@ export class CondenseQuestionChatEngine implements ChatEngine {
   }
 }
 
+export interface Context {
+  message: ChatMessage;
+  nodes: NodeWithScore[];
+}
+
+export interface ContextGenerator {
```
yisding, line 174: I'm not a fan of creating a separate class for generating context (in my opinion the response builder/retriever should suffice), but will merge this and push a new snapshot.

yisding (approved):
Thanks for doing this and sorry for the delay in reviewing.

PR #152, opened 2023-10-25
Add SimilarityPostprocessor

TomPenguin:
I used the following as a guide to create SimilarityPostprocessor.

https://github.com/run-llama/llama_index/blob/72edf30c2c5c9765a0e714c6bb29197bdceae048/llama_index/indices/postprocessor/node.py#L62-L92

### DIFF
--- a/apps/simple/vectorIndexCustomize.ts
+++ b/apps/simple/vectorIndexCustomize.ts
@@ -3,6 +3,7 @@ import {
   OpenAI,
   RetrieverQueryEngine,
   serviceContextFromDefaults,
+  SimilarityPostprocessor,
   VectorStoreIndex,
 } from "llamaindex";
 import essay from "./essay";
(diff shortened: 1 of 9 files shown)


apps/simple/vectorIndexCustomize.ts
```diff
@@ -21,8 +22,16 @@ async function main() {
...
+    similarityCutoff: 0.7,
+  });
   // TODO: cannot pass responseSynthesizer into retriever query engine
-  const queryEngine = new RetrieverQueryEngine(retriever);
+  const queryEngine = new RetrieverQueryEngine(
+    retriever,
+    undefined,
+    undefined,
+    [nodePostprocessor],
+  );
```
TomPenguin, line 34: It may be necessary to make the argument an object, but I hesitate because this API is exposed to the outside world and would be a destructive change.

apps/simple/vectorIndexCustomize.ts
```diff
@@ -21,8 +22,16 @@ async function main() {
 
   const retriever = index.asRetriever();
   retriever.similarityTopK = 5;
+  const nodePostprocessor = new SimilarityPostprocessor({
+    similarityCutoff: 0.7,
+  });
   // TODO: cannot pass responseSynthesizer into retriever query engine
```
TomPenguin, line 28: Can this comment be deleted?

yisding: Yes, thanks.

yisding (approved):
This looks good. Thank you @TomPenguin !

PR #153, opened 2023-10-25
Add create-llama CLI tool

marcusschiesser:
Testing is quite simple. 

Build `create-llama`:
```
cd packages/create-llama
pnpm install
pnpm run build
pnpm link --global
```

Then just call `create-llama`


logan-markewich:
Is `poetry shell` supposed to come before `poetry install` ? Just looking at the flask readme

marcusschiesser:
> Is `poetry shell` supposed to come before `poetry install` ? Just looking at the fast-api readme

Hmm, worked when I tested it. From my understanding is 'install' creating a virtual env (if it doesn't exist yet) and installing the dependencies. And 'shell' is just activating the virtual environment.

What do you suggest?

logan-markewich:
Ah neat. I always thought `shell` created the env, and `install` installed into that env, but I guess it works both ways 👍

packages/create-llama/pnpm-lock.yaml
```diff
@@ -0,0 +1,882 @@
+lockfileVersion: '6.0'
```
himself65, line 1: Why there are the second `lock` file, I believe this monorepo should only have one such file

packages/create-llama/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "got": "10.7.0",
+    "picocolors": "1.0.0",
+    "prompts": "2.1.0",
+    "tar": "6.1.15",
+    "update-check": "1.5.4",
+    "validate-npm-package-name": "3.0.0",
+    "terminal-link": "^3.0.0"
+  },
+  "engines": {
+    "node": ">=16.14.0"
```
himself65, line 53: >= 18?

Since

image

packages/create-llama/index.ts
```diff
@@ -0,0 +1,399 @@
...
+    console.log("Aborting installation.");
+    if (reason.command) {
+      console.log(`  ${cyan(reason.command)} has failed.`);
+    } else {
+      console.log(
+        red("Unexpected error. Please report it as a bug:") + "\n",
+        reason,
+      );
+    }
+    console.log();
```
himself65, line 394: unused?

packages/create-llama/helpers/is-folder-empty.ts
```diff
@@ -0,0 +1,62 @@
...
+  const validFiles = [
+    ".DS_Store",
+    ".git",
+    ".gitattributes",
+    ".gitignore",
+    ".gitlab-ci.yml",
+    ".hg",
+    ".hgcheck",
+    ".hgignore",
+    ".idea",
```
himself65, line 16: add `.vscode` ?

packages/create-llama/index.ts
marcusschiesser: you mean console.log? it adds a newline

packages/create-llama/pnpm-lock.yaml
marcusschiesser: good catch, i removed it

packages/create-llama/helpers/is-folder-empty.ts
marcusschiesser: helpers are from https://github.com/vercel/next.js/tree/v13.5.6/packages/create-next-app/helpers so I wouldn't touch them if not necessary. Then we can just copy them with new releases.

packages/create-llama/create-app.ts
```diff
@@ -0,0 +1,109 @@
...
+    console.log("Initialized a git repository.");
+    console.log();
+  }
+
+  console.log(`${green("Success!")} Created ${appName} at ${appPath}`);
+
+  console.log(
+    `Now have a look at the ${terminalLink(
+      "README.md",
+      "file://${appName}/README.md",
```
himself65, line 105:
```suggestion
      `file://${appName}/README.md`,
```

himself65 (changes requested):

yisding (approved):
Thanks for doing this Marcus!

PR #154, opened 2023-10-26
Add HTMLReader, sample app and HTML file

mtutty:
Pretty much what the title says.  Adds HTMLReader to core/src/readers, a sample ts file in apps/simple, and a pretty gnarly HTML file in apps/simple/data.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,11 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[xml]": {
     "editor.defaultFormatter": "redhat.vscode-xml"
+  },
+  "[typescript]": {
+    "editor.defaultFormatter": "vscode.typescript-language-features"
+  },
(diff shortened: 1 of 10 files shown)


.vscode/settings.json
```diff
@@ -4,5 +4,11 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[xml]": {
     "editor.defaultFormatter": "redhat.vscode-xml"
+  },
```
yisding, line 7: Can you revert the changes here? We want to use prettier if possible.

mtutty: Yup, check it now.  That wasn't a conscious choice, but suggested by vscode while I was working in the IDE.  If you're doing linting stuff on commit/push, maybe the .vscode dir should be ignored?

yisding (approved):
OK this look good. Thank you for doing this!

PR #159, opened 2023-10-29
Add local models and support browser

mbrunel:
- [x] llamaindexts running on browser
- [x] pulling model for embeddings and llm from huggingface
- [ ] fix node build (can't import transformers.js in cjs nor in esm)
- [ ] improve code quality, clean up and rebase

There is an example in examples/browser_example

### DIFF
--- a/examples/browser_example/README.md
+++ b/examples/browser_example/README.md
@@ -1,0 +1,18 @@
+# Guide to test in browser
+
+1. clone LlamaIndexTS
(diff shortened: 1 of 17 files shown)


himself65 (changes requested):
We are currently using bunchee different platform support. It will be easier for browser support but this is not our first priority for now.

If you are interested on this feature, fell free to re-open the PR

PR #160, opened 2023-10-30
Add observer/callback feature to SimpleDirectoryReader

mtutty:
Add an observer/callback pattern to the SimpleDirectoryReader.  This callback gives the user information about the start/end of each directory traversal, the import of each file, and any errors encountered.  At each callback, the user can return false to skip the directory, file, or quit importing altogether (in the case of an error report).

I found this functionality very helpful for troubleshooting problems with large document imports, but it will also be useful for managing UX (progress reporting / cancel ability) when the time comes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # dependencies
 node_modules
 .pnp
+.pnpm-store
 .pnp.js
 
 # testing
--- a/apps/simple/directory.ts
+++ b/apps/simple/directory.ts
@@ -1,8 +1,13 @@
 import { SimpleDirectoryReader } from "llamaindex";
 
-function callback(category: string, name: string, status: any, message?: string): boolean {
+function callback(
+  category: string,
+  name: string,
+  status: any,
+  message?: string,
+): boolean {
   console.log(category, name, status, message);
-  if (name.endsWith('.pdf')) {
+  if (name.endsWith(".pdf")) {
     console.log("I DON'T WANT PDF FILES!");
     return false;
   }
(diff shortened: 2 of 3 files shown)


yisding (approved):
Thanks for this @mtutty. I'm going to merge this, although if you have some more time I have the following idea:

Right now, it appears your observer is doing two things: 1. acting as a filtering function. 2. doing something similar to our CallbackManager module, in terms of notifying the user or an external observability module about what's going on.

If your particular application doesn't require the latter, then I would recommend reworking the observer to just do the former. If you require both, I think at some point (once I have more time to work on it) I'm going to make the callbackmanager easier to grok using decorators, so might be worth looking into.

mtutty:
@yisding I really meant for this PR to start the kind of conversation in your last comment.  I took a look at the CallbackManager and wondered if expanding it would be a good idea.  In the end, my PR consisted only of changes that wouldn't affect any existing LITS users.

I'm perfectly happy to take this in whatever direction you'd like, but here are a few of the motivations and constraints I considered - let me know which ones are important, not relevant, or flat-out incorrect.

1. Need visibility into potentially long-running processes, for troubleshooting (dev) and user experience (prod)

2. Cancelling individual steps or the whole process isn't exactly compatible with the Observer pattern.  But having a separate mechanism for observing vs filtering or controlling the process might not be worth the effort/complexity?

3. The CallbackManager is built for a very narrow use case.  I don't see a clear path to generalizing for Observable use cases.

4. Adding to the Reader's ctor is fine, and keeps backward compatibility.  Making it part of the loadData would work for Reader classes, but other classes (storage, LLM query, etc)  don't have that method, making implementation somewhat inconsistent and harder to grok?

Issue #141, opened 2023-10-11
Vercel Edge Runtime does not support WASM modules


juncho154930:
### Details

Need support for Vercel Edge Runtime 

possibly resolved by using this package
https://www.npmjs.com/package/@dqbd/tiktoken#vercel-edge-runtime

```
Error: Missing tiktoken_bg.wasm
    at 79251 (/var/task/.next/server/chunks/670.js:160:47344)
    at __webpack_require__ (/var/task/.next/server/webpack-runtime.js:1:161)
    at 95670 (/var/task/.next/server/chunks/670.js:114:1089)
    at __webpack_require__ (/var/task/.next/server/webpack-runtime.js:1:161)
    at 54703 (/var/task/.next/server/app/api/v1/llamaIndex/route.js:1:1217)
    at __webpack_require__ (/var/task/.next/server/webpack-runtime.js:1:161)
    at __webpack_exec__ (/var/task/.next/server/app/api/v1/llamaIndex/route.js:1:2164)
    at /var/task/.next/server/app/api/v1/llamaIndex/route.js:1:2207
    at __webpack_require__.X (/var/task/.next/server/webpack-runtime.js:1:1735)
    at /var/task/.next/server/app/api/v1/llamaIndex/route.js:1:2177 {
  page: '/api/v1/llamaIndex'
}

```

<details open>
<summary>Checklist</summary>

- [X] ``apps/docs/docs/environments.md`` ✅ Commit [`8b5725b`](https://github.com/run-llama/LlamaIndexTS/commit/8b5725b420e589bb736c4635dc10b8cc2e11c7cf)
- [X] ``README.md`` ✅ Commit [`0577bd2`](https://github.com/run-llama/LlamaIndexTS/commit/0577bd2e4a1a8ee8e322a8a3cbca84a657de93a5)
- [X] ``packages/eslint-config-custom/package.json`` ✅ Commit [`8128b6a`](https://github.com/run-llama/LlamaIndexTS/commit/8128b6ae74f4361ce3a994815934941669ee1a88)
- [X] ``src/api/v1/llamaIndex/route.ts`` ✅ Commit [`4853a74`](https://github.com/run-llama/LlamaIndexTS/commit/4853a74e9f3f676cdf363062c2a12cc9520a599f)
</details>

lili21:
run into same issue.

PR #164, opened 2023-10-30
replace tiktoken with js-tiktoken

v4n:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/141

Replaces Tiktoken with JS-Tiktoken, a Pure JavaScript port. Both packages are provided by same mantainer.


yisding (approved):
Looks good thanks.

PR #170, opened 2023-10-31
Add AssemblyAI integration

Swimburger:
This PR adds multiple document readers that either convert audio files to transcripts as documents, or retrieves existing transcripts as documents.
I added an app and example similar to the Notion ones.

Do I need to add any additional documentation?


yisding (approved):
Thanks so much for doing this! Merging now.

PR #178, opened 2023-11-09
Add support for GPT4 Vision Model

marcusschiesser:
Just for `HistoryChatEngine` (used by `chat-llamaindex`) and base LLM

### DIFF
--- a/packages/core/src/ChatEngine.ts
+++ b/packages/core/src/ChatEngine.ts
@@ -328,6 +328,17 @@ export class ContextChatEngine implements ChatEngine {
   }
 }
 
+export interface MessageContentDetail {
+  type: "text" | "image_url";
+  text: string;
+  image_url: { url: string };
+}
+
+/**
+ * Extended type for the content of a message that allows for multi-modal messages.
+ */
+export type MessageContent = string | MessageContentDetail[];
+
 /**
  * HistoryChatEngine is a ChatEngine that uses a `ChatHistory` object
  * to keeps track of chat's message history.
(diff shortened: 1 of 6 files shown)


packages/core/src/ChatEngine.ts
```diff
@@ -1,12 +1,11 @@
 import { v4 as uuidv4 } from "uuid";
-import { Event } from "./callbacks/CallbackManager";
+import { defaultMultiModalPrompt } from "../dist";
```
yisding, line 2: I think this is a typo. We shouldn't be importing from dist.

packages/core/src/Prompt.ts
```diff
@@ -357,6 +357,21 @@ ${context}
 
 export type ContextSystemPrompt = typeof defaultContextSystemPrompt;
 
+export const defaultMultiModalPrompt = ({
+  originalMessage = "",
+  response = "",
+}) => {
+  return `Answer to the given original message. Base your answer on the given previous response and add details from the given context. The new answer should be twice as long as the previous response.
```
yisding, line 364: I'm not sure I understand why it says twice as long here.

packages/core/src/llm/LLM.ts
```diff
@@ -36,7 +37,7 @@ export type MessageType =
   | "memory";
 
 export interface ChatMessage {
-  content: string;
+  content: any;
```
yisding, line 40: This is fine for now, but let's do some typing around this in a different PR.
Screenshot 2023-11-09 at 3 49 36 PM

yisding (commented):
Have a few questions, but otherwise looks good.

yisding (approved):
Thanks!

PR #180, opened 2023-11-14
fix: use llamaindex version and not create-llama version

marcusschiesser:
use llamaindex version for referencing llamaindex in the generated `package.json` and not the wrong create-llama version (worked before as both versions have been in sync)


yisding (approved):
Thanks!

PR #196, opened 2023-11-15
fix: label for simple chat

marcusschiesser:
example run:
```
✔ What is your project named? … my-app
✔ Which template would you like to use? › Chat with streaming
✔ Which framework would you like to use? › NextJS
✔ Which UI would you like to use? › Just HTML
? Which chat engine would you like to use? › - Use arrow-keys. Return to submit.
❯  ContextChatEngine
   SimpleChatEngine (no data, just chat)
```


yisding (approved):
Thanks for catching this!

PR #197, opened 2023-11-15
Expanding README docs

seldo:
We're pointing people directly to create-llama from our homepage now, so these README docs should be authoriative.

This PR includes a minor change to the Python backend template to explicitly set the LLM, to make it easier for somebody to set it to gpt-4.


yisding (approved):
Looks great, thanks!

Issue #198, opened 2023-11-16
[create-llama] Backend readme doesn't mention need to export env file before running


dantheman0207:
https://github.com/run-llama/LlamaIndexTS/blob/b64716d3f7e41e0b8db7421b7150cafd13026843/packages/create-llama/templates/types/simple/fastapi/README-template.md?plain=1#L15

I had provided my OpenAI API key during setup, and followed the instructions in the README exactly. However it still failed as it doesn't automatically load the `.env` file using dot-env or any other package. If that's the case the instructions should tell you to export this file before running, otherwise someone who follows the instructions exactly will have their first experience be triggering an error and having to debug it despite following the golden path.

dantheman0207:
Fixed by this PR which unfortunately doesn't seem to automatically be linked: https://github.com/run-llama/LlamaIndexTS/pull/199

logan-markewich:
Should be fixed in https://github.com/run-llama/LlamaIndexTS/pull/216

PR #199, opened 2023-11-16
Update backend README Fixes run-llama/LlamaIndexTS#198

dantheman0207:
Describe how to export your API key added during setup before running the backend server (this is required if you provided an OpenAI API key).

### DIFF
--- a/packages/create-llama/templates/types/simple/fastapi/README-template.md
+++ b/packages/create-llama/templates/types/simple/fastapi/README-template.md
@@ -9,9 +9,10 @@ poetry install
 poetry shell
 ```
 
-Second, run the development server:
+Second, export your API key. Then run the development server:
 
 ```
+export $(cat .env)
 python main.py
 ```
 


packages/create-llama/templates/types/simple/fastapi/README-template.md
```diff
@@ -9,9 +9,10 @@ poetry install
 poetry shell
 ```
 
-Second, run the development server:
+Second, export your API key. Then run the development server:
 
 ```
+export $(cat .env)
```
marcusschiesser, line 15: It shouldn't be required as the Python backend is calling:

https://github.com/run-llama/LlamaIndexTS/blob/b64716d3f7e41e0b8db7421b7150cafd13026843/packages/create-llama/templates/types/simple/fastapi/main.py#L9

@dantheman0207 Is there something special in your environment that prevents the .env from being loaded? Maybe you already have another API key set?

dantheman0207: I don't think so, fresh install and set the api key via prompt. I don't believe I did have anything in my env but if I did it would have been the same API key and same variable name. When I added it here it failed but I no longer have the exact error. I then manually exported the env file and it failed with a different error (HTTP 401 as OpenAI needed me to verify my phone number). I updated to a different API key but it didn't update until I manually exported again. Strange as I wondered why it wasn't just loaded with dotenv but I see now it is. If I `unset OPENAI_API_KEY` and put the empty string in my .env it fails. But then when I update the .env to again have the same key it still fails with error "No API key found for OpenAI."

oktoberlin: > Author

I also have the same issue. I did not modify anything after the installation completed except the OPENAI_API_KEY.

EmanuelCampos:
Closing due to inactivity

FastAPI readme template includes the guide to import the ENV

https://github.com/run-llama/LlamaIndexTS/blob/main/packages/create-llama/templates/types/simple/fastapi/README-template.md

Feel free to reopen if it still an issue

Issue #200, opened 2023-11-16
Loading indicator for chatbot


himself65:
### Details

I noticed that tweet https://twitter.com/jerryjliu0/status/1724956092340613281. Found that there is no loading UI when you send the message. This looks weird since llm needs time to process the data.

<details open>
<summary>Checklist</summary>

- [X] Modify `packages/create-llama/templates/types/streaming/nextjs/app/components/chat-section.tsx` ✓ https://github.com/run-llama/LlamaIndexTS/commit/f4d3d818ccebdda5b109f4c0b51898c9ac2ff658
- [X] Check `packages/create-llama/templates/types/streaming/nextjs/app/components/chat-section.tsx` ✗
- [X] Modify `packages/create-llama/package.json` ! No changes made
- [X] Check `packages/create-llama/templates/types/streaming/nextjs/app/components/chat-section.tsx` ✗

![Flowchart](http://24.199.78.105:8082/public/10a6f123993036ae38c1fe880a6714f6869cf54b7c6f55296530feadbf5a0a8a_200_flowchart.svg)
</details>

PR #203, opened 2023-11-17
feat: add loading indicator

himself65:
Before the first character responses by API, a gap in UX shows that the app is running.

Closes: https://github.com/run-llama/LlamaIndexTS/issues/200
Related: https://twitter.com/jerryjliu0/status/1724956092340613281
Related: https://twitter.com/jerryjliu0/status/1725567003501670526

https://github.com/run-llama/LlamaIndexTS/assets/14026360/62f946f4-96be-4397-aca1-2f000f70babc


packages/create-llama/templates/components/ui/shadcn/chat/chat-messages.tsx
```diff
@@ -37,6 +38,13 @@ export default function ChatMessages(
         {props.messages.map((m) => (
           <ChatMessage key={m.id} {...m} />
         ))}
+        {props.isPending && (
+          <div
+            className='flex justify-center items-center pt-10'
+          >
+          <Spinner/>
```
marcusschiesser, line 45: how about just using `lucide-react'?
```
<Loader2 className="h-4 w-4 animate-spin" />
```

marcusschiesser (changes requested):
@himself65 Great idea! Thanks for the PR. Added some comments

packages/create-llama/templates/components/ui/shadcn/chat/chat-messages.tsx
```diff
@@ -23,6 +24,12 @@ export default function ChatMessages(
   const showReload =
     props.reload && !props.isLoading && isLastMessageFromAssistant;
   const showStop = props.stop && props.isLoading;
+  const content = lastMessage?.content ?? ''
+  const deferredContent = useDeferredValue(content)
+  const isPending = props.isLoading ? !isLastMessageFromAssistant : (
+    props.isLoading &&
+    isLastMessageFromAssistant && content === deferredContent
+  )
```
himself65, line 32: I have implemented it in another way.

I just got the idea from this section of the document (https://react.dev/reference/react/useDeferredValue#indicating-that-the-content-is-stale).

packages/create-llama/templates/types/streaming/nextjs/app/components/chat-section.tsx
```diff
@@ -2,8 +2,12 @@
 
 import { useChat } from "ai/react";
 import { ChatInput, ChatMessages } from "./ui/chat";
+import { useCallback, useState } from 'react'
```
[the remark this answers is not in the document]
himself65: Thanks for the feedback, updated

himself65:
Just a question: How do you test the templates? I have to copy/paste from my existing project, and there's no type check (and seems no lint check) for templates in llamaindexts

marcusschiesser:
> Just a question: How do you test the templates? I have to copy/paste from my existing project, and there's no type check (and seems no lint check) for templates in llamaindexts. Feels dangerous

current approach is to use https://github.com/run-llama/LlamaIndexTS/tree/main/packages/create-llama/templates/types/streaming and copy the UI folder

packages/create-llama/templates/components/ui/shadcn/chat/chat-messages.tsx
marcusschiesser: I'll test this with the next release

packages/create-llama/templates/types/streaming/nextjs/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "ai": "^2",
     "llamaindex": "0.0.31",
+    "lucide-react": "^0.291.0",
```
marcusschiesser, line 13: don't commit changes to package.json, they are auto-generated by `create-llama`

packages/create-llama/templates/components/ui/shadcn/chat/chat-messages.tsx
marcusschiesser: @himself65 You might also wanna have a look at https://github.com/marcusschiesser/ui - that's where the components are taken from

packages/create-llama/templates/components/ui/shadcn/chat/chat-messages.tsx
```diff
@@ -23,6 +24,12 @@ export default function ChatMessages(
   const showReload =
     props.reload && !props.isLoading && isLastMessageFromAssistant;
   const showStop = props.stop && props.isLoading;
+  const content = lastMessage?.content ?? "";
+  const deferredContent = useDeferredValue(content);
+  const isPending = props.isLoading ? !isLastMessageFromAssistant : (
+    props.isLoading &&
+    isLastMessageFromAssistant && content === deferredContent
+  );
```
marcusschiesser, line 32: @himself65, this condition is a bit hard to read so a comment could help. E.g., Why do we have `props.isLoading` again in the second clause (when `props.isLoading` is `false`).

himself65: I thought twice and found that I should have made the logic better. updated

himself65 (commented):
Btw, I found that pnpm has a different install logic than others.

For example: `typescript: ^5` will be `5.0.2` in pnpm but `5.3.2` in other package managers. This might cause some bugs in the runtime.

marcusschiesser:
> Btw, I found that pnpm has a different install logic than others.
> 
> For example: `typescript: ^5` will be `5.0.2` in pnpm but `5.3.2` in other package managers. This might cause some bugs in the runtime.

Good catch! I had precisely the same problem yesterday. For that reason, pnpm did not work, but npm did... Would you like to look at this in one of your next PRs and fixate the version numbers?

marcusschiesser (approved):

himself65:
> > Btw, I found that pnpm has a different install logic than others.
> > For example: `typescript: ^5` will be `5.0.2` in pnpm but `5.3.2` in other package managers. This might cause some bugs in the runtime.
> 
> Good catch! I had precisely the same problem yesterday. For that reason, pnpm did not work, but npm did... Would you like to look at this in one of your next PRs and fixate the version numbers?

Yes, I will check that

PR #204, opened 2023-11-17
Feat: added support for MongoDB as vector DB

marcusschiesser:
- compatible with Python 0.9
- added ingestion and query from https://github.com/run-llama/mongodb-demo as example
- updated SimpleMongoReader to have the same features as Python 0.9 version


apps/mongodb/README.md
```diff
@@ -0,0 +1,127 @@
...
+Now you're creating a vector search client for Mongo. In addition to a MongoDB client object, you again tell it what database everything is in. This time you give it the name of the collection where you'll store the vector embeddings, and the name of the vector search index you'll create in the next step.
+
+### Create a vector search index
+
+Now if all has gone well you should be able to log in to the Mongo Atlas UI and see two collections in your database: the original data in `tiny_tweets_collection`, and the vector embeddings in `tiny_tweets_vectors`.
+
+![MongoDB Atlas collections](./docs/3_vectors_in_db.png)
+
+Now it's time to create the vector search index so that you can query the data.
+It's not yet possible to programmatically create a vector search index using the [`createIndex`](https://www.mongodb.com/docs/manual/reference/method/db.collection.createIndex/) function, therefore we have to create one manually in the UI.
```
marcusschiesser, line 83: @yisding creating the index programmatically is not yet possible

apps/mongodb/package.json
```diff
@@ -0,0 +1,17 @@
+{
```
marcusschiesser, line 1: @yisding i saw that there are examples in `apps` and `examples` - where is the better place for this example?

packages/core/src/storage/vectorStore/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  return metadata;
+}
+
+export function metadataDictToNode(metadata: Metadata): BaseNode {
+  const nodeContent = metadata["_node_content"];
+  if (!nodeContent) {
+    throw new Error("Node content not found in metadata.");
+  }
+  const nodeObj = JSON.parse(nodeContent);
```
marcusschiesser, line 47: @yisding it works to reconstruct objects from Python 0.9, but there is an issue that the node attributes in Python are snake_case and in TS they are camelCase, so a reconstructed `TextNode` in TS (coming from Python) looks like this:

```
_TextNode {
      id_: '95d3f69c-4197-41ca-9f87-b78d9892867f',
      metadata: {},
      excludedEmbedMetadataKeys: [],
      excludedLlmMetadataKeys: [],
      relationships: [Object],
      hash: '4JN1LIB89VS6Hq1tZbtoYrJDCN0uLyHs68Et8K4B/ZA=',
      embedding: null,
      excluded_embed_metadata_keys: [],
      excluded_llm_metadata_keys: [],
      text: "@bradleymeck @aredridel @nebrius @addaleax All fair points. We couldn't figure out a way to get around HTTP so we think bundles will still be used as a perf optimization, but getting away without a build step at development time is a worthwhile win, I think.",
      start_char_idx: null,
      end_char_idx: null,
      text_template: '{metadata_str}\n\n{content}',
      metadata_template: '{key}: {value}',
      metadata_seperator: '\n',
      class_name: 'TextNode',
      metadataSeparator: '\n'
    }
    ```
    I think we should either use snake_case in TS too or add a mapping

yisding (approved):
Thanks!

packages/core/src/storage/vectorStore/utils.ts
yisding: Let's create a TS class to track and convert for now. The interface on the python side may also not be entirely stable yet, so let's create one that says TextNodev09 etc.

PR #208, opened 2023-11-17
fix: generated frontend is sending text/plain

seldo:
Express middleware was expecting a JSON payload and getting text/plain, so parse the JSON in the controller instead.

There may be a smarter way to do this.


yisding (approved):
Thanks!

PR #215, opened 2023-11-19
Fix: minor bugs when deploying express frontend/backend to render

seldo:
* .env was not being gitignored so it kept exposing my OpenAI key
* There was no CORS header being set so the frontend couldn't call the backend in prod
* The next.js frontend must be configured differently to be deployed as a totally static site


yisding (approved):
Thanks Laurie!

PR #218, opened 2023-11-20
feat: use cryptojs instead of crypto

marcusschiesser:
- Generated hashes are the same (tested with Node.test.ts)
- Advantage: No node.js dependency
- Disadvantage: Pure JS => slower

Thanks to @mbrunel for providing the basis in #159


yisding (approved):
Thanks! Looks good!

PR #231, opened 2023-11-24
fix(create-llama): lock deps version

himself65:
This PR fixes the pnpm install might lead to the minimum version of each package.

Related: https://github.com/run-llama/LlamaIndexTS/pull/203#issuecomment-1825114528

I realize that you Imitate the code structure from `create-next-app`.  But I think they have some e2e test coverage and I'm working on doing the same thing to ensure the stability of our codebase


marcusschiesser:
> I realize that you Imitate the code structure from `create-next-app`. But I think they have some e2e test coverage and I'm working on doing the same thing to ensure the stability of our codebase

Great idea! I didn't have time to add these e2e tests yet. I think it would be best to add them to this PR; then we ensure that the versions are working.

yisding (approved):
Looks good. Thanks!

PR #233, opened 2023-11-24
Feat: add GPT4 Vision support (and file upload) to create-llama

marcusschiesser:
This PR adds GPT4 Vision support (and file upload) to create-llama

Current issues:
1. We can't show the images in the already sent messages due to https://github.com/vercel/ai/issues/768
2. BLOCKER: Frontend generation doesn't work for Python


packages/create-llama/templates/types.ts
himself65, line 1: `TemplateType` in `packages/create-llama/templates/types.ts` should also update, since you remove the simple example

marcusschiesser: Simple template is only removed for NextJS, so we have to keep that type for now

yisding (approved):
Looks good. Merging.

PR #234, opened 2023-11-24
Feat: Adds multi modal retriever 

marcusschiesser:
LGTM, here's a run of the multi-modal example:

```bash
cd examples/
pnpx ts-node multiModal.ts
...
Text: Vincent Willem van Gogh (Dutch: [ˈvɪnsɛnt ˈʋɪləɱ vɑŋ ˈɣɔx] ; 30 March 1853 – 29 July 1890) was a Dutch Post-Impressionist painte
ID: ef56b772-5195-4dad-8a11-46b83a190c70
Similarity: 0.8810361156948888
Image: /Users/marcus/code/private/LlamaIndexTS/examples/data/multi_modal/62.jpg
ID: data/multi_modal/62.jpg
Similarity: 0.3020945635708776
Image: /Users/marcus/code/private/LlamaIndexTS/examples/data/multi_modal/61.jpg
ID: data/multi_modal/61.jpg
Similarity: 0.29769412544841267
Image: /Users/marcus/code/private/LlamaIndexTS/examples/data/multi_modal/60.jpg
ID: data/multi_modal/60.jpg
Similarity: 0.26017858258006277
```


packages/core/src/indices/vectorStore/VectorStoreIndex.ts
```diff
@@ -21,32 +35,46 @@ import {
...
+  embedModel: BaseEmbedding;
+  imageVectorStore?: VectorStore;
+  imageEmbedModel?: MultiModalEmbedding;
 
   private constructor(init: VectorIndexConstructorProps) {
     super(init);
-    this.vectorStore = init.vectorStore;
+    this.indexStore = init.indexStore;
+    this.vectorStore = init.vectorStore ?? init.storageContext.vectorStore;
+    this.embedModel = init.serviceContext.embedModel;
```
marcusschiesser, line 69: @yisding the old code (and the Python code) are inconsistent in retrieving objects from the service context (e.g. `embedModel`). Sometimes, objects are retrieved at construction time, and sometimes, when a function is called.
We should fix this. Options: 
1. Retrieving at construction time, then we can override a setting in the constructor (like in the line above: we can set another vector store instead of using the one from the storage context).
2. Retrieving at function call time, then we can override a setting if needed in the function call 
As some overrides (e.g. the vector store) don't make sense at function call time, I would opt for 1.

yisding: This is a good point. So I think the vision for service context longer term is for it to go away entirely.

What service context really is is two things: 1. a set of default configurations, and 2. a set of prebuilt objects (for optimization/lack of duplication more than anything).

The problem with it is that, kind of like our classes, it contains state that's not easy to reason with. Often times, the class will have its own state that's somewhat shared with servicecontext, and different classes will have state that's very very similar (and sometimes those classes are also part of servicecontext, which makes things extra fun). And the way we currently do configurations makes sense in some cases, but doesn't make sense in others (for example the temp and prompts are probably things that people want to change more frequently).

That said, there is some necessity to have a set of default configuration. imagine if every time you wanted to use llamaindex to make a single LLM you had to pass back in the url, your key, the system prompt, and everything over again.

So something to think about and to put on our roadmap as to how to handle.

marcusschiesser:
@yisding the failed CodeQL check is from an unchanged file - not sure how to deal with this

yisding (approved):
Awesome, thanks Marcus!

PR #236, opened 2023-11-25
feat: add epub file reader support

m1911star:
since I have too many epub files, I tried to add epub reader

deps

- https://github.com/gaoxiaoliangz/epub-parser


yisding:
Thanks for doing this! Is this ready to merge?

m1911star:
> Thanks for doing this! Is this ready to merge?

nope. I need more tests.

README.md
```diff
@@ -97,7 +97,7 @@ export const runtime = "nodejs"; // default
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   experimental: {
-    serverComponentsExternalPackages: ["pdf-parse"], // Puts pdf-parse in actual NodeJS mode with NextJS App Router
+    serverComponentsExternalPackages: ["pdf-parse", "@gxl/epub-parser"], // Puts pdf-parse and @gxl/epub-parser in actual NodeJS mode with NextJS App Router
```
m1911star, line 100: since I cannot test nextjs app, so I just copy like pdf-parse config here.
Advice needed. thank you

yisding: I don't think you necessarily need to put this here, but I'll give it a try with the NextJS configuration and see what happens.

`create-llama` is probably your best bet to test this. @marcusschiesser thoughts?

marcusschiesser: @m1911star and @yisding 
yes, in `create-llama` we're just using:
```
serverComponentsExternalPackages: ["llamaindex"]
```
This should work here, too, but explicitly enumerating the problematic dependencies (before just "pdf-parse" and now "@gxl/epub-parser") has the advantage that we know who is the culprit and what we have to do.
This brings us to the question: @m1911star, is there another epub parser package that works with nextjs?
There seem to be alternatives: https://www.npmjs.com/search?q=epub-parser

m1911star: I have dived into alternatives of @gxl/epub-parser and did not find the right choice suitable for next app.

Also `@gxl/epub-parser` can not work in next app.

Im working to find another solution. Thanks.

yisding: Thanks @m1911star! Appreciate the contribution!

EmanuelCampos:
looks cool! 

bump @m1911star any update here?

packages/core/package.json
```diff
@@ -4,6 +4,12 @@
   "license": "MIT",
   "type": "module",
   "dependencies": {
+    "@anthropic-ai/sdk": "^0.9.1",
+    "@gxl/epub-parser": "2.0.4",
+    "@notionhq/client": "^2.2.13",
+    "@xenova/transformers": "^2.8.0",
+    "crypto-js": "^4.2.0",
+    "js-tiktoken": "^1.0.8",
```
marcusschiesser, line 12: this should only add the epub parser

marcusschiesser (commented):
Thanks @m1911star looking good. Please have a look at the comments and also run `pnpx changeset` in root to generate a changeset file

packages/core/package.json
m1911star: Yes, I'm still working on this. Sorry for bothering you.
Will change this pr to draft until it's ready.

m1911star:
let's have a try

examples/data/wells.epub
marcusschiesser, line 1: where is this example taken from? what license?

packages/core/src/readers/EpubReader.ts
```diff
@@ -0,0 +1,63 @@
...
+    const sections = book.sections ?? [];
+    const header = `${book.info?.author}\n${book.info?.publisher}\n${book.info?.title}`;
+    const options = this.getOptions();
+    const results = await Promise.all(
+      sections.map((section) => {
+        return new Promise(async (resolve) => {
+          const parsed = await this.parseContent(section.htmlString, options);
+          resolve(parsed);
+        });
+      }),
```
marcusschiesser, line 28: this doesn't work?

```suggestion
      sections.map(async (section) => {
       return await this.parseContent(section.htmlString, options);
        });
      }),
```

packages/core/src/readers/EpubReader.ts
```diff
@@ -0,0 +1,63 @@
...
+  getOptions() {
+    return {
+      skipHtmlDecoding: true,
+      stripTogetherWithTheirContents: [
+        "script", // default
+        "style", // default
+        "xml", // default
+        "head", // <-- custom-added
+      ],
+    };
```
marcusschiesser, line 61: Modifying these options is hard for the user (needs to override the method)

how about passing the options via a constructor to `EpubReader` and using these values above as default? (if the user doesn't pass any options). 

also has `@gxl/epub-parser` a type for the options?

m1911star: the options here is for `string-strip-html`, not parser, so I just make it as the default options and remove getOptions

packages/core/src/readers/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from "./AssemblyAIReader.js";
 export * from "./CSVReader.js";
 export * from "./DocxReader.js";
+export * from "./EpubReader.js";
```
himself65, line 4: I prefer not to export this to root. And be careful with edge runtime

examples/data/wells.epub
himself65, line 1: could we download from somewhere instead of leave it here

examples/epub.ts
```diff
@@ -0,0 +1,18 @@
+import { EpubReader, VectorStoreIndex } from "llamaindex";
+
+async function main() {
+  // Load PDF
+  const reader = new EpubReader();
+  const documents = await reader.loadData("data/wells.epub");
```
marcusschiesser, line 6: @m1911star you can load the ebub from an URL here instead

himself65:
outdated, closing

PR #238, opened 2023-11-26
test: add coverage on create-llama

himself65:
Refs: https://github.com/run-llama/LlamaIndexTS/pull/231#issuecomment-1825240140

This is different with vercel way. They compare the output file to see if it's correct. But we did run the server to see if it's working.

### DIFF
--- a/packages/create-llama/e2e/basic.spec.ts
+++ b/packages/create-llama/e2e/basic.spec.ts
@@ -1,3 +1,4 @@
+/* eslint-disable turbo/no-undeclared-env-vars */
 import { expect, test } from "@playwright/test";
 import { exec } from "child_process";
 import { execSync } from "node:child_process";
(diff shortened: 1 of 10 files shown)


packages/create-llama/e2e/basic.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+test("try create-llama", async ({ page }) => {
+  const createLlama = fileURLToPath(
+    new URL("../dist/index.js", import.meta.url),
+  );
+
+  const name = "basic";
+  const template: TemplateType = "streaming";
+  const framework: TemplateFramework = "nextjs";
+  const engine: TemplateEngine = "simple";
+  const ui: TemplateUI = "shadcn";
```
marcusschiesser, line 32: @himself65 so currently we're testing only this configuration right? How about adding other permutations?

himself65: I will add them later

himself65:
I think this E2E is too costly since all enumeration will be generated. Thinking to improve this later

marcusschiesser:
> I think this E2E is too costly since all enumeration will be generated. Thinking to improve this later
Some ideas:
- These tests only need to run for create-llama - move it to its own repo? (So changes on LITS are not triggering this e2e)
- Run each test permutation in its own process

marcusschiesser:
closing, will open a new one

PR #240, opened 2023-11-27
Minor fixes in PGVectorStore

mtutty:
This commit fixes a mistake in query output mappings in the query method, and adds a specific check for empty embeddingResults in the add method.  This last change isn't a bug per se but in working on another vector store impl, I began to suspect that add is being called with an empty array when creating the service context (e.g., storageContextFromDefaults function).


yisding (approved):
Awesome, thank you.

PR #244, opened 2023-11-28
Handle Azure specific responses, where choices could be 0

parhammmm:
When streaming Azure replies the following as part of a chunk. Notice that there are no `choices`.

```
{
  id: '',
  object: '',
  created: 0,
  model: '',
  prompt_filter_results: [ { prompt_index: 0, content_filter_results: [Object] } ],
  choices: []
}
```

The suggested fix skips over these chunks as so not to cause `TypeError: Cannot set properties of undefined (setting 'index')` error


yisding (approved):
Thanks for the bug fix!

PR #246, opened 2023-11-29
Unify apps/simple and examples - move everything to examples

marcusschiesser:
After this PR, all `simple` examples are in the `examples/` folder and can be used with the local build and the published package. 
I made a diff between `apps/simple` and `examples` to unify the examples.


examples/pg-vector-store/load-docs.ts
marcusschiesser, line 1: @yisding the pg-vector-store example is not working as we don't the store in `llamaindex`. Can be fixed in a separate PR

pnpm-lock.yaml
```diff
@@ -135,15 +135,15 @@ importers:
         specifier: ^11.1.0
         version: 11.1.0
       llamaindex:
-        specifier: workspace:*
-        version: link:../../packages/core
+        specifier: latest
+        version: link:../packages/core
```
marcusschiesser, line 139: @yisding this is pretty cool: If we would keep this line, the `examples` package would always use the local version (when using `pnpm`).
As we might overwrite this by accident, I added the `pnpm link` command to the readme anyhow

examples/pg-vector-store/load-docs.ts
yisding: Thanks for the catch!

examples/README.md
```diff
@@ -1,9 +1,29 @@
...
 ```
+
+Then run the examples with `ts-node`, for example `npx ts-node vectorIndex.ts`
+
+## Using the local build
+
+```bash
+pnpm install
+pnpm --filter llamaindex build
+pnpm link ../packages/core
```
yisding, line 26: Nice!

yisding (approved):
Awesome, thanks for doing this Marcus!

PR #257, opened 2023-12-04
Feat: Add support for installing templates from community repo

marcusschiesser:
- Added option to install templates from https://github.com/run-llama/create_llama_projects
- Simplified questions logic and used strong-typing for question args

To test:
```
pnpm build --filter create-llama
./packages/create-llama/dist/index.js 
```


yisding (approved):
Looks good. Thanks!

PR #259, opened 2023-12-05
Fixes for create-llama release

marcusschiesser:
- Use shadcn UI as default for NextJS template (simplifies development)
- Added minor/patch versions to be reproducible between using pnpm and npm
- Lowered `maxTokens` to 2048. We can't leave this value to `undefined` as before because of the vision model


yisding (approved):
This looks good. Thank you Marcus!

PR #260, opened 2023-12-05
[WIP] French documentation

hexapode:
Automatic translation of documentation in French

Todo
[] Review all page to check for bad french


yisding (approved):
I'm going to merge this.

Longer term we're going to need a way to re-generate these and also preserve any manual edits.

PR #266, opened 2023-12-07
Feat: Added JSON extract example

marcusschiesser:
No changes to LITS are needed. Just an example of how to use it.


yisding (approved):
Love it, thank you!

PR #270, opened 2023-12-09
Add a .usage property on LLM object

hexapode:
Add a .usage property on LLM object allowing tracking of token/sec consuption and Amount accross all call for OpenAI and Anthropic.

Also update the LLM module doc to reflect the change

### DIFF
--- a/packages/core/src/llm/LLM.ts
+++ b/packages/core/src/llm/LLM.ts
@@ -46,7 +46,7 @@ export interface ChatResponse {
   raw?: Record<string, any>;
   delta?: string;
   metrics?: any;
-  usage?: any;
+  usage?: Usage;
 }
 
 // NOTE in case we need CompletionResponse to diverge from ChatResponse in the future
--- a/packages/core/src/llm/LLM.ts
+++ b/packages/core/src/llm/LLM.ts
@@ -98,19 +98,44 @@ export interface LLM {
    * Calculates the number of tokens needed for the given chat messages
    */
   tokens(messages: ChatMessage[]): number;
+
+  /**
+   * Returns the usage information of the LLM
+   */
+  usage: Usage;
 }
 
 export const GPT4_MODELS = {
-  "gpt-4": { contextWindow: 8192 },
-  "gpt-4-32k": { contextWindow: 32768 },
-  "gpt-4-1106-preview": { contextWindow: 128000 },
-  "gpt-4-vision-preview": { contextWindow: 8192 },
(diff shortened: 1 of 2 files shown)


marcusschiesser (commented):
@hexapode Great idea, but how do we keep this in sync with the price changes?

packages/core/src/llm/LLM.ts
```diff
@@ -45,6 +45,8 @@ export interface ChatResponse {
   message: ChatMessage;
   raw?: Record<string, any>;
   delta?: string;
+  metrics?: any;
+  usage?: any;
```
himself65, line 49: don't use `any` if possible

hexapode: Yes, changed it for usage.

hexapode:
> @hexapode Great idea, but how do we keep this in sync with the price changes?

@marcusschiesser: We have to manually check? No better idea for now / no provider expose a pricing API?

marcusschiesser:
> > @hexapode Great idea, but how do we keep this in sync with the price changes?
> 
> @marcusschiesser: We have to manually check? No better idea for now / no provider expose a pricing API?

@hexapode I think for a start we could just track the usage (i.e. tokens and compute seconds) and not the costs. That's still valuable and if needed a user could calculate the costs - e.g. you could add an example computing the costs in `examples/`

EmanuelCampos:
Agreed, we can just return the total tokens used, the price isn't trustable and we can't let users think that's the current real price

bump @hexapode if you have any plans for it, when I have free time can go with it as well

marcusschiesser:
TODOS:

1. remove `cost` from usage (as the price is not stable)
2. add example (same as in docs)

himself65:
> I will try to use `AsyncLocalStorage` on this

I think the ideal API would be

```ts
const usage = new Usage();
const llm = new OpenAI();
usage.track(llm)
// ...
console.log(usage.cost) // in dollar
```

marcusschiesser:
@himself65, your API looks great!

We probably don't need to support reusing `Usage` objects between LLMs, so you can even simplify it to:
```
const llm = new OpenAI();
const openAIUsage = new Usage(llm);
// ...
console.log(usage.tokens) 

{
	const newLLM = new AnotherAI();
        const anotherAIUsage = new Usage(newLLM);
}
```
Btw: I don't think we should start with the costs, see https://github.com/run-llama/LlamaIndexTS/pull/270#issuecomment-1867576025

himself65:
closing as you should implement this in user land

PR #277, opened 2023-12-13
Update Astra client and associated type changes

mfortman11:
- Updates `@datastax/astra-db-ts` to `0.1.2`
- Connect to AstraDB with `ASTRA_DB_ENDPOINT` instead of DB ID and region
- Align `create` collection functions param `options` to be of type `CreateCollectionOptions`
- Update Readme and examples accordingly


marcusschiesser (approved):
thanks @mfortman11 looking good!

PR #280, opened 2023-12-15
Feat: Add multi modal RAG

marcusschiesser:

PR #281, opened 2023-12-15
Feat: store images in docstore

marcusschiesser:
In #280 the `VectorIndexRetriever` is retrieving image documents from the index and reconstructing their URL based on their ID (see https://github.com/run-llama/LlamaIndexTS/pull/280/files#r1427834636).

Suggestion:
1. Store the images in the doc store (This PR stores images of `ImageNode`s base64 encoded in the doc store)
2. Use the doc store to get the images in the `VectorIndexRetriever`

What are your thoughts?

### DIFF
--- a/packages/core/src/indices/vectorStore/VectorStoreIndex.ts
+++ b/packages/core/src/indices/vectorStore/VectorStoreIndex.ts
@@ -4,7 +4,6 @@ import {
   ImageNode,
   MetadataMode,
   ObjectType,
-  jsonToNode,
   splitNodesByType,
 } from "../../Node";
 import { BaseQueryEngine, RetrieverQueryEngine } from "../../QueryEngine";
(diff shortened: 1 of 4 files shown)


packages/core/src/indices/vectorStore/VectorStoreIndex.ts
```diff
@@ -276,9 +275,11 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
       if (
         !vectorStore.storesText ||
         type === ObjectType.INDEX ||
-        type === ObjectType.IMAGE
+        type === ObjectType.IMAGE ||
+        type === ObjectType.IMAGE_DOCUMENT
```
logan-markewich, line 279: Just curious, what's the difference between IMAGE and IMAGE_DOCUMENT ? Is IMAGE a node?

marcusschiesser: Yes exactly and `IMAGE_DOCUMENT` is a document containing an image. The image loader generates type `IMAGE_DOCUMENT`, so I had to make this change here

PR #284, opened 2023-12-18
feat: add /api/chat e2e test (uses openai key)

thucpn:


marcusschiesser (approved):

marcusschiesser (approved):

packages/create-llama/e2e/utils.ts
```diff
@@ -96,7 +96,7 @@ export function runCreateLlama(
     "--model",
     MODEL,
     "--open-ai-key",
-    "testKey",
+    process.env.OPENAI_API_KEY || "testKey",
```
marcusschiesser, line 99: @thucpn 

In the job, we're still getting `testKey` (https://github.com/run-llama/LlamaIndexTS/actions/runs/7246220748/job/19737708679) - might it be that `test` is starting a new process and this child process doesn't have the right `process.env`?

marcusschiesser: not the case as it works here: https://github.com/thucpn/LlamaIndexTS/actions/runs/7245927631/job/19736878992

.github/workflows/e2e.yml
```diff
@@ -36,6 +36,8 @@ jobs:
         working-directory: ./packages/create-llama
       - name: Run Playwright tests
         run: pnpm exec playwright test
+        env: 
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
marcusschiesser, line 40: @yisding please check if `OPENAI_API_KEY` is set correctly, as e2e tests are not running

yisding: It says it was set 4 days ago, but I tried adding it again (with a different key)

PR #285, opened 2023-12-18
feat: integrate postgres datasource

thucpn:


packages/create-llama/templates/index.ts
```diff
@@ -31,12 +31,18 @@ const createEnvLocalFile = async (
...
       content += `MONGODB_URI=\n`;
       content += `MONGODB_DATABASE=\n`;
       content += `MONGODB_VECTORS=\n`;
       content += `MONGODB_VECTOR_INDEX=\n`;
       break;
     }
+    case "pg": {
+      content += `# For generating a connection URI, see https://www.postgresql.org/docs/16/libpq-connect.html#LIBPQ-CONNSTRING-URIS\n`;
+      content += `PG_CONNECTION_STRING=\n`;
+      break;
```
marcusschiesser, line 44: similar to mongodb, we should be able to parameterize schema and table used to store the embeddings.
Looking at `PGVectorStore`, we need to update also that class.

packages/create-llama/templates/components/vectordbs/pg/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  serviceContextFromDefaults,
+} from "llamaindex";
+import pg from "pg";
+import { CHUNK_OVERLAP, CHUNK_SIZE, checkRequiredEnvVars } from "./shared.mjs";
+
+async function getDataSource(llm: LLM) {
+  checkRequiredEnvVars();
+  const pgvs = new PGVectorStore();
+  pgvs.db = new pg.Client({
+    connectionString: process.env.PG_CONNECTION_STRING,
```
marcusschiesser, line 16: Did work with neither:
1. elephantsql nor
2. postgres running locally on windows

PR #288, opened 2023-12-20
Feat: Removed pdf-parse, and directly use latest pdf.js

marcusschiesser:


himself65 (commented):
You might need to update [README.md](https://github.com/run-llama/LlamaIndexTS#note-nextjs) and everything related to next.js

marcusschiesser:
@himself65 good catch, that's one todo for `create-llama`

PR #289, opened 2023-12-20
fix: align separator with llama_index

himself65:
https://github.com/run-llama/llama_index/blob/df9a30d63d402a21ccbbf5948afc68f057d913f2/llama_index/node_parser/text/sentence.py#L20

BREAKING:

remove `unixLineSeparator` and `windowsLineSeparator`, use `EOL` from `node:os` instead

### DIFF
--- a/packages/core/src/TextSplitter.ts
+++ b/packages/core/src/TextSplitter.ts
@@ -1,5 +1,5 @@
+import { EOL } from 'node:os'
 // GitHub translated
-
 import { globalsHelper } from "./GlobalsHelper";
 import { DEFAULT_CHUNK_OVERLAP, DEFAULT_CHUNK_SIZE } from "./constants";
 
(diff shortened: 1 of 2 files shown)


packages/core/src/TextSplitter.ts
```diff
@@ -1,5 +1,5 @@
+import { EOL } from 'node:os'
```
marcusschiesser, line 1: @himself65 does this run with Deno and/or Bun? Do you know a website like caniuse.com for this?

himself65: first, this is node-only, but Deno supports `node:xxx` URL schema now

second, caniuse is for Web API

himself65: I think there is no such website?

marcusschiesser: I also couldn't find such a website.

How about we put this platform-specific code in an own folder, like `packages/core/src/platforms/node`, and we import `EOL` from there?

In the long run, we could put all platform-specific code there, e.g. we're having this in `FileSystem.ts`:
```
export function getNodeFS(): CompleteFileSystem {
  const fs = require("fs/promises");
  return fs;
}
```
More and more platforms are coming out, and keeping this code in a dedicated place is good.
@yisding, what are your thoughts?

PR #290, opened 2023-12-21
refactor: sentence split

himself65:


packages/core/src/TextSplitter.ts
```diff
@@ -18,28 +19,27 @@ class TextSplit {
...
- */
-export const cjkSentenceTokenizer = (text: string) => {
-  // Accepts english style sentence endings with space and
-  // CJK style sentence endings with no space.
-  return text.match(
-    /.+?[.?!]+[\])'"`’”]*(?:\s|$)|.+?[。？！]+[\])'"`’”]*(?:\s|$)?|.+/g,
-  );
-};
+// Refs: https://github.com/fxsjy/jieba/issues/575#issuecomment-359637511
+const resentencesp = /([﹒﹔﹖﹗．；。！？]["’”」』]{0,2}|：(?=["‘“「『]{1,2}|$))/;
```
yisding, line 27: Interesting find. Are you sure this one doesn't have the same DoS issue?

packages/core/src/tests/TextSplitter.test.ts
```diff
@@ -1,4 +1,6 @@
-import { SentenceSplitter, cjkSentenceTokenizer } from "../TextSplitter";
+import { cjkSentenceTokenizer, SentenceSplitter } from '../TextSplitter'
+// @ts-expect-error
+import { Segment, useDefault as d } from 'segmentit'
```
yisding, line 3: Does this need to be added to the package.json?

yisding (commented):
Thanks Alex!

packages/core/src/tests/TextSplitter.test.ts
himself65: oh, forgot to remove the testing code

packages/core/src/TextSplitter.ts
himself65: image

https://devina.io/redos-checker

packages/core/src/TextSplitter.ts
```diff
@@ -18,28 +19,27 @@ class TextSplit {
...
+export function cjkSentenceTokenizer(sentence: string): string[] {
```
marcusschiesser, line 28:
```suggestion
/**
 * Tokenizes sentences. Suitable for Chinese, Japanese, and Korean. Use instead of `defaultSentenceTokenizer`.
 * @param text
 * @returns string[]
 */
export function cjkSentenceTokenizer(sentence: string): string[] {
```

PR #291, opened 2023-12-21
feat: integrate postgresql datasource

thucpn:


packages/create-llama/templates/components/vectordbs/pg/generate.mjs
```diff
@@ -0,0 +1,41 @@
...
+  const documents = await new SimpleDirectoryReader().loadData({
+    directoryPath: STORAGE_DIR,
+  });
+
+  // create postgres vector store
+  const vectorStore = new PGVectorStore();
+  vectorStore.db = new pg.Client({
+    connectionString: process.env.PG_CONNECTION_STRING,
+  });
+  vectorStore.db.connect();
```
marcusschiesser, line 25: @thucpn just seeing that there is an await missing:
```suggestion
  await vectorStore.db.connect();
```

marcusschiesser:
@thucpn i think your problem comes from not calling `vectorStore.checkSchema(db)`, but I also think that the user should not have to call this, let me refactor `PGVectorStore` and I'll re-assign this PR to you

packages/create-llama/templates/index.ts
```diff
@@ -39,8 +39,12 @@ const createEnvLocalFile = async (
       break;
     }
     case "pg": {
-      content += `# For generating a connection URI, see https://www.postgresql.org/docs/16/libpq-connect.html#LIBPQ-CONNSTRING-URIS\n`;
-      content += `PG_CONNECTION_STRING=\n`;
+      content += `# For generating a connection URI, see https://docs.timescale.com/use-timescale/latest/services/create-a-service\n`;
```
marcusschiesser, line 42: Why not using the connection string?

PR #295, opened 2023-12-28
Feat: Add MongoDB Vector DB support for Python projects in create-llama

marcusschiesser:


packages/create-llama/templates/index.ts
```diff
@@ -42,7 +42,7 @@ const createEnvLocalFile = async (
 
   switch (opts?.vectorDb) {
     case "mongo": {
-      content += `MONGODB_URI=\n`;
+      content += `MONGO_URI=\n`;
```
marcusschiesser, line 45:
```suggestion
      content += `# For generating a connection URI, see https://www.mongodb.com/docs/guides/atlas/connection-string\n`;
      content += `MONGO_URI=\n`;
```

packages/create-llama/templates/index.ts
```diff
@@ -317,7 +318,11 @@ const installPythonTemplate = async ({
   template,
   framework,
   engine,
-}: Pick<InstallTemplateArgs, "root" | "framework" | "template" | "engine">) => {
+  vectorDb,
```
marcusschiesser, line 321: are we already passing that parameter?

packages/create-llama/templates/types/simple/fastapi/pyproject.toml
```diff
@@ -12,6 +12,7 @@ uvicorn = { extras = ["standard"], version = "^0.23.2" }
 llama-index = "^0.9.19"
 pypdf = "^3.17.0"
 python-dotenv = "^1.0.0"
+pymongo = "^4.6.1"
```
marcusschiesser, line 15: Similar to modifying the package.json, I think we should programmatically modify the TOML file (in this case, adding pymongo if `mongo` is selected as vector DB). 
https://github.com/squirrelchat/smol-toml looks like a good solution for parsing and stringifying TOML files

PR #303, opened 2024-01-02
refactor: remove unused deps

himself65:
https://nodejs.org/dist/latest-v18.x/docs/api/crypto.html#class-hash


yisding (approved):
Thanks!

PR #304, opened 2024-01-02
style: prettier format

himself65:
run `pnpm run format`


marcusschiesser (commented):
@himself65 great catch! I also added `pnpm format` to husky's pre-commit script

PR #305, opened 2024-01-03
feat(core): init support for Ollama

himself65:
Document: https://github.com/jmorganca/ollama/blob/main/docs/api.md


examples/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+  "compilerOptions": {
+    "target": "es2016",
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true,
+    "paths": {
+      "llamaindex": ["../packages/core/src/index.ts"]
+    }
```
marcusschiesser, line 11: Currently, this was my approach to use the local build: https://github.com/run-llama/LlamaIndexTS/blob/844561b6cc2f7418ce3d529c05a2e908ff3a5a72/examples/README.md#using-the-local-build

I tried using `paths` before but was running into issues with multi-file examples (e.g. MongoDB)

If we add this, we should make sure of the following:
1. multi-file examples are running too
2. examples are using the llamaindex package from npm per default (customer use case)
3. update the documentation

marcusschiesser (commented):
@himself65 looks great. please have a look at my remarks

grigoriy-a:
@himself65 
Hey guys, sorry if I comment in the wrong place. I made a few runs of this new ollama code. There might be a problem with *streamChat() function, specifically with JSON.parse(Buffer.from(value).toString("utf-8")) line.
First, there is no try/catch there. Second, ollama sometimes returns several this in one go
`
{...}
{...}
{...}
`
So JSON.parse() throws.
So locally I did this
`
      let textChunks = Buffer.from(value).toString("utf-8")
        .split("\n")
        .filter(string => (string?.length ?? 0) > 0)
      for (let string of textChunks) {
        const json = JSON.parse(string)
        if (json.error) {
          throw new Error(json.error);
        }
        yield accessor(json)
      }
`
Looks like a hack, yeah. But it is what it is.

himself65:
> @himself65 Hey guys, sorry if I comment in the wrong place. I made a few runs of this new ollama code. There might be a problem with *streamChat() function, specifically with JSON.parse(Buffer.from(value).toString("utf-8")) line. First, there is no try/catch there. Second, ollama sometimes returns several this in one go
> 
> ```
> {...}
> {...}
> {...}
> ```
> 
> So JSON.parse() throws. So locally I did this
> 
> ```
>       let textChunks = Buffer.from(value).toString("utf-8")
>         .split("\n")
>         .filter(string => (string?.length ?? 0) > 0)
>       for (let string of textChunks) {
>         const json = JSON.parse(string)
>         if (json.error) {
>           throw new Error(json.error);
>         }
>         yield accessor(json)
>       }
> ```
> 
> Looks like a hack, yeah. But it is what it is.

Thanks for the comment, I will update this part later

himself65:
@grigoriy-a I've updated the code, could please give some feedback?

grigoriy-a:
@himself65 Made a few runs on your old code (just to reproduce) and on your new code.
Old code still throws. New code doesn't. Should be OK.
Thanks for your great work!

examples/package.json
```diff
@@ -7,7 +7,7 @@
     "@datastax/astra-db-ts": "^0.1.2",
     "@pinecone-database/pinecone": "^1.1.2",
     "commander": "^11.1.0",
-    "llamaindex": "latest",
+    "llamaindex": "workspace:*",
```
marcusschiesser, line 10: if use `workspace:*` here, it's probably not working with `npm`?

himself65: yeah, let me use another way

PR #306, opened 2024-01-03
Fix mongo issues

yisding:


packages/core/src/Node.ts
```diff
@@ -141,11 +141,13 @@ export abstract class BaseNode<T extends Metadata = Metadata> {
   }
 
   /**
-   * Used with built in JSON.stringify
+   * Creates a deep-clone of the node as JSON
    * @returns
    */
   toJSON(): Record<string, any> {
-    return { ...this, type: this.getType() };
+    const json: Record<string, any> = structuredClone(this);
```
marcusschiesser, line 148: @yisding I think the core issue is that this function did not return a deep clone.
Without this change, the test case `nodeToMetadata should not modify a node's metadata` would fail. 

Not using `structuredClone`, and instead doing an explicit deep clone in each node class would be better, but I think we need better unit testing for such a structural change.

Besides the unit tests, I also tested this PR with MongoDB vector store and `VectorStoreIndex`. 
As far as I know, the MongoDB load took longer than last time. it might be because of the deep cloning, but functionally, all tests passed.

If you're ok with these changes, then we can build a release.

yisding: OK so `toJSON` is a special function that's specifically designed to be used with JSON.stringify.

https://javascript.info/json

So inherently for that purpose a deep clone wouldn't be necessary because the object is being turned into a string anyways.

However, we're calling toJSON directly in the other function to further modify before stringification. I think what we should do is move the structuredClone outside of this function and into the caller.

packages/core/src/storage/vectorStore/utils.ts
```diff
@@ -16,20 +17,17 @@ export function nodeToMetadata(
   textField: string = DEFAULT_TEXT_KEY,
   flatMetadata: boolean = false,
 ): Metadata {
-  const nodeObj = node.toJSON();
-  const metadata = node.metadata;
+  const { metadata, embedding, ...rest } = _.cloneDeep(node.toJSON());
```
yisding, line 20: @marcusschiesser we should remove lodash and use structuredClone in the future, but right now we need to support Node 16 (unofficially)

packages/core/src/Node.ts
marcusschiesser: @yisding, great catch! I encapsulated these serialization functions and documented them in BaseNode, so we're not running into similar issues again.

PR #307, opened 2024-01-04
fix: don't create new hash deserialization a node

marcusschiesser:
@yisding While having fun with https://github.com/run-llama/LlamaIndexTS/pull/306, I found something that looks wrong to me: deserialization of a `Node` generates a new hash. I didn't test this extensively so that I would merge it after the next release. Please have a look.


himself65 (commented):
We use a hash cache map should be better?

marcusschiesser:
> We use a hash cache map should be better?

What do you mean? 

I think we can change the constructor to not generate a hash if the passed JSON has a hash included - that seems cleaner to me - what are your thoughts?

PR #309, opened 2024-01-04
Feat: Add sentence window retrieval

marcusschiesser:


packages/core/src/nodeParsers/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  const document = doc as Document;
+  const nodes: TextNode[] = [];
+
+  const textSplits = getTextSplitsFromDocument(document, textSplitter);
+
+  textSplits.forEach((textSplit) => {
+    const node = new TextNode({
+      text: textSplit,
+      metadata: includeMetadata ? _.cloneDeep(document.metadata) : {},
```
marcusschiesser, line 55: @yisding fun fact: we also had here just `document.metadata` before, which means that the existing NodeParser is working on the same metadata

PR #318, opened 2024-01-08
Feat: Add postgresql vector store for fastapi version

leehuwuj:
Following the TS version and the MongoDB codebase, I added the Postgres vector store component for FastAPI.

### DIFF
--- a/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
+++ b/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
@@ -30,7 +30,7 @@ def generate_datasource(service_context):
         show_progress=True,  # this will show you a progress bar as the embeddings are created
     )
     logger.info(
-        f"Successfully created embeddings in the PGVector, schema={store.schema_name} table={store.table_name}"
+        f"Successfully created embeddings in the PG vector store, schema={store.schema_name} table={store.table_name}"
     )
 
 
--- a/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
+++ b/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
@@ -1,0 +1,40 @@
+from dotenv import load_dotenv
+
+load_dotenv()
+import logging
+
+from app.engine.constants import DATA_DIR
+from app.engine.context import create_service_context
+from app.engine.utils import init_pg_vector_store_from_env
+
+from llama_index import (
+    SimpleDirectoryReader,
+    VectorStoreIndex,
+    StorageContext,
+)
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger()
+
+
+def generate_datasource(service_context):
+    logger.info("Creating new index")
+    # Initialize the PGVector environment variables
+
+    # load the documents and create the index
+    documents = SimpleDirectoryReader(DATA_DIR).load_data()
+    store = init_pg_vector_store_from_env()
+    storage_context = StorageContext.from_defaults(vector_store=store)
+    VectorStoreIndex.from_documents(
+        documents,
+        service_context=service_context,
+        storage_context=storage_context,
+        show_progress=True,  # this will show you a progress bar as the embeddings are created
+    )
+    logger.info(
+        f"Successfully created embeddings in the PGVector, schema={store.schema_name} table={store.table_name}"
+    )
+
+
+if __name__ == "__main__":
+    generate_datasource(create_service_context())
--- a/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
+++ b/packages/create-llama/templates/components/vectordbs/python/pg/generate.py
@@ -19,8 +19,6 @@ logger = logging.getLogger()
 
 def generate_datasource(service_context):
     logger.info("Creating new index")
-    # Initialize the PGVector environment variables
-
     # load the documents and create the index
     documents = SimpleDirectoryReader(DATA_DIR).load_data()
     store = init_pg_vector_store_from_env()
--- a/packages/create-llama/templates/components/vectordbs/python/pg/utils.py
+++ b/packages/create-llama/templates/components/vectordbs/python/pg/utils.py
@@ -1,0 +1,68 @@
+import os
+from llama_index.vector_stores import PGVectorStore
+from pydantic import BaseModel
+from urllib.parse import urlparse
+
+
+class PGConnectionElements(BaseModel):
+    user: str
+    password: str
+    host: str
+    port: int
+    database: str
+
+
+def destruct_pg_connection_string(conn_string: str) -> PGConnectionElements:
+    """
+    Destructs the PostgreSQL connection string into its elements.
+    """
+    parsed_url = urlparse(conn_string)
+
+    user = parsed_url.username
+    password = parsed_url.password
+    host = parsed_url.hostname
+    port = parsed_url.port
+    database = parsed_url.path[1:]
+
+    return PGConnectionElements(
+        user=user,
+        password=password,
+        host=host,
+        port=port,
+        database=database
+    )
+
+
+def get_pg_conn_elements_from_env() -> PGConnectionElements:
+    """
+    Returns the PostgreSQL connection elements from the environment variables:
+    [PGUSER, PGPASSWORD, PGHOST, PGPORT, PGDATABASE]
+    """
+    env_conn_str = os.environ.get("PG_CONNECTION_STRING")
+    if env_conn_str is not None:
+        return destruct_pg_connection_string(os.environ["PG_CONNECTION_STRING"])
+    else:
+        user = os.environ.get("PGUSER")
+        password = os.environ.get("PGPASSWORD")
+        host = os.environ.get("PGHOST")
+        port = int(os.environ.get("PGPORT", 5432))
+        database = os.environ.get("PGDATABASE")
+        return PGConnectionElements(
+            user=user,
+            password=password,
+            host=host,
+            port=port,
+            database=database
+        )
+
+
+def init_pg_vector_store_from_env():
+    # Initialize the PGVectorStore instance from environment variables
+    pg_conn_elements = get_pg_conn_elements_from_env()
+    pg_schema = os.environ.get("PGSCHEMA", "public")
+    pg_table = os.environ.get("POTABLE", "llamaindex_embedding")
+    return PGVectorStore.from_params(
+        **pg_conn_elements.model_dump(),
+        schema_name=pg_schema,
+        table_name=pg_table
+    )
--- a/packages/create-llama/templates/components/vectordbs/python/pg/utils.py
+++ b/packages/create-llama/templates/components/vectordbs/python/pg/utils.py
@@ -1,68 +1,23 @@
 import os
 from llama_index.vector_stores import PGVectorStore
-from pydantic import BaseModel
 from urllib.parse import urlparse
-
-
-class PGConnectionElements(BaseModel):
-    user: str
-    password: str
-    host: str
-    port: int
-    database: str
-
-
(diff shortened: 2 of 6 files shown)


packages/create-llama/templates/python.ts
```diff
@@ -21,6 +21,29 @@ const getAdditionalDependencies = (vectorDb?: TemplateVectorDB) => {
...
+    case "pg": {
+      dependencies.push(
+        ...[
+          {
+            name: "psycopg2-binary",
+            version: "^2.9.1",
+          },
+          {
+            name: "sqlalchemy",
+            version: "^2.0.25",
```
marcusschiesser, line 33: seems like that's already in `llama-index`: https://github.com/run-llama/llama_index/blob/7ab3f5008696e99ee78d06312b124d785d8c9fd8/pyproject.toml#L48C7-L48C7

also, the other dependencies are optional in `llama-index` - is there an easy way to use the version number specified in `llama-index`? e.g. by not specifying a number here?

packages/create-llama/templates/components/vectordbs/python/pg/generate.py
```diff
@@ -0,0 +1,38 @@
...
+    store = init_pg_vector_store_from_env()
+    storage_context = StorageContext.from_defaults(vector_store=store)
+    VectorStoreIndex.from_documents(
+        documents,
+        service_context=service_context,
+        storage_context=storage_context,
+        show_progress=True,  # this will show you a progress bar as the embeddings are created
+    )
+    logger.info(
+        f"Successfully created embeddings in the PGVector, schema={store.schema_name} table={store.table_name}"
```
marcusschiesser, line 33:
```suggestion
        f"Successfully created embeddings in the PG vector store, schema={store.schema_name} table={store.table_name}"
```

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
```diff
@@ -0,0 +1,68 @@
...
+            port=port,
+            database=database
+        )
+
+
+def init_pg_vector_store_from_env():
+    # Initialize the PGVectorStore instance from environment variables
+    pg_conn_elements = get_pg_conn_elements_from_env()
+    pg_schema = os.environ.get("PGSCHEMA", "public")
+    pg_table = os.environ.get("POTABLE", "llamaindex_embedding")
```
marcusschiesser, line 63: how about getting this from `constants.py` (similar to TS code, see https://github.com/run-llama/LlamaIndexTS/blob/85da7985388d92cb71f60f34e7d205fa142a58e3/packages/create-llama/templates/components/vectordbs/typescript/pg/shared.mjs#L4C1-L4C1)

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        host = os.environ.get("PGHOST")
+        port = int(os.environ.get("PGPORT", 5432))
+        database = os.environ.get("PGDATABASE")
+        return PGConnectionElements(
+            user=user,
+            password=password,
+            host=host,
+            port=port,
+            database=database
+        )
```
marcusschiesser, line 56: It's a great idea to provide an alternative way to configure the DB parameters. 
Here I wouldn't do it, though as the generator is only generating `PG_CONNECTION_STRING` - if it doesn't exist, I would just throw an exception and tell the user to configure `PG_CONNECTION_STRING`

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+class PGConnectionElements(BaseModel):
+    user: str
+    password: str
+    host: str
+    port: int
+    database: str
+
+
+def destruct_pg_connection_string(conn_string: str) -> PGConnectionElements:
```
marcusschiesser, line 15: maybe add as static class method to `PGConnectionElements`? like `from_connection_string`?

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
marcusschiesser: as this method becomes then i one-liner, you can inline it to `init_pg_vector_store_from_env`

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
marcusschiesser: if the connection string can not be parsed, you can throw an exception (see comment below)

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
leehuwuj: > It's a great idea to provide an alternative way to configure the DB parameters. Here I wouldn't do it, though as the generator is only generating `PG_CONNECTION_STRING` - if it doesn't exist, I would just throw an exception and tell the user to configure `PG_CONNECTION_STRING`

this document says that the user can also set the credential variables (https://github.com/run-llama/LlamaIndexTS/blob/85da7985388d92cb71f60f34e7d205fa142a58e3/examples/pg-vector-store/README.md?plain=1#L22), is it related to this template or just setup the environment itself?

If so, because the PGVectorStore in llama_index required the async connection string so how about constructing the async one from the PG_CONNECTION_STRING by replacing the schema then passing them to PGVectorStore directly? (https://github.com/run-llama/llama_index/blob/2edd31c83b3648fc81eff636cebc183697958d12/llama_index/vector_stores/postgres.py#L130)

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
leehuwuj: okay, let me update this one

packages/create-llama/templates/python.ts
leehuwuj: Yes, we can add the `postgres` extras attribute to the `llama-index` dependency in pyproject.toml but the python render codebase does not support that, let me work further on that.

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
leehuwuj: I pushed an update, please help me to review it again, thank you!

packages/create-llama/templates/python.ts
```diff
@@ -21,6 +22,13 @@ const getAdditionalDependencies = (vectorDb?: TemplateVectorDB) => {
       });
       break;
     }
+    case "pg": {
+      dependencies.push({
+        name: "llama-index",
+        version: "^0.9.19",
```
marcusschiesser, line 28: the problem here is that we need to update now llama-index version in two places, here in in the original toml

marcusschiesser: so i would leave the version out - and the updater below is using the the existing version

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
marcusschiesser: @leehuwuj that looks great thanks. The doc you're referring to is just an example for LlamaIndexTS; it's not about how to use the Python version

packages/create-llama/templates/components/vectordbs/python/pg/utils.py
```diff
@@ -0,0 +1,23 @@
...
+def init_pg_vector_store_from_env():
+    original_conn_string = os.environ.get("PG_CONNECTION_STRING")
+    if original_conn_string is None:
+        raise ValueError("PG_CONNECTION_STRING environment variable is not set.")
+
+    # The PGVectorStore requires both two connection strings, one for psycopg2 and one for asyncpg
+    # Update the configured scheme with the psycopg2 and asyncpg schemes
+    original_scheme = urlparse(original_conn_string).scheme + "://"
+    conn_string = original_conn_string.replace(original_scheme, "postgresql+psycopg2://")
+    async_conn_string = original_conn_string.replace(original_scheme, "postgresql+asyncpg://")
```
marcusschiesser, line 16: @logan-markewich FYI: I checked the llama-index code - @leehuwuj is correct; a user has to pass both connection strings - so we will keep that code now in `create-llama` - you probably want to move this code here to the `PGVectorStore` constructor, so that a user only has to pass `connection_string`

packages/create-llama/templates/python.ts
leehuwuj: how about this update @marcusschiesser https://github.com/leehuwuj/LlamaIndexTS/commit/5df2967065c6ab9e977a9a3634752d524983088b ? 
i implemented in an easy approach so it could be broken if we did not set the value in toml by a string of package version but in more complex toml scheme. In the long term, we may need a generic approach to merge dependencies in toml with the adding one in code.

marcusschiesser: i think it should behave like this:
1. dependency doesn't exist: add it, but version must exist (if not throw an error)
2. if dependency does exist, update version and/or extras field

marcusschiesser: as your `extras` approach is better, please also add it for MongoDB (if possible - and test it)

leehuwuj: @marcusschiesser i updated the code logic. I also checked the `llama-index` the pymongo is not a extras dependency on that so i think we should consider it as additional one in our code base.

packages/create-llama/templates/python.ts
```diff
@@ -43,7 +50,29 @@ const addDependencies = async (
     const tool = fileParsed.tool as any;
     const existingDependencies = tool.poetry.dependencies as any;
     for (const dependency of dependencies) {
-      existingDependencies[dependency.name] = dependency.version;
+      const orignalValue = existingDependencies[dependency.name];
```
marcusschiesser, line 53: typo `originalValue`

packages/create-llama/templates/python.ts
```diff
@@ -43,7 +50,29 @@ const addDependencies = async (
...
+      if (orignalValue === undefined) {
+        // Add dependency
+        if (dependency.version === undefined) {
+          throw Error(
+            `Dependency ${dependency.name} is missing version attribute!`,
+          );
+        } else {
+          existingDependencies[dependency.name] = dependency.version;
```
marcusschiesser, line 61: missing extras

packages/create-llama/templates/python.ts
```diff
@@ -43,7 +50,29 @@ const addDependencies = async (
...
+        // Update dependency
+        if (typeof orignalValue === "string") {
+          existingDependencies[dependency.name] = {
+            version: dependency.version ?? orignalValue,
+            extras: dependency.extras,
+          };
+        } else {
+          orignalValue.version = dependency.version ?? orignalValue.version;
+          orignalValue.extras = dependency.extras;
+          existingDependencies[dependency.name] = orignalValue;
```
marcusschiesser, line 73: i would generate a `value` and then always call this:
```suggestion
          existingDependencies[dependency.name] = value;
```

marcusschiesser: this function is getting complicated - better factor it out, then we can easily unit test it (once we have unit tests for create-llama)

packages/create-llama/templates/python.ts
```diff
@@ -21,12 +22,43 @@ const getAdditionalDependencies = (vectorDb?: TemplateVectorDB) => {
...
+        extras: ["postgres"],
+      });
+    }
   }
 
   return dependencies;
 };
 
-const addDependencies = async (
+const mergePoetryDependencies = (
```
marcusschiesser, line 36: @leehuwuj yeah, this is nice!

marcusschiesser (approved):

PR #368, opened 2024-01-12
feat: support showing image on chat message

thucpn:


packages/create-llama/templates/types/streaming/nextjs/app/components/chat-section.tsx
```diff
@@ -12,17 +14,22 @@ export default function ChatSection() {
...
+    data,
   } = useChat({
     api: process.env.NEXT_PUBLIC_CHAT_API,
     headers: {
       "Content-Type": "application/json", // using JSON because of vercel/ai 2.2.26
     },
   });
 
+  const transformedMessages = useMemo(() => {
+    return transformMessages(messages, data);
```
marcusschiesser, line 26: how about this name?
```suggestion
    return insertDataIntoMessages(messages, data);
```

marcusschiesser (commented):
@thucpn looking great! just a few comments

packages/create-llama/templates/types/streaming/nextjs/app/api/chat/route.ts
```diff
@@ -42,7 +42,7 @@ export async function POST(request: NextRequest) {
     }
 
     const llm = new OpenAI({
-      model: process.env.MODEL || "gpt-3.5-turbo",
+      model: process.env.MODEL || ("gpt-3.5-turbo" as any),
```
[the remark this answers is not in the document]
marcusschiesser: Ah, i thought it would be something like `(process.env.MODEL as any) ?? "gpt-3.5-turbo"` then

packages/create-llama/templates/index.ts
```diff
@@ -156,6 +156,10 @@ export const installTemplate = async (
       props.openAiKey,
       props.vectorDb,
     );
+  } else if (props.forBackend) {
```
marcusschiesser, line 159: why do we need this check?  isn't `!backend` sufficient?

packages/create-llama/templates/index.ts
```diff
@@ -156,6 +156,10 @@ export const installTemplate = async (
...
+    // this is a frontend for fullstack templates, create .env file with uncredentialed variables
```
marcusschiesser, line 160:
```suggestion
    // this is a frontend for a full-stack app, create .env file with model information
```

packages/create-llama/templates/types/streaming/nextjs/app/api/chat/route.ts
```diff
@@ -42,7 +42,7 @@ export async function POST(request: NextRequest) {
     }
 
     const llm = new OpenAI({
-      model: process.env.MODEL || "gpt-3.5-turbo",
+      model: (process.env.MODEL as any) ?? "gpt-3.5-turbo",
```
himself65, line 45: Is this necessary?

himself65: Reduce `any` type as possible

marcusschiesser: I had the same question. Problem is that model is of type 'keyof ALL_MODELS'. Any idea besides using any or exporting a type alias (means change to core?)

himself65: oh, I've changed the type to `keyof ALL_MODELS | string` in the last commit, so it's not be ok

PR #370, opened 2024-01-12
chore(core): use `bunchee` to bundle

himself65:


yisding:
For context here: @himself65  suggested we try bunchee to make the process of doing ESM exports (particularly in the future if we want to have separate import paths like llamaindex/loaders) easier. Let's give this a try and make sure it doesn't break the build.

PR #371, opened 2024-01-12
refactor: use `crypto.randomUUID`

himself65:
Reduce bundle size and use web standards.

Ref: https://nodejs.org/docs/latest/api/crypto.html#cryptorandomuuidoptions

### DIFF
--- a/packages/core/src/ChatEngine.ts
+++ b/packages/core/src/ChatEngine.ts
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from "uuid";
+import { randomUUID } from "node:crypto";
 import { ChatHistory } from "./ChatHistory";
 import { NodeWithScore, TextNode } from "./Node";
 import {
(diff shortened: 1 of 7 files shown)


packages/core/src/ChatEngine.ts
```diff
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from "uuid";
+import { randomUUID } from "node:crypto";
```
marcusschiesser, line 1: @himself65 we have to be careful with adding so much node specific code - will that work with edge runtime? Is there an alternative in https://edge-runtime.vercel.app/features/available-apis ?

himself65: `crypto.randomUUID` is also web standard, but yes we should use another way to import such APIs

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

himself65: No, I believe edge has polyfill for such APIs

https://github.com/vercel/edge-runtime/blob/a6304c3a792482cec729d17a8981465bb96bd131/packages/primitives/src/primitives/crypto.js#L4

PR #372, opened 2024-01-12
fix: cover type check on all ts files

himself65:
Use ts monorepo to cover all ts files. Make type check in CI to make sure all files are type correct

Similar PR: https://github.com/dai-shi/waku/pull/132, https://github.com/toeverything/AFFiNE/pull/2732


packages/core/package.json
```diff
@@ -44,17 +44,11 @@
...
   "exports": {
     ".": {
       "types": "./dist/index.d.mts",
       "import": "./dist/index.mjs",
       "require": "./dist/index.js"
-    },
-    "./*": {
-      "types": "./dist/*.d.mts",
-      "import": "./dist/*.mjs",
-      "require": "./dist/*.js"
```
himself65, line 57: cannot use, some type might be missing

packages/create-llama/tsconfig.json
```diff
@@ -1,11 +1,19 @@
 {
+  "extends": "../../tsconfig.json",
```
marcusschiesser, line 2: @himself65 what's the advantage of this dependency? a disadvantage that I see is that it makes it more difficult to separate `create-llama` out of this repository (which is something we might want to do as it grows more and more)

himself65: I originally wanted to make typescript check all ts files because some examples code was already outdated but we didn't know it. I think it's very simple to separate `create-llama`, just copy upstream tsconfig and merge it into one tsconfig.

marcusschiesser: > I originally wanted to make typescript check all ts files because some examples code was already outdated but we didn't know it. 
Yes that's a very good idea. I just wondered if we need this tsconfig.json dependency for this?

himself65: Not required

Issue #367, opened 2024-01-11
How to use Together AI llms instead of OpenAI llms?


jamessss213:
Can anyone tell me how to use Together AI or Replicate Llama-2 instead of GPTS? How to remove default OpenAI APi ?

jamessss213:
@yisding @himself65 Can you pls help?

himself65:
We don't have together AI for ts package now. But this is a good feature

himself65:
read the code looks like very easy to implement, let me do this quickly

jamessss213:
@himself65 Alex, I want to ask questions based on my data. Any example for that?

himself65:
https://www.npmjs.com/package/llamaindex/v/0.0.45

I'd like for you to read these examples to get some ideas.

- https://github.com/run-llama/LlamaIndexTS/blob/main/examples/together-ai.ts

- https://github.com/run-llama/LlamaIndexTS/blob/main/examples/vectorIndexGPT4.ts

jamessss213:
@himself65 i tried this but it gives an error Error: Set OpenAI Key in OPENAI_API_KEY env variable 

```
import fs from "node:fs/promises";

import {
  Document,
  serviceContextFromDefaults,
  VectorStoreIndex,
  TogetherLLM,
} from "llamaindex";

async function main() {
  const path = "node_modules/llamaindex/examples/abramov.txt";
  const essay = await fs.readFile(path, "utf-8");

  const document = new Document({ text: essay, id_: path });

  const serviceContext = serviceContextFromDefaults({
    llm: new TogetherLLM({ model: "mistralai/Mixtral-8x7B-Instruct-v0.1" }),
  });

  const index = await VectorStoreIndex.fromDocuments([document], {
    serviceContext,
  });

  const queryEngine = index.asQueryEngine();
  const response = await queryEngine.query(
    "What did the author do in college?",
  );

  console.log(response.toString());
}

main().catch(console.error);
```

himself65:
You might also need to change embedModel to TogetherEmbending

himself65:
Thanks for report, I will update the example and fix the bug

himself65:
Please set `TOGETHER_API_KEY`

jamessss213:
[...]

PR #373, opened 2024-01-12
feat: support Together AI

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/367

### DIFF
--- a/packages/core/src/llm/LLM.ts
+++ b/packages/core/src/llm/LLM.ts
@@ -129,7 +129,7 @@ export class OpenAI implements LLM {
   hasStreaming: boolean = true;
 
   // Per completion OpenAI params
-  model: keyof typeof ALL_AVAILABLE_OPENAI_MODELS;
+  model: keyof typeof ALL_AVAILABLE_OPENAI_MODELS | string;
   temperature: number;
   topP: number;
   maxTokens?: number;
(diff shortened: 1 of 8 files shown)


packages/core/src/llm/LLM.ts
```diff
@@ -205,12 +205,16 @@ export class OpenAI implements LLM {
   }
 
   get metadata() {
+    const contextWindow =
+      ALL_AVAILABLE_OPENAI_MODELS[
+        this.model as keyof typeof ALL_AVAILABLE_OPENAI_MODELS
+      ]?.contextWindow ?? 1024;
```
marcusschiesser, line 211: where does 1024 come from?

packages/core/src/llm/LLM.ts
```diff
@@ -129,7 +129,7 @@ export class OpenAI implements LLM {
   hasStreaming: boolean = true;
 
   // Per completion OpenAI params
-  model: keyof typeof ALL_AVAILABLE_OPENAI_MODELS;
+  model: keyof typeof ALL_AVAILABLE_OPENAI_MODELS | string;
```
marcusschiesser, line 132: what's the advantage of allowing string here? The disadvantage that I see is that it's easy to type a wrong model name, e.g. `"gpt-3.5-turbo-1107"` and then our code won't work

himself65: But it will be easy to extend the class since some models, like together ai use the same RESTful API with open ai.

packages/core/src/llm/LLM.ts
himself65: Good question, I looked together AI document and cannot find a correct value

marcusschiesser: if it's used only in together AI, how about overriding metadata there?

packages/core/src/llm/LLM.ts
marcusschiesser: but aren't you already overriding model to `string` in together ai?

marcusschiesser: what i don't like here is that we should have a type alias for `keyof typeof ALL_AVAILABLE_OPENAI_MODELS` so we can use that in `model`

packages/core/src/llm/LLM.ts
himself65: yes, I will fix then

PR #380, opened 2024-01-15
feat(express): support showing image on chat message express backend

thucpn:


packages/create-llama/templates/types/streaming/express/src/controllers/chat.controller.ts
```diff
@@ -47,14 +47,29 @@ export const chat = async (req: Request, res: Response) => {
...
-    const stream = LlamaIndexStream(response);
+    const { stream, data: streamData } = LlamaIndexStream(response, {
+      parserOptions: {
+        image_url: data?.imageUrl,
+      },
+    });
 
-    streamToResponse(stream, res);
+    // Pipe LlamaIndexStream to response
+    const processedStream = stream.pipeThrough(streamData.stream);
```
marcusschiesser, line 62: would it make sense to do this in `LlamaIndexStream` and just return one value there?

packages/create-llama/templates/types/streaming/express/src/controllers/chat.controller.ts
```diff
@@ -47,14 +47,29 @@ export const chat = async (req: Request, res: Response) => {
...
+    return streamToResponse(processedStream, res, {
+      headers: {
+        // MUST have the `X-Experimental-Stream-Data: 'true'` header
+        // in response so the client uses the correct parsing logic
+        // https://sdk.vercel.ai/docs/api-reference/stream-data#on-the-server
```
marcusschiesser, line 67:
```suggestion
        // resonse MUST have the `X-Experimental-Stream-Data: 'true'` header
        // so that the client uses the correct parsing logic, see
        // https://sdk.vercel.ai/docs/api-reference/stream-data#on-the-server
```

marcusschiesser:
merge after https://github.com/run-llama/LlamaIndexTS/pull/396

marcusschiesser (approved):

PR #381, opened 2024-01-15
Draft: Add poetry installation support for FastAPI

leehuwuj:
- Add checking is poetry installed (using `poetry --version`).
- Add installing python dependencies (using `poetry install`).
- Add generating index in FastAPI app if the environment was set-up correctly (`poetry run python app/engine/generate.py`).

### DIFF
--- a/.changeset/silly-bugs-fold.md
+++ b/.changeset/silly-bugs-fold.md
@@ -1,0 +1,7 @@
+---
+"create-llama": patch
+---
+
+- Add support to install python dependencies
+- Add new create-llama option: `--install-dependencies`:
+  This option allows the user to choose whether to install the dependencies automatically or manually when creating a llama project. This improves the development experience of the tool.
--- a/.changeset/silly-bugs-fold.md
+++ b/.changeset/silly-bugs-fold.md
@@ -2,6 +2,4 @@
 "create-llama": patch
 ---
 
-- Add support to install python dependencies
-- Add new create-llama option: `--install-dependencies`:
-  This option allows the user to choose whether to install the dependencies automatically or manually when creating a llama project. This improves the development experience of the tool.
+Added option to automatically install dependencies (for Python and TS)
--- a/packages/create-llama/helpers/poetry.ts
+++ b/packages/create-llama/helpers/poetry.ts
@@ -1,0 +1,34 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { execSync } from "child_process";
+import fs from "fs";
+
+export function isPoetryAvailable(): boolean {
+  try {
+    execSync("poetry --version", { stdio: "ignore" });
+    return true;
+  } catch (_) {}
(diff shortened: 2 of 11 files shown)


packages/create-llama/helpers/python.ts
```diff
@@ -146,7 +148,26 @@ export const installPythonTemplate = async ({
...
 
-  console.log(
-    "\nPython project, dependencies won't be installed automatically.\n",
-  );
+  // install python dependencies
+  if (isPoetryAvailable()) {
+    console.log(
+      `Installing python dependencies using poetry. This may take a while...`,
+    );
+    const is_installed_success = tryPoetryInstall();
```
marcusschiesser, line 156: TS is using camel case for variables ;)
```suggestion
    const installSuccessful = tryPoetryInstall();
```

packages/create-llama/helpers/poetry.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { execSync } from "child_process";
+import fs from "fs";
+
+export function isPoetryAvailable(): boolean {
+  try {
+    execSync("poetry --version", { stdio: "ignore" });
+    return true;
+  } catch (_) {}
```
marcusschiesser, line 9: what error handling is used in `callPackageManager`?

leehuwuj: sorry, i don't get it. Do you mean that we should handle the error explicitly for this function? about callPackageManager we do not catch the error from/in that

marcusschiesser: I mean take a look at `callPackageManager` how they are calling a child process
e.g. they are using `stdio: "inherit"` and checking for the return code which I am not sure is handled here correctly

packages/create-llama/helpers/types.ts
```diff
@@ -23,4 +23,5 @@ export interface InstallTemplateArgs {
   communityProjectPath?: string;
   vectorDb?: TemplateVectorDB;
   externalPort?: number;
+  isPoetryInstall?: boolean;
```
marcusschiesser, line 26: generalize this question (also ask it for TS) and rename variable
```suggestion
  installDependencies?: boolean;
```

marcusschiesser (approved):

.changeset/silly-bugs-fold.md
```diff
@@ -0,0 +1,7 @@
+---
+"create-llama": patch
+---
+
+- Add support to install python dependencies
+- Add new create-llama option: `--install-dependencies`:
+  This option allows the user to choose whether to install the dependencies automatically or manually when creating a llama project. This improves the development experience of the tool.
```
marcusschiesser, line 7:
```suggestion
Added option to automatically install dependencies (for Python and TS)
```

PR #384, opened 2024-01-15
refactor: remove usage of `lodash`

himself65:
BREAKING:
- _.isEqual is replaced by `===` so object equal will be checked by reference.
- will drop node.js 16 support


yisding (changes requested):
Unfortunately Node 16 doesn't support `structuredClone.`

PR #392, opened 2024-01-16
fix: replace missing import * as

marcusschiesser:


himself65 (approved):
use node:path would be better, but we could do it in the future

PR #393, opened 2024-01-16
Feat: Add streaming support for QueryEngine 

marcusschiesser:
- Unify streaming interface with ChatEngine (use OpenAI method)
- Add parameter object for ChatEngine and QueryEngine methods

### DIFF
--- a/apps/docs/docs/modules/high_level/chat_engine.md
+++ b/apps/docs/docs/modules/high_level/chat_engine.md
@@ -11,7 +11,16 @@ const retriever = index.asRetriever();
 const chatEngine = new ContextChatEngine({ retriever });
 
 // start chatting
-const response = await chatEngine.chat(query);
+const response = await chatEngine.chat({ message: query });
+```
+
+The `chat` function also supports streaming, just add `stream: true` as an option:
+
+```typescript
+const stream = await chatEngine.chat({ message: query, stream: true });
+for await (const chunk of stream) {
+  process.stdout.write(chunk.response);
+}
 ```
 
 ## Api References
--- a/packages/core/src/llm/utils.ts
+++ b/packages/core/src/llm/utils.ts
@@ -1,10 +1,24 @@
-// TODO: use for LLM.ts
-
 export async function* streamConverter<S, D>(
   stream: AsyncIterable<S>,
   converter: (s: S) => D,
 ): AsyncIterable<D> {
   for await (const data of stream) {
     yield converter(data);
+  }
+}
+
+export async function* streamReducer<S, D>(params: {
+  stream: AsyncIterable<S>;
+  reducer: (previousValue: D, currentValue: S) => D;
+  initialValue: D;
+  finished?: (value: D | undefined) => void;
+}): AsyncIterable<S> {
+  let value = params.initialValue;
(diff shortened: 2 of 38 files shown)


apps/docs/docs/modules/high_level/chat_engine.md
```diff
@@ -11,7 +11,16 @@ const retriever = index.asRetriever();
...
-const response = await chatEngine.chat(query);
+const response = await chatEngine.chat({ message: query });
+```
+
+The `chat` function also supports streaming, just add `stream: true` as an option:
+
+```typescript
+const stream = await chatEngine.chat({ message: query, stream: true });
+for await (const chunk of stream) {
+  process.stdout.write(chunk.response);
```
marcusschiesser, line 22: @yisding @himself65 I am streaming here (and in `QueryEngine`) `Response` objects and store the delta in the `response` string.

We might alternatively:
1. Use a dedicated interface (e.g. `ResponseChunk`) containing the deltas (like we're doing it for the LLM interface)
2. Extend the `Response` interface with a `delta` 

What are your thoughts?

marcusschiesser: Also, we might want to rename the `message` parameter. This was the original name, but now that it's a named parameter, it feels kind of weird, as it's just a `string` and not a message object.

packages/core/src/llm/utils.ts
```diff
@@ -8,3 +6,19 @@ export async function* streamConverter<S, D>(
     yield converter(data);
   }
 }
+
+export async function* streamReducer<S, D>(params: {
```
marcusschiesser, line 10: @himself65 @yisding I wrote these streaming helpers - they simplify much of our streaming code. Do you know a library that we can use to deal with this concern?

himself65: nodejs utils might have such thing but we shouldn't use that

apps/docs/docs/modules/high_level/chat_engine.md
himself65: > we might want to rename the `message` parameter

use `text`?

himself65 (approved):

packages/core/src/llm/utils.ts
marcusschiesser: ok, then we just leave it as is for now. Please have a look, those helpers are also useful for adding new LLMs

apps/docs/docs/modules/high_level/chat_engine.md
marcusschiesser: Actually, it's not just `text` - we also have multi-modal messages. 

see a multi-modal usage:
https://github.com/run-llama/LlamaIndexTS/blob/ab9beefa0d18ef8dc775a5d1c5d253828d16df05/packages/create-llama/templates/types/streaming/nextjs/app/api/chat/route.ts#L51-L62

marcusschiesser: The precise naming would be `userMessageContent`, which is too wordy, and I think just `content` is a bit confusing. Let's stick with `message` for now.

PR #396, opened 2024-01-17
fix: update create-llama examples for new chat engine

marcusschiesser:


packages/create-llama/templates/types/streaming/express/src/controllers/llamaindex-stream.ts
```diff
@@ -4,27 +4,25 @@ import {
...
-      const { value, done } = await res.next();
-      if (done) {
-        controller.close();
-        return;
-      }
-
-      const text = trimStartOfStream(value ?? "");
-      if (text) {
-        controller.enqueue(text);
+      for await (const message of res) {
```
marcusschiesser, line 13: @thucpn good catch! We're using here now an asynciterable. But better not consume all of it in the pull function (`pull` is called for pulling just one token).

You'll still have to use `next` by doing `let it = res[Symbol.asyncIterator]();`

see
https://github.com/vercel/ai/blob/b2349746892a292cfbbda5a59e8a125f53aa55b5/packages/core/streams/ai-stream.ts#L263

PR #397, opened 2024-01-17
feat: support showing image for fastapi template

thucpn:


marcusschiesser:
Will continue working on this in https://github.com/run-llama/create-llama

PR #398, opened 2024-01-17
refactor: encourage using parameter objects 

marcusschiesser:
eslint shows errors for functions with more than four parameters


yisding (approved):
I see a few lint errors with the new lint rules in the e2e tests. Can you disable the rule on that file? Otherwise looks good.

himself65 (approved):

PR #399, opened 2024-01-17
add run app option

leehuwuj:
Add an option that allows the user to run the app after installation.


packages/create-llama/questions.ts
```diff
@@ -39,9 +39,9 @@ const getVectorDbChoices = (framework: TemplateFramework) => {
     { title: "PostgreSQL", value: "pg" },
   ];
 
-  const vectodbLang = framework === "fastapi" ? "python" : "typescript";
+  const vectordbLang = framework === "fastapi" ? "python" : "typescript";
```
marcusschiesser, line 42: yeah, we hate typos :D

packages/create-llama/index.ts
```diff
@@ -234,6 +237,30 @@ async function run(): Promise<void> {
...
+        cps.map(
+          (cp) =>
+            new Promise((resolve, reject) => {
+              cp.on("exit", resolve);
+              cp.on("error", reject);
+            }),
+        ),
+      );
+    } catch (e) {
+      throw new RunApplicationError();
```
marcusschiesser, line 261: error handling in this app is not currently by throwing a specific error type.
if you don't want the generic error handler,  how about just logging the error here and calling `process.exit(1)`?

marcusschiesser: you could then also just do here:

```
await runApp(
        resolvedProjectPath,
        program.frontend,
        program.framework,
        program.externalPort,
      );
 ```

and put the process locking into `runApp`

File: packages/create-llama/questions.ts [hunk 3]
```diff
@@ -54,6 +54,27 @@ const getVectorDbChoices = (framework: TemplateFramework) => {
...
 };
 
+const getActionChoices = (program: QuestionArgs) => {
+  let actionChoices = [
+    {
+      title: "Just generate code (~1 sec)",
+      value: "1",
+    },
+    {
+      title: "Generate code and install dependencies (~2 min)",
```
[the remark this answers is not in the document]
leehuwuj: yes, it takes me 2 minutes

packages/create-llama/questions.ts
```diff
@@ -54,6 +54,27 @@ const getVectorDbChoices = (framework: TemplateFramework) => {
...
+      value: "2",
+    },
+  ];
+
+  if (program.vectorDb === "none" && program.openAiKey !== "") {
```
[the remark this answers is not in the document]
leehuwuj: ah yeah, we should allow user to choose to run the app if they had set the OPENAI_API_KEY in the environment.

packages/create-llama/questions.ts
```diff
@@ -326,6 +334,32 @@ export const askQuestions = async (
...
+    switch (action) {
+      case "1":
+        program.installDependencies = false;
+        break;
+      case "2":
+        program.installDependencies = true;
+        break;
+      case "3":
+        program.installDependencies = true;
+        program.runApp = true;
```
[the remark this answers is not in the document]
leehuwuj: It requires updating the feature that we had merged recently, but not much effort so I've just updated. Thank you for the suggestion.

packages/create-llama/index.ts
leehuwuj: I updated the runApp behavior to support better logging and handle child processes as well. Please have a look, thank you!

packages/create-llama/e2e/basic.spec.ts
```diff
@@ -48,6 +48,7 @@ for (const templateType of templateTypes) {
           let cwd: string;
           let name: string;
           let cps: ChildProcess[] = [];
+          const postInstallAction = "dependencies";
```
marcusschiesser, line 51: why not using `runApp`?

```suggestion
          const postInstallAction = "runApp";
```

packages/create-llama/index.ts
```diff
@@ -120,10 +121,10 @@ Select external port.
 `,
   )
   .option(
-    "--install-dependencies",
+    "--post-install-action <action>",
     `
 
-Whether install dependencies (backend/frontend) automatically or not.
+Choose an action after installation. For example, runApp or dependencies. Let the default option be to just generate the app.
```
marcusschiesser, line 127:
```suggestion
Choose an action after installation. For example, 'runApp' or 'dependencies'. The default option is just to generate the app.
```

packages/create-llama/helpers/typescript.ts
```diff
@@ -211,17 +240,7 @@ export const installTSTemplate = async ({
...
-
-    console.log("\nInstalling devDependencies:");
-    for (const dependency in packageJson.devDependencies)
-      console.log(`- ${cyan(dependency)}`);
-
-    console.log();
-
-    await callPackageManager(packageManager, isOnline);
+  if (postInstallAction !== "none") {
+    await installTSDependencies(root, packageManager, true);
```
marcusschiesser, line 244: why not like this?
```suggestion
    await installTSDependencies(packageJson, packageManager, isOnline);
```

packages/create-llama/helpers/python.ts
```diff
@@ -154,28 +180,7 @@ export const installPythonTemplate = async ({
...
-        yellow(
-          `Poetry is not available in the current environment. The Python dependencies will not be installed automatically.
-Please check ${terminalLink(
-            "Poetry Installation",
-            `https://python-poetry.org/docs/#installation`,
-          )} to install poetry first, then install the dependencies manually.`,
-        ),
-      );
-    }
+  if (postInstallAction) {
```
marcusschiesser, line 183: is that condition correct? looks like it would also be executed for `none`

packages/create-llama/e2e/utils.ts
```diff
@@ -3,6 +3,7 @@ import crypto from "node:crypto";
 import { mkdir } from "node:fs/promises";
 import * as path from "path";
 import waitPort from "wait-port";
+import { TemplateFramework } from "../helpers";
```
marcusschiesser, line 6: could also import other type, e.g. TemplatePostInstallAction

packages/create-llama/e2e/basic.spec.ts
marcusschiesser: then we could simplify the test code and also cover the test for runapp

leehuwuj: runApp action will block the process until it gets an exit signal, while the testing will start the app and return child processes and kill them afterward. It'll be better to refactor the runApp as a non-blocking function to improve the flexibility, do you want me to implement it in this PR?

packages/create-llama/helpers/python.ts
leehuwuj: ah, my bad. this one is not covered by in tests

packages/create-llama/e2e/utils.ts
```diff
@@ -1,27 +1,46 @@
...
+export type CreateLlamaResult = {
+  projectName: string;
+  appProcess: ChildProcess;
+};
 
 export async function runApp(
   cwd: string,
   name: string,
   appType: AppType,
+  framework: TemplateFramework,
```
marcusschiesser, line 25: can't we remove `runApp` function?

packages/create-llama/index.ts
```diff
@@ -111,20 +111,27 @@ const program = new Commander.Command(packageJson.name)
     `
 
   Select OpenAI model to use. E.g. gpt-3.5-turbo.
+`,
```
marcusschiesser, line 114: It's a good idea to make both ports configurable, but now it seems we're having a naming problem.
As the user of `create-llama`, there are two ports:

1. the `appPort`
2. the `backendPort` (optional)

if there's no frontend, we just have the app port, if there is a frontend, the frontend will use the `appPort` and connect to the backend via the `backendPort`.

Can you please check that the naming is consistent inside of `create-llama`?

leehuwuj: Yes, it is consistent, but we already used `externalPort` and `port` so I did not change the name. In some functions, i mapped them internally to backendPort and frontendPort for better code logic.

PR #403, opened 2024-01-19
fix: Add package name

owencraston:
## What
add package name to top level package.json in order to enable checking out specific commits from npm/yarn/pnpm. Previously if you were referencing a specific commit in your package.json it would fail with the following error. 
` ERR_PNPM_MISSING_PACKAGE_NAME  Can't install git+https://github.com/run-llama/LlamaIndexTS.git#1bd47969b37c86a226f10d1e1a6498de798ca39c: Missing package name`. By adding the name property to the top level package.json this error goes away. 

Screenshot 2024-01-19 at 12 07 07 AM

## Why
This enables developers to experiment with features that are not yet published to NPM.

## Testing
- I created a test app to demonstrate this. clone this repo and checkout this branch: https://github.com/owencraston/test-app/pull/1
- checkout this commit: https://github.com/owencraston/test-app/pull/1/commits/48541659e5d7373b69357868708e8990322913b6
- run pnpm install
- you should get the error:  `ERR_PNPM_MISSING_PACKAGE_NAME  Can't install git+https://github.com/run-llama/LlamaIndexTS.git#1bd47969b37c86a226f10d1e1a6498de798ca39c: Missing package name`
- then checkout this commit: https://github.com/owencraston/test-app/pull/1/commits/4787248eb953dd8c2c51763c5e0ee07a53ac6d36. This is the commit hash for this PR.
- run `pnpm install`. there should be no errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "llamaindex",
+  "name": "@llamaindex/monorepo",
   "private": true,
   "scripts": {
     "build": "turbo run build",
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "llamaindex",
   "private": true,
   "scripts": {
     "build": "turbo run build",


himself65 (commented):
We already have a package called `llamaindex`

marcusschiesser:
@owencraston have you tried `pnpm link` to fix your problem (testing features that are not on npm?) see https://github.com/run-llama/LlamaIndexTS/blob/main/examples/README.md#using-the-local-build

owencraston:
@marcusschiesser can this work in production?

owencraston:
@himself65 I still think we need to add A package name to the top level. this is a best practice with mono repos.

owencraston:
a better solution could be to have a scope: https://docs.npmjs.com/about-scopes
Then we could import from `@llamaindex/core`

himself65:
> a better solution could be to have a scope: [docs.npmjs.com/about-scopes](https://docs.npmjs.com/about-scopes) Then we could import from `@llamaindex/core`

yes, I'm thinking about that too, package name on top is a good sugession. will add that

yisding:
While "@llamaindex/core" might be slight more semantically correct in the future, it would add significantly more friction to the new user experience (not to mention breaking existing users' code). @owencraston could the same result be achieved if we just name the top level package.json "@llamaindex/gitrepo" ?

yisding (dismissed):
This is a good idea except for the name duplication. Let's name it `@llamaindex/gitrepo` or basically anything else. It's not getting published anyways, but we have the llamaindex scope, so anything in there would be safe.

yisding:
OK apparently with monorepos there's no great way to enable the kind of functionality we're looking for: https://stackoverflow.com/questions/39194648/npm-install-package-from-github-repo-subfolder I'm still not opposed to adding a name to the package.json if that fixes your particular use case, but it might not be something we want to support in general.

owencraston:
@yisding @himself65 I went ahead and changed the name to `@llamaindex/monorepo` following the example of this well maintained [repo](https://github.com/MetaMask/core/blob/main/package.json#L2).

himself65 (approved):
LGTM

PR #404, opened 2024-01-19
feat(extractors): add keyword extractor and base extractor

EmanuelCampos:


packages/core/src/extractors/MetadataExtractors.ts
```diff
@@ -0,0 +1,400 @@
...
+
+      title = completion.text;
+    }
+
+    if (nodesToExtractTitle.length === 1) {
+      title = titlesCandidates[0];
+    }
+
+    return nodes.map((_) => ({
+      documentTitle: title.trim().replace(STRIP_REGEX, ""),
```
yisding, line 194: What exactly is TitleExtractor supposed to do? I see here we're using the same title for all of the nodes?

packages/core/src/extractors/MetadataExtractors.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    return {
+      questionsThisExcerptCanAnswer: questions.text.replace(STRIP_REGEX, ""),
+    };
+  }
+
+  /**
+   * Extract answered questions from a list of nodes.
+   * @param {BaseNode[]} nodes Nodes to extract questions from.
+   * @returns {Promise<Record<string, any>[]>} Questions extracted from the nodes.
```
yisding, line 290: Please type the metadata explicitly instead of Record<string, any> (for example here, you can specify {questionsThisExcerptCanAnswer: string}). Will help users better understand what the various extractors are actually doing.

packages/core/src/extractors/prompts.ts
```diff
@@ -0,0 +1,52 @@
+export const defaultKeywordExtractorPromptTemplate = ({
+  context_str,
+  keywords,
+}: {
+  context_str: string;
+  keywords: number;
+}) => `
+  /${context_str}. Give ${keywords} unique keywords for this \
```
yisding, line 8: Unlike Python, the backslash doesn't work for JS template literals to continue the line. I would just write out the whole line, but if you really want you can use string concatenation.

Similarly the slash (forward slash) at the beginning of the template literal doesn't do want you want it to, which I'm guessing is to not have the newline.

packages/core/src/extractors/prompts.ts
```diff
@@ -0,0 +1,52 @@
...
+  document. Format as comma separated. Keywords:
+`;
+
+export const defaultTitleExtractorPromptTemplate = ({
+  context_str,
+}: {
+  context_str: string;
+}) => `
+  /${context_str}. Give a title that summarizes all of the unique entities, titles or themes found in the context. Title:
```
yisding, line 17: I would try to camel case if possible (if there are inconsistencies we can fix them). Also, if you look at the other prompts I've created, I think it's useful to create an "implied type" for each prompt so that the user can easily figure out what variables they need to handle if they're customizing the prompt.

packages/core/src/extractors/types.ts
```diff
@@ -0,0 +1,69 @@
+import { BaseNode, MetadataMode, TextNode } from "../Node";
+
+const DEFAULT_NODE_TEXT_TEMPLATE =
+  "\
+[Excerpt from document]\n{metadata_str}\n\
```
yisding, line 5: This will need to be a function/SimplePrompt/JS template literal to work properly. LangchainJS does support the conversion of Python f-strings into JS prompts, but I don't feel like the juice is worth the squeeze. If folks prefer to use something other than JS template literals, there's a wide variety of templating options in the JS ecosystem from handlebars to ejs that people can choose.

packages/core/src/extractors/prompts.ts
EmanuelCampos: Yes, we are using camelCase on almost all, just a nit, thanks for pointing out

packages/core/src/extractors/types.ts
EmanuelCampos: I ended up letting it go, thanks for pointing it out

packages/core/src/index.ts
```diff
@@ -15,6 +15,7 @@ export * from "./callbacks/CallbackManager";
 export * from "./constants";
 export * from "./embeddings";
 export * from "./engines/chat";
+export * from "./extractors";
```
himself65, line 18: Could we explicit export class from file instead of *.

I'm working on bundle splitting feature and this cause lots of confusion

packages/core/tsconfig.json
```diff
@@ -12,8 +12,14 @@
...
+  },
+  "ts-node": {
+    "files": true,
+    "compilerOptions": {
+      "module": "commonjs",
+    },
   },
   "include": ["./src"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules"],
```
himself65, line 24: Why there are commas here? Have you ran prettier?

himself65 (commented):
LGTM except some small things

packages/core/tsconfig.json
EmanuelCampos: @himself65 Yes!

`pnpm run format:write`

himself65 (commented):
Also you might need run npx changelog to write some update log

packages/core/tsconfig.json
EmanuelCampos: we can setup `trailingComma: "none"` on another PR, was setting here but around 200 file changes

yisding: Let's keep it as is. Makes sense.

packages/core/src/index.ts
EmanuelCampos: Maybe @yisding can jump in here as well, there's a PR that change. the another ones?

yisding: @himself65 if you want to do it, let's do it in a different PR.

The reason I started with `export *` was for convenience and to prevent us from accidentally forgetting to export something. (even so sometimes we forget to export from a particular file and get a bug report).

If there are bundling reasons why we want to explicitly re-export every single export from every single file, let's find a way to write a test case or lint rule because otherwise I guarantee you we will to export a bunch of things.

packages/core/tsconfig.json
```diff
@@ -14,6 +14,12 @@
     "target": "ES2015",
     "resolveJsonModule": true,
   },
+  "ts-node": {
+    "files": true,
```
himself65, line 18: just curious, Is this necessary? I didn't see the usage inside core pacakge

EmanuelCampos: @himself65 on the documentation is recommend to use ts-node to run the examples and do some tests, I wasn't able to use without this

himself65: I mean, this config is inside `package/core/tsconfig.json`, if you want run examples, should it put inside the top level?

EmanuelCampos: on the top level it didn't work, but will investigate why, thanks for pointing it!

himself65 (approved):
LGTM

packages/core/src/extractors/prompts.ts
```diff
@@ -0,0 +1,81 @@
...
+) => `
+  ${contextStr}. Give a title that summarizes all of the unique entities, titles or themes found in the context. Title:
+`;
+
+export const defaultTitleCombinePromptTemplate = (
+  { contextStr = "" }: DefaultPromptTemplate = {
+    contextStr: "",
+  },
+) => `
+  ${contextStr}. Based on the above candidate titles and content,s
```
yisding, line 41: @EmanuelCampos these prompts still need to be fixed. Can you open a new PR to do so?

PR #405, opened 2024-01-19
lint(prettier): add traillingComma

EmanuelCampos:


himself65 (approved):

yisding (changes requested):
Sorry, let's not do this.

himself65:
> Sorry, @himself65 @EmanuelCampos we haven't added any custom prettier rules to LITS up until now and I don't see a compelling reason to do it now.

it's okay. Keep the lint style same anyway

Issue #375, opened 2024-01-14
Embed batch size for OpenAIEmbedding


ericho-bbai:
In the python version, we could specify the `embed_batch_size` for OpenAIEmbedding to reduce the number of calls to OpenAI API. It would be great if we can do that in LlamaIndexTS to prevent rate limits.

marcusschiesser:
great idea, i added it to our backlog Thanks

PR #407, opened 2024-01-20
feat(embedding): add batch embed size

EmanuelCampos:
Solves #375


packages/core/src/embeddings/types.ts
```diff
@@ -21,4 +28,30 @@ export abstract class BaseEmbedding {
...
   abstract getTextEmbedding(text: string): Promise<number[]>;
   abstract getQueryEmbedding(query: string): Promise<number[]>;
+
+  async getTextEmbeddingBatch(texts: string[]): Promise<Array<Embedding>> {
+    const resultEmbeddings: Array<Embedding> = [];
+
+    const queue = texts;
+    let batchIndex = 0;
+
+    await BlueBird.map(
```
himself65, line 38: I think BlueBird here is not necessary?

You can do the same thing using Promise.all with array function

```ts
let queue = [];
let idx = 0;
for (const text of texts) {
  queue.push(this.getTextEmbedding(text).then(embedding => {
    console.log('xxx');
    idx++;
    return embedding;
  }));
}
return Promise.all(queue)
```

EmanuelCampos: @himself65 Promise.all would try to embed all the texts at once and then get rate limited, so that's why I like to use the Bluebird concurrency

EmanuelCampos: @himself65 applied some changes, did a bit different but you're right wouldn't be benefitial for now

packages/core/src/embeddings/types.ts
```diff
@@ -10,7 +10,13 @@ export enum SimilarityType {
   EUCLIDEAN = "euclidean",
 }
 
+type Embedding = number[];
```
marcusschiesser, line 13: good idea using a type for this, but then we should introduce it for all embeddings

packages/core/src/embeddings/types.ts
```diff
@@ -21,4 +27,32 @@ export abstract class BaseEmbedding {
 
   abstract getTextEmbedding(text: string): Promise<number[]>;
   abstract getQueryEmbedding(query: string): Promise<number[]>;
+
+  async getTextEmbeddingBatch(texts: string[]): Promise<Array<Embedding>> {
```
marcusschiesser, line 31: 1. How about having a default implementation in `BaseEmbedding` for the batching that just falls back and uses the non-batching method?
Then the consumers of `getTextEmbedding` can be changed to use batching generally - I think it's just `VectorStoreIndex`.
2. How about overloading `getTextEmbedding` instead?
`getTextEmbedding(texts: string[]): Promise<number[][]>;`

packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -16,6 +16,8 @@ export enum OpenAIEmbeddingModelType {
 export class OpenAIEmbedding extends BaseEmbedding {
   model: OpenAIEmbeddingModelType | string;
 
+  embedBatchSize: number;
```
marcusschiesser, line 19: doesn't use batching yet, needs to send one batch at once to openai api

packages/core/src/embeddings/types.ts
EmanuelCampos: @marcusschiesser nice catches! can you expand a bit on the second one?

> How about overloading getTextEmbedding instead?
getTextEmbedding(texts: string[]): Promise<number[][]>;

packages/core/src/embeddings/OpenAIEmbedding.ts
EmanuelCampos: @marcusschiesser the batching will be used when updated on `VectorStoreIndex`, correct?

marcusschiesser: Once `VectorStoreIndex` is updated to use batching, the batching method from `BaseEmbedding` will be used, yes.
But what I mean is that the `OpenAIEmbedding` implementation should use the batching support of the OpenAI Embedding API (see https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-input):
The `input` parameter supports passing a `string[]` to generate multiple embeddings at once. 
Most of the other APIs have similar support, but we can add this step-by-step in separate PRs.

packages/core/src/embeddings/types.ts
EmanuelCampos: I think for now it would case more breaking changes, and personally I think that's a bit confuse

packages/core/src/embeddings/OpenAIEmbedding.ts
EmanuelCampos: @marcusschiesser 

Thanks for the detailed explanation

I'm updating it now, wasn't ware of the batch on OpenAI side as well

EmanuelCampos: @EDIT tbh honest I think currently works on this way, it also can be manipulated by the chunk size, maybe we can later further discuss more bout it

yisding: To follow up on Marcus's comment here, what we're looking for in this feature is to probably use OpenAI/the underlying provdier's batching functionality.

EmanuelCampos: I can look into it, but the feature is to help avoid rate limiting, even if we fill the input[] with more text and send a lot of requests it will rate limit, so that's why the configurable embedBatchSize

EmanuelCampos: But for sure adding more text into the batch will make it faster and reduce the amounts of requests as well, so it's 2 parts of the feature

marcusschiesser: @EmanuelCampos you're right, these are two concerns, and we can split them into two PRs.

One advantage of doing it now is that you can factor out your `processChunk` function below on class level and then override `processChunk` here in the OpenAIEmbedding to send multiple strings in one request (a few lines of code).

You won't even need then here the `embedBatchSize`

packages/core/src/embeddings/types.ts
```diff
@@ -10,15 +10,55 @@ export enum SimilarityType {
...
+    const totalChunks = Math.ceil(queue.length / chunkSize);
+
+    const processChunk = async (chunk: string[], chunkIndex: number) => {
+      const embeddings = await Promise.all(
+        chunk.map(async (text) => {
+          return await this.getTextEmbedding(text);
+        }),
+      );
+
+      resultEmbeddings.push(...embeddings);
```
marcusschiesser, line 50: factor out this part (without the logging, can be kept here) as `protected processChunk` on the class level (see comment above)

this will then be the default chunk processor unless a subclass overrides the behavior

EmanuelCampos:
@marcusschiesser We opted in yesterday to accumulate to batch the text

Then for now I removed the old way wouldn't have many advantages now

So currently the batch size can be controlled by the embed batch size , making it more flexible to the user of how much text will be sent in batches finding a proportion that doesn't get rate limited and fills well the maximum tokens available

The same of the Python package which look to works well

cc: @yisding

packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -82,8 +86,8 @@ export class OpenAIEmbedding extends BaseEmbedding {
     return data[0].embedding;
   }
 
-  async getTextEmbedding(text: string): Promise<number[]> {
-    return this.getOpenAIEmbedding(text);
+  async getTextEmbedding(texts: string | string[]): Promise<number[]> {
```
marcusschiesser, line 89: looks great, but if `texts.length` > `embedBatchSize`, we might end up having more than `embedBatchSize` embeddings per request.

EmanuelCampos: > looks great, but if `texts.length` > `embedBatchSize`, we might end up having more than `embedBatchSize` embeddings per request.

the embedBatchSize is just for getTextEmbeddingBatch, it doesn't imply on how getTextEmbedding works, and I even think that it shouldn't because is not a max tokens or something like this

marcusschiesser: I guess it all depends on the contract we want to give to the user.

I assumed that the user could use any embedding (OpenAI, Mixtral, etc.) with this contract:

```
getTextEmbedding(texts: string | string[]): Promise<Embedding>; getQueryEmbedding(query: string | string[]): Promise<Embedding>;
 ```
 
Using this contract, the user can send a single string or an array.
 
And the user can optionally set a batch size.

Our code takes care to split the text into batches of the batch size, iterates over the batches, and then either
1. sends a whole batch to the embedding API (as supported by OpenAI and other providers) or
2. iterates over the batch entries and sends one request per entry

As long as we don't implement 1. for each provider, we could fallback to 2. using some code in the base class

This protects our users from sending too much data in one request.

The contract above uses overloading; we could alternatively use the same contract as in Python:
```
getTextEmbedding(text: string): Promise<Embedding>; 
getQueryEmbedding(query: string): Promise<Embedding>;
getTextEmbeddingsBatch(texts: string[]): Promise<Array<Embedding>>;
getQueryEmbeddingsBatch(queries: string[]): Promise<Array<Embedding>>;
 ```
 
Personally, I would prefer the first contract as it's easier for the user. @himself65 @yisding, what are your thoughts?

EmanuelCampos: I think we can keep it for another PR, we have nodeParsers that already solve a lot of split problems, and currently, we will add more complexity without much return, since we will need to count tokens from node text to fill batches correctly, and I don't see much benefits on overloading as well yet, I think it makes harder for adding new implementations since you'll always need to handle 2 different other cases, and would need to change the implementation on all current embeddings

EmanuelCampos: Obs: agree we can change later about splitting the text as well if needed, but now changing all it would be much more effort, and I think currently it will already help a lot.

packages/core/src/embeddings/types.ts
```diff
@@ -10,15 +10,60 @@ export enum SimilarityType {
...
+    embedding1: Embedding,
+    embedding2: Embedding,
     mode: SimilarityType = SimilarityType.DEFAULT,
   ): number {
     return similarity(embedding1, embedding2, mode);
   }
 
-  abstract getTextEmbedding(text: string): Promise<number[]>;
-  abstract getQueryEmbedding(query: string): Promise<number[]>;
+  abstract getTextEmbedding(textS: string | string[]): Promise<Embedding>;
```
marcusschiesser, line 31: @EmanuelCampos I agree to move forward with what we have. We don't want this PR to go stale :) 

To merge this PR, I suggest the following:

1. Use the same contract as in Python (will be less confusing for our users):
```
getTextEmbedding(text: string): Promise<Embedding>; 
getQueryEmbedding(query: string): Promise<Embedding>; 
getTextEmbeddings(texts: string[]): Promise<Embedding>; 
getTextEmbeddingsBatch(texts: string[]): Promise<Array<Embedding>>; 
```
Like in Python, the default implementation of `getTextEmbeddings` just iterates over all strings and calls `getTextEmbedding` - subclasses can override this behavior and send multiple strings in one request (like your OpenAI version).  

2. Either use `Embedding` type consistently for all embeddings, or remove introducing this type (It's a good idea, but it should be everywhere if we do it)

packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -110,8 +114,8 @@ export class OpenAIEmbedding extends BaseEmbedding {
...
```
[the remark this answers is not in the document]
EmanuelCampos: why override? since getTextEmbedding does the job and calls getTextEmbedding to call getOpenAIEmbedding

packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -32,6 +32,8 @@ export class OpenAIEmbedding extends BaseEmbedding {
   /** number of dimensions of the resulting vector, for models that support choosing fewer dimensions. undefined will default to model default */
   dimensions: number | undefined;
 
+  embedBatchSize: number;
```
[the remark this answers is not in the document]
marcusschiesser: but it's already in `BaseEmbedding`

packages/core/src/embeddings/OpenAIEmbedding.ts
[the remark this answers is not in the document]
marcusschiesser: because `getTextEmbeddingsBatch` is calling `getTextEmbeddings`, and its default implementation iterates over all items in a batch.

packages/core/src/embeddings/OpenAIEmbedding.ts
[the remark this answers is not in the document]
EmanuelCampos: right, but on the constructor we still need to give the user option to update the parameters

packages/core/src/embeddings/OpenAIEmbedding.ts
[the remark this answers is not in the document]
EmanuelCampos: `getTextEmbeddings` is calling `getTextEmbedding` and each embedding implementation currently needs to have `getTextEmbedding` which provides access to the client (openai, mistral) while  `getTextEmbeddings` and `getTextEmbeddingsBatch` is responsible to do it in batches

marcusschiesser (approved):

marcusschiesser:
Thanks @EmanuelCampos ! Looking good to merge, just fix the conflict please

Issue #410, opened 2024-01-20
[RFC]: Add nodes paramters on fromVectorStore


EmanuelCampos:
## Problem Statement

Currently, we aren't able to index nodes directly, just through a parser

Example case:

```py
const nodes = new SimpleNodeParser().getNodesFromDocuments(documents)

  for (let i = 0; i < nodes.length; i++) {
    nodes[i].id_ = i.toString();
  }
    
  // Split text and create embeddings. Store them in a VectorStoreIndex
  const index = await VectorStoreIndex.fromVectorStore(nodes, {
    vectorStore,
  })
  ```
  
  ## Possible solution
  Would be interesting maybe on fromVectorStore if we add the nodes parameter?

PR #411, opened 2024-01-20
[WIP]: Add nodes parameter on fromVectorStore

EmanuelCampos:
it have a breaking change, if we put as last parameter would make imageVectorStore required or any workaround to use it

https://github.com/run-llama/LlamaIndexTS/issues/410


EmanuelCampos:
bump @yisding @marcusschiesser @himself65

EmanuelCampos:
closing we have other ways to follow witth this as `init` or insertNodes

PR #412, opened 2024-01-21
fix(metadataFiltering): prefilters not being passed to vector query

EmanuelCampos:


marcusschiesser (commented):
@EmanuelCampos LGTM - do you see any breaking changes, e.g. with using other VectorDBs?

EmanuelCampos:
@marcusschiesser Yes! there's no breaking changes, before that wasn't working

yisding:
Thanks for fixing this bug @EmanuelCampos !

PR #416, opened 2024-01-21
feat: OpenAI agent

EmanuelCampos:


apps/docs/docs/examples/agent.mdx
```diff
@@ -0,0 +1,85 @@
...
+  const sumFunctionTool = new FunctionTool(sumNumbers, {
+    name: "sumNumbers",
+    description: "Use this function to sum numbers together",
+    parameters: sumJSON,
+  });
+
+  // Create a function tool from the divide function
+  const divideFunctionTool = new FunctionTool(divideNumbers, {
+    name: "divideNumbers",
+    description: "Use this function to divide numbers together",
```
yisding, line 63: This description a bit awkward. I would say something like `"divide two numbers"`

apps/docs/docs/examples/agent.mdx
```diff
@@ -0,0 +1,85 @@
...
+  return a / b;
+}
+
+// Define the parameters of the sum function as a JSON schema
+const sumJSON = {
+  type: "object",
+  properties: {
+    a: {
+      type: "number",
+      description: "The argument a to sum",
```
yisding, line 26: here, I would say something like `"the first number"`

apps/docs/docs/examples/agent.mdx
```diff
@@ -0,0 +1,85 @@
...
+  required: ["a", "b"],
+};
+
+// Define the parameters of the divide function as a JSON schema
+const divideJSON = {
+  type: "object",
+  properties: {
+    a: {
+      type: "number",
+      description: "The argument a to divide",
```
yisding, line 42: Here I would say `"the dividend"` and `"the divisor"` below

examples/agent/openai.ts
```diff
@@ -0,0 +1,71 @@
...
+    b: {
+      type: "number",
+      description: "The argument b to sum",
+    },
+  },
+  required: ["a", "b"],
+};
+
+// Define the parameters of the divide function as a Zod schema
+const dividArgsSchema = z
```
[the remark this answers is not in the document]
yisding: Was your point Marcus that zodToJsonSchema would have zod as a dependency? Unfortunately it doesn't (only as a devDependency)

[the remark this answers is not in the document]
yisding: OK, I kind of see Marcus's point. The only place we actually use zod in the library is the `isZodSchema` function, which isn't actually being called and toolParametersSchema, which doesn't need to be defined as a zod schema. So zod could be a dev dependency once we fix those two up. But not a blocker for merging.

packages/core/src/agent/openai/base.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ChatMessage, OpenAI } from "../../llm";
+import { ObjectRetriever } from "../../objects/base";
+import { BaseTool } from "../../types";
+import { AgentRunner } from "../runner/base";
+import { OpenAIAgentWorker } from "./worker";
+
+type OpenAIAgentParams = {
+  tools: BaseTool[];
+  llm?: OpenAI;
+  memory?: any;
```
yisding, line 11: Not a fan of the any here, but I know we have the memory stuff to fix up.

packages/core/src/agent/openai/utils.ts
```diff
@@ -0,0 +1,35 @@
+import { ZodSchema } from "zod";
+import { ToolMetadata } from "../../types";
+import { getProperties } from "../utils";
+
+export type OpenAiFunction = {
```
yisding, line 5: Let's just use the all uppercase `AI` in our class names. Yes, it's inconsistent with the way we do `Id` but too late/not enough reason to change now I think.

apps/docs/docs/modules/agent/openai.mdx
```diff
@@ -0,0 +1,215 @@
...
+  name: "divideNumbers",
+  description: "Use this function to divide numbers together",
+  parameters: divideJSON,
+});
+```
+
+### Zod Schema
+
+```ts
+import { z } from "zod";
```
yisding, line 83: Blocker: so I think the whole point of using zod here is we want to show that you can define a function using zod arguments one time, and then having it work "automatically." Here, we're defining the zod schema separately from the underlying function, which defeats the purpose.

packages/core/src/agent/openai/worker.ts
```diff
@@ -0,0 +1,403 @@
+// Assuming that the necessary interfaces and classes (like BaseTool, OpenAI, ChatMessage, CallbackManager, etc.) are defined elsewhere
+
+import { randomUUID } from "crypto";
```
yisding, line 3: Blocker: This is a NodeJS specific import correct? If so we should import from `"node:crypto"` to not break the other bundles.

apps/docs/docs/modules/agent/openai.mdx
```diff
@@ -0,0 +1,215 @@
...
+  .describe("the arguments");
+
+const dividArgsSchema = z
+  .object({
+    a: z.number().describe("The argument a to divide"),
+    b: z.number().describe("The argument b to divide"),
+  })
+  .describe("the arguments");
+
+const sumFunctionTool = new FunctionTool(sumNumbers, {
```
yisding, line 99: So for example, sumNumbersZod could be a zod function schema https://zod.dev/?id=functions with a `.implement`, and then FunctionTool should only need to take in the single argument.

packages/core/src/engines/chat/types.ts
```diff
@@ -52,3 +57,32 @@ export interface Context {
 export interface ContextGenerator {
   generate(message: string, parentEvent?: Event): Promise<Context>;
 }
+
+export enum ChatResponseMode {
+  WAIT = "wait",
+  STREAM = "stream",
+}
+
+export class AgentChatResponse {
```
yisding, line 66: Why do we need the new class and enum here?

packages/core/src/storage/chatStore/SimpleChatStore.ts
```diff
@@ -0,0 +1,92 @@
+import { ChatMessage } from "../../llm";
+import { BaseChatStore } from "./types";
+
+/**
+ * Simple chat store.
+ */
+export class SimpleChatStore implements BaseChatStore {
```
yisding, line 7: Not a blocker: So this class feels like a bit unnecessary in JS. Not sure it adds anything particularly useful over just the underlying data structure.

packages/core/src/tools/functionTool.ts
```diff
@@ -0,0 +1,35 @@
+import { ZodSchema } from "zod";
+import { BaseTool, ToolMetadata } from "../types";
+
+type IfZodSchema<T, True, False> = T extends ZodSchema ? True : False;
+
+type Metadata<T = any> = {
```
yisding, line 6: We should pick a different name for this type.

yisding (commented):
Thanks for doing this. Left some comments!

yisding (changes requested):
@EmanuelCampos's working on fixing a bug. Don't merge yet.

packages/core/src/llm/LLM.ts
```diff
@@ -203,6 +203,23 @@ export class OpenAI extends BaseLLM {
     }
   }
 
+  toOpenAIMessage(messages: ChatMessage[]) {
+    return messages.map((message) => {
+      const additionalKwargs = message.additionalKwargs ?? {};
+
+      if (message.additionalKwargs?.toolCalls) {
```
yisding, line 210: I don't like kwargs as a name because that's a python (not JS) concept, but we can fix that in another PR.

PR #419, opened 2024-01-23
fix: spawn run fail on window

thucpn:


packages/create-llama/helpers/run-app.ts
```diff
@@ -48,30 +59,32 @@ export async function runApp(
...
       });
     }).catch((err) => {
       console.error(err);
       killAppProcesses();
     });
   } else {
     return new Promise((resolve, reject) => {
-      backendAppProcess = spawn(backendCommand, backendArgs, {
+      // eslint-disable-next-line turbo/no-undeclared-env-vars
+      console.log({ PATH: process.env.PATH });
```
himself65, line 81: remove console.log

packages/create-llama/helpers/run-app.ts
```diff
@@ -64,14 +78,11 @@ export async function runApp(
     });
   } else {
     return new Promise((resolve, reject) => {
-      backendAppProcess = spawn(backendCommand, backendArgs, {
+      backendAppProcess = createProcess(backendCommand, backendArgs, {
         stdio: "inherit",
-        cwd: appPath,
+        cwd: path.join(appPath, "backend"),
```
marcusschiesser, line 83: @thucpn @leehuwuj this looks wrong as this is the case where the app is just in one folder

packages/create-llama/helpers/run-app.ts
```diff
@@ -1,8 +1,22 @@
...
 import { TemplateFramework } from "./types";
 
+const createProcess = (
+  command: string,
+  args: string[],
+  options: SpawnOptions,
+) => {
+  return spawn(command, args, {
+    ...options,
+    shell: /^win/.test(process.platform),
```
marcusschiesser, line 13: @leehuwuj @thucpn We didn't use `shell` in `callpackageManager` (and it works on windows):

https://github.com/run-llama/LlamaIndexTS/blob/656bb56e91d0c6f51f61a1b97f572cbbdd197609/packages/create-llama/helpers/install.ts#L31-L41

another difference is that it's using
`process.chdir(props.root);` instead of the `cwd` argument - maybe that has an error on windows? or the path is not resolved correctly?

marcusschiesser (approved):

PR #430, opened 2024-01-24
feat: create-llama - add local pdf file option

leehuwuj:
Introduce a new option: `Select another PDF file` to select another pdf file located in local machine. (**Support for Windows and MacOS**)
```
? Which data source would you like to use? › - Use arrow-keys. Return to submit.
   No data, just a simple chat
   Use an example PDF
❯  Select another PDF file
```
Flow:
- If user chose `Select another PDF file` then:
+ Show a popup to select a file (macos native window)
+ User only allowed to choose a file. 
    1. If it's a PDF file, then go to the next question. 
    2. If they cancel the selection, we will exit the app. 
    3. If the file is not a PDF file, we will show an error message and exit the app.

### DIFF
--- a/.changeset/stale-geckos-bathe.md
+++ b/.changeset/stale-geckos-bathe.md
@@ -2,4 +2,4 @@
 "create-llama": patch
 ---
 
-add a new option to select local PDF file
+Add an option to select a local PDF file as data source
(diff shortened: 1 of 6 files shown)


.changeset/stale-geckos-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+add a new option to select local PDF file
```
marcusschiesser, line 5:
```suggestion
Add an option to select a local PDF file as data source
```

marcusschiesser (approved):

PR #439, opened 2024-01-25
docs: add stackblitz playground

himself65:


examples/README.md
```diff
@@ -1,29 +1,19 @@
...
+export OPENAI_API_KEY="sk-..."
 
-```bash
-npm install
+npx ts-node ./chatEngine.ts
 ```
 
-Then run the examples with `ts-node`, for example `npx ts-node vectorIndex.ts`
-
-## Using the local build
```
marcusschiesser, line 21: @himself65 why did you remove how to use the local build?

himself65: No you don’t need that if you’re using pnpm. It will automatically use the monorepo I believe

himself65: https://pnpm.io/workspaces#workspace-protocol-workspace

```
By default, pnpm will link packages from the workspace if the available packages match the declared ranges. For instance, foo@1.0.0 is linked into bar if bar has "foo": "^1.0.0" in its dependencies and foo@1.0.0 is in the workspace. However, if bar has "foo": "2.0.0" in dependencies and foo@2.0.0 is not in the workspace, foo@2.0.0 will be installed from the registry. This behavior introduces some uncertainty.
```

marcusschiesser: In my tests this only worked if you set `"llamaindex": "workspace:*"`, but I set it on purpose to `"llamaindex": "latest"` so it works without pnpm (for our users).

PR #440, opened 2024-01-25
minor bug fixes with together AI

yisding:
Fixed small bugs with the Together.AI integration.
* the init params were being overriden, so for example if the user set an apiKey in the init it would be overriden by the environment variable.
* the example was missing a newline in the output
* if there was no key specified, the user would be prompted for an OpenAI API key
* the OpenAI embeddings interface didn't allow model specification (which didn't matter when there was only one supported model)

### DIFF
--- a/.changeset/tender-horses-rhyme.md
+++ b/.changeset/tender-horses-rhyme.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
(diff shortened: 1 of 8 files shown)


yisding:
@marcusschiesser @himself65 one thing we need to think about with the TogetherAI integration is how we want to handle the context window size per model, the tokenizer per model, and the vector dimensionality per model.

PR #445, opened 2024-01-25
Openai embeddings 3

yisding:
Also supporting the new GPT-4 preview model

### DIFF
--- a/packages/core/src/embeddings/OpenAIEmbedding.ts
+++ b/packages/core/src/embeddings/OpenAIEmbedding.ts
@@ -27,25 +27,37 @@ export const ALL_OPENAI_EMBEDDING_MODELS = {
 };
 
 export class OpenAIEmbedding extends BaseEmbedding {
+  /** embeddding model. defaults to "text-embedding-ada-002" */
   model: string;
+  /** number of dimensions of the resulting vector, for models that support choosing fewer dimensions. undefined will default to model default */
   dimensions: number | undefined;
 
   // OpenAI session params
+
+  /** api key */
   apiKey?: string = undefined;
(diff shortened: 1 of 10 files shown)


seldo (approved):
LGTM

packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -30,7 +44,8 @@ export class OpenAIEmbedding extends BaseEmbedding {
   constructor(init?: Partial<OpenAIEmbedding> & { azure?: AzureOpenAIConfig }) {
     super();
 
-    this.model = init?.model ?? OpenAIEmbeddingModelType.TEXT_EMBED_ADA_002;
+    this.model = init?.model ?? "text-embedding-ada-002";
+    this.dimensions = init?.dimensions;
```
EmanuelCampos, line 48: backward compatibility’s would be good to have a default dimensions as well

yisding: Good point. So the way this works is the dimensions is undefined if the user doesn't specify so it automatically uses the default dimensions of the model (by not sending the parameter) but agreed this is not obvious just from the code. Will add some comments to make this clearer.

himself65 (approved):
lgtm

packages/core/src/embeddings/OpenAIEmbedding.ts
yisding: https://github.com/run-llama/LlamaIndexTS/pull/445/commits/f27fd93b881c6ecbdc257d1286724f0faf8ee0c4 Really appreciate the feedback Emanuel!

PR #448, opened 2024-01-26
Update packages

yisding:
Updated packages, except for some integrations that we should do some more testing for:

```
❯ pnpm outdated
┌─────────────────────────────┬─────────┬─────────┐
│ Package                     │ Current │ Latest  │
├─────────────────────────────┼─────────┼─────────┤
│ pdfjs-dist                  │ 4.0.269 │ 4.0.379 │
├─────────────────────────────┼─────────┼─────────┤
│ chromadb                    │ 1.7.3   │ 1.8.1   │
├─────────────────────────────┼─────────┼─────────┤
│ @pinecone-database/pinecone │ 1.1.3   │ 2.0.1   │
├─────────────────────────────┼─────────┼─────────┤
│ @types/node (dev)           │ 18.19.9 │ 20.11.6 │
├─────────────────────────────┼─────────┼─────────┤
│ file-type                   │ 18.7.0  │ 19.0.0  │
├─────────────────────────────┼─────────┼─────────┤
│ notion-md-crawler           │ 0.0.2   │ 0.0.8   │
├─────────────────────────────┼─────────┼─────────┤
│ portkey-ai                  │ 0.1.16  │ 1.1.3   │
└─────────────────────────────┴─────────┴─────────┘
```


packages/core/src/types/mistral.d.ts
```diff
@@ -1,3 +0,0 @@
-declare module "@mistralai/mistralai" {
```
yisding, line 1: @marcusschiesser I think they fixed their TSDs, but can you verify that this is no longer needed?

marcusschiesser: yes, looks good, i run our mistral example

PR #450, opened 2024-01-26
feat: Add chat with web data

leehuwuj:
Add a new option to chat with web data (only available in fastapi backend), use [WholeSite](https://llamahub.ai/l/web-whole_site) data loader from LlamaHub to crawl and load website data.


packages/create-llama/helpers/index.ts
```diff
@@ -168,18 +179,23 @@ export const installTemplate = async (
...
-      props.packageManager,
-      props.engine,
-      props.openAiKey,
-      props.vectorDb,
-      props.contextFile,
-    );
+    if (props.engine === "context") {
+      if (props.dataSource === "file") {
+        // Copy test pdf file
+        // let dataSourceConfig = props.dataSourceConfig as FileSourceConfig;
```
marcusschiesser, line 189: assert instead of comment? (or leave it)

packages/create-llama/helpers/types.ts
```diff
@@ -6,6 +6,15 @@ export type TemplateEngine = "simple" | "context";
 export type TemplateUI = "html" | "shadcn";
 export type TemplateVectorDB = "none" | "mongo" | "pg";
 export type TemplatePostInstallAction = "none" | "dependencies" | "runApp";
+export type TemplateDataSource = "none" | "file" | "web";
```
marcusschiesser, line 9: @leehuwuj good idea to introduce the datasource type! How about putting type and config in one object? (to simplify parameter passing)
```suggestion
export type TemplateDataSource = {
  type: "none" | "file" | "web";
  config: TemplateDataSourceConfig;
}  
```

packages/create-llama/templates/components/vectordbs/python/mongo/index.py
```diff
@@ -20,4 +20,4 @@ def get_chat_engine():
     )
     index = VectorStoreIndex.from_vector_store(store, service_context)
     logger.info("Finished connecting to index from MongoDB.")
-    return index.as_chat_engine(similarity_top_k=5)
+    return index.as_chat_engine(chat_mode="context", similarity_top_k=5)
```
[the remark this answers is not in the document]
leehuwuj: in my experience, chat_mode="context" gives better result than default one, especially we only apply theses change for vectordb code where user did choose to chat with context before.

packages/create-llama/helpers/types.ts
leehuwuj: great suggestion! I'll apply this one. Thank you!

packages/create-llama/helpers/types.ts
```diff
@@ -15,8 +27,8 @@ export interface InstallTemplateArgs {
   template: TemplateType;
   framework: TemplateFramework;
   engine: TemplateEngine;
-  contextFile?: string;
   ui: TemplateUI;
+  dataSource: TemplateDataSource;
```
marcusschiesser, line 31: @leehuwuj did you test this with llamapacks and community templates? pretty sure dataSource would then be undefined and your code might crash

packages/create-llama/templates/components/vectordbs/python/mongo/index.py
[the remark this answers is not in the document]
marcusschiesser: AFAIK the default chat mode is `BEST` - so if `BEST` is not the best, it seems like a llama_index issue that should not be handled here.

packages/create-llama/helpers/types.ts
leehuwuj: @marcusschiesser they are working well, but you are right, we may have undefined dataSource so let me change it to optional attribute

packages/create-llama/questions.ts
```diff
@@ -389,6 +396,11 @@ export const askQuestions = async (
         },
         handlers,
       );
+      // Initialize with default config
+      program.dataSource = {
```
marcusschiesser, line 400: @leehuwuj this is sufficient as you already define the default above:
```suggestion
      // Initialize with default config
      program.dataSource = getPrefOrDefault("dataSource");
```

marcusschiesser (approved):

marcusschiesser (approved):

PR #451, opened 2024-01-26
build(core): release with node resolution compatibility

himself65:
idea from: https://github.com/pmndrs/jotai/blob/main/package.json


packages/core/package.json
```diff
@@ -63,6 +76,10 @@
     "lint": "eslint .",
     "test": "jest",
     "build": "bunchee",
+    "postbuild": "pnpm run copy && pnpm run modify-package-json",
+    "copy": "cp -r package.json CHANGELOG.md ../../README.md ../../LICENSE examples src dist/",
+    "modify-package-json": "node ./scripts/modify-package-json.mjs",
+    "prepublish": "echo \"please cd ./dist and run pnpm publish\" && exit 1",
```
marcusschiesser, line 82: @himself65 does this work with changeset publish?

packages/core/package.json
```diff
@@ -63,6 +76,10 @@
     "lint": "eslint .",
     "test": "jest",
     "build": "bunchee",
+    "postbuild": "pnpm run copy && pnpm run modify-package-json",
+    "copy": "cp -r package.json CHANGELOG.md ../../README.md ../../LICENSE examples src dist/",
```
marcusschiesser, line 80: @himself65 why can't we use the `package.json` of the core package directly? As this is uncommon, can you document it?

packages/core/package.json
himself65: yes

packages/core/package.json
himself65: OK I will explain details here, for the people who might don't understand in the future.

previously our package structure is like this


```
dist
	./index.js
	./env.js
package.json
```

for the people who want to import like `llamaindex/env`, in the nodejs >=16, we can use conditional exports

```json
{
	"exports": {
		// ...
		"./a": {
			"types": "./dist/a.d.ts"
			"default": "./dist/a.js"
		}
	}
}
```

But, typescript have four module resolution mode https://www.typescriptlang.org/tsconfig#moduleResolution

If you want to use this with type-safe, you should set `node16` or `bundler`(currently they are the same), and typescript will go check `exports` fields and everything will be okay.

BUT, the default option is `node10`/`node`, typescript won't check that, and it will treat `llamaindex/env` as `llamaindex/env.js` and go check `node_modules/llamaindex/env.d.ts` and previously output are inside `dist`, so this will make type broken (even you force run it and runtime is okay)

So, that's why I do this PR

himself65: Here is full list of compatibility of the release way

https://github.com/NWYLZW/How-to-publish-a-npm-package

and the conclusion is `directory is fully compatible, it is absolutely possible in any user usage scenario.`

PR #461, opened 2024-01-27
Remove unused arg in qdrant docs

iansinnott:
Going through the docs I noticed a minor issue. It seems there is no `port` parameter, only `url` which can optionally include the port number.


EmanuelCampos:
@iansinnott nice catch, can you fix the conflict?

himself65 (approved):

PR #466, opened 2024-01-27
update CONTRIBUTING.md for e2e test

stmtk1:


stmtk1:
I was wondering if I should add .python-version to my .gitignore or commit .python-version. Which do you think is better?

CONTRIBUTING.md
```diff
@@ -78,3 +78,13 @@ pnpm start
 That should start a webserver which will serve the docs on https://localhost:3000
 
 Any changes you make should be reflected in the browser. If you need to regenerate the API docs and find that your TSDoc isn't getting the updates, feel free to remove apps/docs/api. It will automatically regenerate itself when you run pnpm start again.
+
+### e2e test
```
himself65, line 82:
```suggestion
### E2E Test
```

himself65 (approved):

CONTRIBUTING.md
```diff
@@ -78,3 +78,13 @@ pnpm start
 That should start a webserver which will serve the docs on https://localhost:3000
 
 Any changes you make should be reflected in the browser. If you need to regenerate the API docs and find that your TSDoc isn't getting the updates, feel free to remove apps/docs/api. It will automatically regenerate itself when you run pnpm start again.
+
+### E2E Test
```
marcusschiesser, line 82: as the e2e tests are part of create-llama, i think we should add a create-llama section here first

stmtk1: There is no CONTRIBUTE.md in create-llama. So you wanna add CONTRIBUTE.md to create-llama directory, and write these sentence to the file. Right?

marcusschiesser: @seldo, how do you think about handling this? We could probably add a CONTRIBUTING.md to create-llama (then we can better split out create-llama in a separate repo if needed).

seldo: That sounds right to me but I think this is @yisding's call.

marcusschiesser:
Thanks @stmtk1 closing this for now, will create a own contributing.md for https://github.com/run-llama/create-llama

PR #479, opened 2024-01-30
chore: split readers into different files

himself65:


packages/core/package.json
```diff
@@ -62,6 +62,56 @@
       "import": "./dist/env.mjs",
       "edge-light": "./dist/env.edge-light.mjs",
       "require": "./dist/env.js"
+    },
+    "./readers/AssemblyAIReader": {
```
marcusschiesser, line 66: @himself65, what happens if we add a new reader and forget to mention it here? I assume it will be added to the main bundle.

marcusschiesser: if that's the case i think it's good to merge. Thanks @himself65 !

.github/workflows/test.yml
```diff
@@ -32,10 +32,35 @@ jobs:
...
+        run: pnpm run build --filter llamaindex
+      - name: Copy examples
+        run: rsync -rv --exclude=node_modules ./examples ${{ runner.temp }}
+      - name: Pack
+        run: pnpm pack --pack-destination ${{ runner.temp }}
+        working-directory: packages/core
+      - name: Install llamaindex
+        run: npm add ${{ runner.temp }}/*.tgz
+        working-directory: ${{ runner.temp }}/examples
+      - name: Run Type Check
```
marcusschiesser, line 64: great job @himself65 !

packages/core/package.json
himself65: > it will be added to the main bundle.

Yes

PR #481, opened 2024-01-30
Feat: add + fix chat with web

leehuwuj:


packages/create-llama/helpers/index.ts
```diff
@@ -108,9 +92,14 @@ const copyTestData = async (
     const hasOpenAiKey = openAiKey || process.env["OPENAI_API_KEY"];
     const hasVectorDb = vectorDb && vectorDb !== "none";
     if (framework === "fastapi") {
-      if (hasOpenAiKey && vectorDb === "none" && isHavingPoetryLockFile()) {
+      if (hasOpenAiKey && !hasVectorDb && isHavingPoetryLockFile()) {
```
leehuwuj, line 95: Already have hasVectorDb variable, just reuse it

packages/create-llama/helpers/index.ts
```diff
@@ -174,18 +188,28 @@ export const installTemplate = async (
...
+        if ("contextFile" in props.dataSource.config) {
+          await copyContextData(
+            props.root,
+            props.dataSource.config.contextFile,
+          );
+        } else {
+          await copyContextData(props.root);
+        }
+      }
+      await installDependencies(
```
leehuwuj, line 205: We missed calling `await` for installDependencies that why the error there were not caught.

packages/create-llama/index.ts
```diff
@@ -278,7 +278,6 @@ async function notifyUpdate(): Promise<void> {
           "\n",
       );
     }
-    process.exit();
```
leehuwuj, line 281: this function is triggered once we have error but it exits without error code

packages/create-llama/helpers/index.ts
marcusschiesser: that's a nasty bug, can you make a new PR adding https://eslint.org/docs/latest/rules/require-await to eslint to avoid this in the future

packages/create-llama/helpers/index.ts
```diff
@@ -174,18 +188,26 @@ export const installTemplate = async (
...
-    );
+    if (props.engine === "context") {
+      if (
+        props.dataSource?.type === "file" &&
+        "contextFile" in props.dataSource.config
+      ) {
+        await copyContextData(props.root, props.dataSource.config.contextFile);
+      } else {
+        await copyContextData(props.root);
+      }
```
marcusschiesser, line 202: @leehuwuj not a big think (for next time):
probably easier to read if you deal with this logic in `copyContextData`:

```suggestion
      await copyContextData(props.root, props.dataSource);
```

marcusschiesser (approved):

PR #484, opened 2024-01-30
feat(router): setup router query engine

EmanuelCampos:


yisding (commented):
Thanks! A few notes (halfway through the review).

packages/core/src/selectors/llmSelectors.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  return { selections };
+}
+
+type LLMPredictorType = LLM;
+
+/**
+ * A selector that uses the LLM to select a single or multiple choices from a list of choices.
+ */
+export class LLMMultiSelector extends BaseSelector {
```
himself65, line 37: Could you keep index.ts only have `exports *`

packages/core/src/selectors/prompts.ts
```diff
@@ -0,0 +1,28 @@
...
+  queryStr: string,
+  maxOutputs: number,
+) => {
+  return `Some choices are given below. It is provided in a numbered list (1 to ${numChoices}),
+where each item in the list corresponds to a summary.
+---------------------
+${contextList}
+---------------------
+Using only the choices above and not prior knowledge, return the top choices (no more than ${maxOutputs}, but only select what is needed) that are most relevant to the question: '${queryStr}'
+`;
```
himself65, line 27: Where does this prompt come from?

packages/core/tsconfig.json
```diff
@@ -14,6 +14,12 @@
     "target": "ES2015",
     "resolveJsonModule": true,
   },
+  "ts-node": {
```
himself65, line 17: How would you have to run ts-node on this package?

I guess if you do this it will no need for this

```
pnpm run dev # on packages/core
npx ts-node ./xxx.ts # on examples
```

packages/core/src/selectors/prompts.ts
EmanuelCampos: https://github.com/run-llama/llama_index/blob/main/llama_index/selectors/prompts.py#L26-L35

packages/core/src/QueryEngine.ts
```diff
@@ -206,3 +209,152 @@ export class SubQuestionQueryEngine implements BaseQueryEngine {
...
+        const finalResponse = await combineResponses(
+          this._summarizer,
+          responses,
+          queryBundle,
+        );
+        return finalResponse;
+      } else {
+        return responses[0];
+      }
+    } else {
```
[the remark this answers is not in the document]
EmanuelCampos: these are different checks, one is for tools selections other one is for how much responses and if need to be combined

packages/core/src/selectors/llmSelectors.ts
EmanuelCampos: @himself65 What do you mean by keep index.ts only? rename the llmSelectors file?

packages/core/src/selectors/base.ts
```diff
@@ -0,0 +1,74 @@
...
+    const metadatas = choices.map((choice) => wrapChoice(choice));
+    const queryBundle = wrapQuery(query);
+    return await this._select(metadatas, queryBundle);
+  }
+
+  abstract _select(
+    choices: ToolMetadata[],
+    query: QueryBundle,
+  ): Promise<SelectorResult>;
+  abstract asQueryComponent(): unknown;
```
marcusschiesser, line 73: why do  we need this function?

packages/core/src/selectors/llmSelectors.ts
marcusschiesser, line 37: isn't that the same class as `LLMSingleSelector` with a different prompt?

packages/core/src/selectors/llmSelectors.ts
```diff
@@ -0,0 +1,162 @@
...
+  const texts: string[] = [];
+  for (const [ind, choice] of choices.entries()) {
+    let text = choice.description.split("\n").join(" ");
+    text = `(${ind + 1}) ${text}`; // to one indexing
+    texts.push(text);
+  }
+  return texts.join("");
+}
+
+function _structuredOutputToSelectorResult(output: any): SelectorResult {
```
marcusschiesser, line 20:
```suggestion
function _structuredOutputToSelectorResult(output: StructuredOutput<Answer[]>): SelectorResult {
```

packages/core/src/selectors/base.ts
```diff
@@ -0,0 +1,74 @@
+import { QueryBundle, ToolMetadata } from "../types";
+
+export class SingleSelection {
+  index!: number;
+  reason!: string;
+}
+
+export class MultiSelection {
```
marcusschiesser, line 8: is this used somewhere?

packages/core/src/selectors/base.ts
```diff
@@ -0,0 +1,74 @@
+import { QueryBundle, ToolMetadata } from "../types";
+
+export class SingleSelection {
```
marcusschiesser, line 3: class is not initiated somewhere - use interface or type?

examples/routerQueryEngine.ts
```diff
@@ -0,0 +1,86 @@
...
+  // Create a router query engine
+  const queryEngine = RouterQueryEngine.fromDefaults({
+    selector: new LLMSingleSelector({
+      llm: serviceContext.llm,
+    }),
+    queryEngineTools: [
+      {
+        queryEngine: vectorQueryEngine,
+        metadata: {
+          name: "vector_query_engine",
```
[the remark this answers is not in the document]
EmanuelCampos: agreed, updated

examples/routerQueryEngine.ts
```diff
@@ -0,0 +1,86 @@
...
+          description: "Useful for summarization questions related to Abramov",
```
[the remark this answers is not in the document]
EmanuelCampos: agreed, updated

examples/routerQueryEngine.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  // Create query engines
+  const vectorQueryEngine = vectorIndex.asQueryEngine();
+  const summaryQueryEngine = summaryIndex.asQueryEngine();
+
+  // Create a router query engine
+  const queryEngine = RouterQueryEngine.fromDefaults({
+    selector: new LLMSingleSelector({
+      llm: serviceContext.llm,
+    }),
```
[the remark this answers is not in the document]
EmanuelCampos: agreed, updated

packages/core/src/QueryEngine.ts
```diff
@@ -6,14 +6,17 @@ import { ServiceContext, serviceContextFromDefaults } from "./ServiceContext";
 import { Event } from "./callbacks/CallbackManager";
```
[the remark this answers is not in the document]
EmanuelCampos: agreed, did the changes, just not moved all types since it's shared with a lot of modules, and isn't the focus of this PR

packages/core/src/QueryEngine.ts
```diff
@@ -206,3 +209,159 @@ export class SubQuestionQueryEngine implements BaseQueryEngine {
...
+    return new RouterQueryEngine({
+      selector: init.selector,
+      queryEngineTools: init.queryEngineTools,
+      serviceContext,
+      summarizer: init.summarizer,
+      verbose: init.verbose,
+    });
+  }
+
+  async _query(queryBundle: QueryBundle): Promise<Response> {
```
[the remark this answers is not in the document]
EmanuelCampos: updated

packages/core/src/QueryEngine.ts
```diff
@@ -206,3 +209,159 @@ export class SubQuestionQueryEngine implements BaseQueryEngine {
...
+  });
+
+  return new Response(summary, sourceNodes);
+}
+
+/**
+ * A query engine that uses multiple query engines and selects the best one.
+ */
+export class RouterQueryEngine implements BaseQueryEngine {
+  _selector: BaseSelector;
```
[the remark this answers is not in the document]
EmanuelCampos: updated

packages/core/src/QueryEngine.ts
```diff
@@ -206,3 +209,159 @@ export class SubQuestionQueryEngine implements BaseQueryEngine {
...
+  query(params: QueryEngineParamsStreaming): Promise<AsyncIterable<Response>>;
+  query(params: QueryEngineParamsNonStreaming): Promise<Response>;
+  async query(
+    params: QueryEngineParamsStreaming | QueryEngineParamsNonStreaming,
+  ): Promise<Response | AsyncIterable<Response>> {
+    const { query, stream } = params;
+
+    const response = await this._query({ queryStr: query });
+
+    if (stream) {
```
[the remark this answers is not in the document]
EmanuelCampos: throwing an error for now, stream wasnt figured out yet, just compability with overloading

packages/core/src/selectors/base.ts
EmanuelCampos: agreed, updated

packages/core/src/selectors/llmSelectors.ts
EmanuelCampos: Yes, not sure yet what's the plan to expand this later, but if not we would change to an multi arg parameter

packages/core/src/engines/query/RouterQueryEngine.ts
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      let selectedQueryEngine;
+
+      try {
+        selectedQueryEngine = this.queryEngines[result.selections[0].index];
+
+        const logStr = `Selecting query engine ${result.selections[0].index}: ${result.selections[0].reason}`;
+        console.log(logStr);
+        if (this.verbose) {
+          console.log(logStr + "\n");
```
yisding, line 160: double console.log here

packages/core/src/outputParsers/selectors.ts
```diff
@@ -0,0 +1,53 @@
+import { parseJsonMarkdown } from "../OutputParser";
+import { BaseOutputParser, StructuredOutput } from "../types";
+
+export type Answer = {
+  choice: number;
+  reason: string;
+};
+
+const formatStr = `
+The output should be ONLY JSON formatted as a JSON instance.
```
yisding, line 10: nit: in Python there's no newline at the beginning of this.

packages/core/src/outputParsers/selectors.ts
```diff
@@ -0,0 +1,53 @@
...
+          `Got invalid JSON object. Error: ${e}. Got JSON string: ${output}`,
+        );
+      }
+    }
+
+    return { rawOutput: output, parsedOutput: parsed };
+  }
+
+  format(output: string): string {
+    return output + "\n\n" + formatStr;
```
yisding, line 51: I'm guessing you don't need escape_curly_braces here because we don't have the same problem of the prompt being treated as a format string, but would be good to add a comment here.

packages/core/src/selectors/base.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export type SelectorResult = {
+  selections: SingleSelection[];
+};
+
+type QueryType = string | QueryBundle;
+
+function wrapChoice(
+  choice: string | ToolMetadataOnlyDescription,
```
yisding, line 15: Why do we want this ToolMetadataOnlyDescription? This should be able to take in any ToolMetadata.

packages/core/src/selectors/base.ts
```diff
@@ -0,0 +1,45 @@
...
+): ToolMetadataOnlyDescription {
+  if (typeof choice === "string") {
+    return { description: choice };
+  } else {
+    return choice;
+  }
+}
+
+function wrapQuery(query: QueryType): QueryBundle {
```
yisding, line 24: So we currently don't use the QueryBundle in LITS, and it looks like you're not really using the QueryBundle either since it only has the string, so I would 1. take it out and just use a query string for now, and 2. work with Logan to understand why we might want a QueryBundle in the future (if we actually do).

packages/core/src/selectors/prompts.ts
```diff
@@ -0,0 +1,28 @@
+export const getDefaultSingleSelectPrompt = (
```
yisding, line 1: Prompts in LITS are functions, but we don't need the get prefix for them. I would also export a type to make it easier for the user to replace the prompt with something else. See the main prompts.ts file for examples.

packages/core/src/selectors/prompts.ts
```diff
@@ -0,0 +1,28 @@
...
+  numChoices: number,
+  contextList: string,
+  queryStr: string,
+): string => {
+  return `Some choices are given below. It is provided in a numbered list (1 to ${numChoices}),
+where each item in the list corresponds to a summary.
```
yisding, line 7: nit: In python, this is on the same line as the previous line.

packages/core/src/selectors/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  serviceContext: ServiceContext,
+  isMulti: boolean = false,
+): BaseSelector => {
+  let selector: BaseSelector | null = null;
+  const llm = serviceContext.llm;
+
+  if (isMulti) {
+    selector = new LLMMultiSelector({
+      llm,
+      maxOutputs: 5,
```
yisding, line 15: In Python, it looks like this is len(choices) and not a fixed number

yisding (approved):
In general, this looks good. Left some comments. Please fix them and then let's merge. Please test, test, test, this thoroughly before we record the course. Thanks @EmanuelCampos !

packages/core/src/selectors/base.ts
EmanuelCampos: @yisding @marcusschiesser suggestion was to have only a description since we aren't using the name on ToolMetadata, and I agree because makes the use easier and we don't need to make name optional for other engines that requires it

packages/core/src/selectors/base.ts
EmanuelCampos: I think that's not a big priority for this PR then we can keep discussing with Logan to understand better and see if it makes sense or if we remove it

packages/core/src/selectors/utils.ts
EmanuelCampos: on python currently maxOutputs isn't being sent

EmanuelCampos: https://github.com/run-llama/llama_index/blob/main/llama_index/selectors/utils.py#L11-L33

PR #485, opened 2024-01-30
Update Astra DB Vectorstore to support namespaces

mfortman11:
Also improves metadata handling and adds additional example


examples/astradb/load.ts
```diff
@@ -10,9 +10,9 @@ const collectionName = "movie_reviews";
 async function main() {
   try {
     const reader = new PapaCSVReader(false);
-    const docs = await reader.loadData("../data/movie_reviews.csv");
+    const docs = await reader.loadData("./data/movie_reviews.csv");
```
marcusschiesser, line 13: astradb folder doesn't have a data folder anymore - please use the one from the parent dir

packages/core/src/storage/vectorStore/AstraDBVectorStore.ts
```diff
@@ -102,12 +105,13 @@ export class AstraDBVectorStore implements VectorStore {
...
       return [];
     }
+
     const dataToInsert = nodes.map((node) => {
       return {
-        _id: node.id_,
-        $vector: node.getEmbedding(),
-        content: node.getContent(MetadataMode.ALL),
-        metadata: node.metadata,
+        $vector: node.embedding,
```
marcusschiesser, line 111: seems to me like this new format is not compatible with Python, see:

https://github.com/run-llama/llama_index/blob/7cc6b1db27392ff07ecd5d95c5e00a743e212e93/llama_index/vector_stores/astra.py#L115-L129

suggestions:
1. use `nodeToMetadata`
2. defaults for keys being `_id`, `content` and `metadata`

marcusschiesser (commented):
@mfortman11 great job! thanks a lot for your PR - can you have a look at the comments

examples/astradb/load.ts
mfortman11: My readme instructions assumed the user was running this from the `examples` directory but can change that if you think that would be better.

packages/core/src/storage/vectorStore/AstraDBVectorStore.ts
mfortman11: @marcusschiesser thanks for the feedback. I made some changes to better handle metadata and align w/ the python version. I also was hoping to handle for the existing vector db table that doesn't necessarily have a metadata column.

For example:
image


Please let me know if my handling in `query` aligns with how llamaindex expects to handle this situation

examples/astradb/load.ts
marcusschiesser: oops you're right, good catch!

packages/core/src/storage/vectorStore/AstraDBVectorStore.ts
```diff
@@ -189,19 +199,25 @@ export class AstraDBVectorStore implements VectorStore {
...
-        embedding,
-      });
+      if (!metadata?.["_node_content"]) {
+        metadata["_node_content"] = JSON.stringify({
+          id,
+          text: content,
+          metadata,
+          ...rest,
+        });
+      }
```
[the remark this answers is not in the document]
marcusschiesser: If that works, we're ready to merge :)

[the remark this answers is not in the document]
mfortman11: Hey @marcusschiesser the problem I had w/o setting the `_node_content` was that if the vector table was created w/o llama the `metadataDictToNode` throws and error for having an empty `_node_content`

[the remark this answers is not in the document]
mfortman11: This was the situation I was trying to solve for [here](https://github.com/run-llama/LlamaIndexTS/pull/485#discussion_r1473417942)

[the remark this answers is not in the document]
marcusschiesser: @mfortman11 yes, my guess is you have been running in this error, because you have been reading data that was stored previously before adding `metadataDictToNode`.

To support this case, I moved your fallback check to the `metadataDictToNode` in this branch: https://github.com/run-llama/LlamaIndexTS/compare/mfortman11/astra?expand=1

Can you please check and test that branch? if it works, we're good to go

mfortman11:
@marcusschiesser this worked! Big fan of the fallback option thanks

marcusschiesser (approved):

marcusschiesser:
@mfortman11 Thanks a lot - looking great!

Issue #455, opened 2024-01-26
How to delete a document from the local storage?


akudo7:
Hi,

Can we delete a document from the local storage?

```typescript
// create a local storage 
async function main1() {
  try {
    // ストレージ作成
    const serviceContext = serviceContextFromDefaults({
      chunkSize: 512,
      chunkOverlap: 20,
    });
    const documents = await new SimpleDirectoryReader().loadData({
      // copy the abramov.txt from node_modules/llamaindex/examples
      directoryPath: "./data",
    });
    const storageContext = await storageContextFromDefaults({
      persistDir: "storage",
      storeImages: true,
    });
    const index = await VectorStoreIndex.fromDocuments(documents, {
      serviceContext,
      storageContext,
    });
    console.log(`Storage successfully generated.`);

    // Create query engine
    const retriever = index.asRetriever();
    retriever.similarityTopK = 5;
    const nodePostprocessor = new SimilarityPostprocessor({
      similarityCutoff: 0.7,
    });
    // TODO: cannot pass responseSynthesizer into retriever query engine
    const queryEngine = new RetrieverQueryEngine(
      retriever,
      undefined,
      undefined,
      [nodePostprocessor]
    );
    // Query the index
    const question = "What did the author do in college?";
    console.log("question:", question);
    const result = await queryEngine.query({query: question});
    console.log("response:", result.response);
  } catch (error) {
    console.error(error);
  }
}
```

Try to delete a document from the storage, but no luck.

```typescript
async function main() {
  try {
    // 既存のストレージを読み込む
    const storageContext = await storageContextFromDefaults({
      persistDir: "storage",
      storeImages: true,
    });
    const documents = await storageContext.docStore.getAllRefDocInfo();
    console.log("documents:", documents);
    /*
    documents: {
      './data/abramov.txt': {
        nodeIds: [
          'd6a17216-f1bb-4b64-b013-9323f9a1f7ab',
          'f7da1558-9ab6-42cb-9370-277d518f9d25',
          'a77e9fc6-1b03-4339-ac43-cbdf4d31b8ab',
          'c2c83ecc-b1b1-4b8d-9055-44a7ebe45f5d',
          'b5ce1656-dcca-40e4-9b1c-067177a45377',
          '1b5bd7e9-9740-4e4d-9d6a-cfeee51ddabf',
          'a83e58aa-760b-4875-97ad-c4201a8d9537',
          '7bef3937-23e4-41dc-b5cd-67988307748f',
[...]

PR #486, opened 2024-01-30
bugfix: Issue #455: StorageContext not saving data after deleting document/ref info

1997roylee:
Issue #455: StorageContext not saving data after deleting document/ref info

Changes Proposed:
Execute persist after delete the key.

I hv tried to execute the below code after deleting the document, but nothing happened.
```
await storageContext.docStore.persist();
await storageContext.docStore.persist(
  path.join("storage", DEFAULT_DOC_STORE_PERSIST_FILENAME)
);
```

### DIFF
--- a/packages/core/src/storage/kvStore/SimpleKVStore.ts
+++ b/packages/core/src/storage/kvStore/SimpleKVStore.ts
@@ -55,6 +55,11 @@ export class SimpleKVStore extends BaseKVStore {
   ): Promise<boolean> {
     if (key in this.data[collection]) {
       delete this.data[collection][key];
+
+      if (this.persistPath) {
+        await this.persist(this.persistPath, this.fs); // Save the data after deleting the key
+      }
+
       return true;
     }
     return false;


packages/core/src/storage/kvStore/SimpleKVStore.ts
```diff
@@ -55,6 +55,11 @@ export class SimpleKVStore extends BaseKVStore {
   ): Promise<boolean> {
     if (key in this.data[collection]) {
       delete this.data[collection][key];
+
+      if (this.persistPath) {
```
marcusschiesser, line 59: @1997roylee thanks for looking into #455! 
but persisting the whole store can take a long time, so doing it for every change operation seems not to be feasible. That's why we're having the explicit `persist` method.
We're working on a fix for #455.

1997roylee: It makes sense, but why is the 'persist' method checking if the instance is equal to 'BaseInMemoryKVStore'? I don't see any involvement of 'SimpleDocumentStore' with 'BaseInMemoryKVStore'. 

I proposed to change the BaseInMemoryKVStore to SimpleKVStore.

Original code:
```ts
this.kvStore instanceof BaseInMemoryKVStore
```

Now:
```ts
this.kvStore instanceof SimpleKVStore
```

marcusschiesser: @1997roylee good catch, i think we don't need this instanceof check at all - @himself65 can you have a look into this while working on #487 ?

himself65: yes I think instanceof is useless here. But it okay for now

1997roylee: should I close this PR? Or I can help to fix the persist issue of another class.

PR #487, opened 2024-01-30
fix: persist class in `KVIndexStore`

himself65:


packages/core/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "preserveWatchOutput": true,
     "skipLibCheck": true,
     "strict": true,
-    "lib": ["es2015", "dom"],
+    "lib": ["ES2022", "dom"],
```
marcusschiesser, line 13: @himself65 might the upgrade break anythin?

PR #489, opened 2024-01-31
feat(create-llama) add folder selection & support more context data types

leehuwuj:
- Add a folder selection
- Update pdf selection into local file selection with the allowed file types are: [".pdf", ".doc", ".docx", ".ppt", ".pptx", ".csv"]

### DIFF
--- a/.changeset/honest-dragons-cough.md
+++ b/.changeset/honest-dragons-cough.md
@@ -1,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+Add folder selection and support more file types
--- a/packages/create-llama/helpers/index.ts
+++ b/packages/create-llama/helpers/index.ts
@@ -131,13 +131,13 @@ const copyContextData = async (
 
   // Copy file
   if (dataSource?.type === "file") {
-    let srcFile =
-      dataSourceConfig.path ??
-      path.join(__dirname, "..", "templates", "components", "data", "101.pdf");
-    if (dataSourceConfig) {
+    if (dataSourceConfig.path) {
       console.log(`\nCopying file to ${cyan(destPath)}\n`);
       await fs.mkdir(destPath, { recursive: true });
-      await fs.copyFile(srcFile, path.join(destPath, path.basename(srcFile)));
+      await fs.copyFile(
+        dataSourceConfig.path,
+        path.join(destPath, path.basename(dataSourceConfig.path)),
+      );
     } else {
       console.log("Missing file path in config");
       process.exit(1);
@@ -147,16 +147,14 @@ const copyContextData = async (
 
   // Copy folder
   if (dataSource?.type === "folder") {
-    if (dataSourceConfig.path !== undefined) {
-      console.log(`\nCopying data to ${cyan(destPath)}\n`);
-      await copy("**", destPath, {
-        parents: true,
-        cwd: dataSourceConfig.path,
-      });
-    } else {
-      console.log("Missing folder path in config");
-      process.exit(1);
-    }
+    let srcPath =
+      dataSourceConfig.path ??
+      path.join(__dirname, "..", "templates", "components", "data");
+    console.log(`\nCopying data to ${cyan(destPath)}\n`);
+    await copy("**", destPath, {
+      parents: true,
+      cwd: srcPath,
+    });
     return;
   }
 };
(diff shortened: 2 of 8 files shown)


packages/create-llama/index.ts
```diff
@@ -83,6 +83,13 @@ const program = new Commander.Command(packageJson.name)
     `
 
   Select a framework to bootstrap the application with.
+`,
+  )
+  .option(
+    "--files <path>",
+    `
+  
+    Specify the path to a local file or folder for chatting.
```
marcusschiesser, line 92:
```suggestion
    Specify the path to a local file or folder to use as datasource.
```

packages/create-llama/helpers/index.ts
```diff
@@ -120,28 +121,43 @@ const installDependencies = async (
...
-      cwd: srcPath,
-    });
+
+  let dataSourceConfig = dataSource?.config as FileSourceConfig;
+
+  // Copy file
+  if (dataSource?.type === "file") {
+    let srcFile =
+      dataSourceConfig.path ??
+      path.join(__dirname, "..", "templates", "components", "data", "101.pdf");
```
marcusschiesser, line 136: can we just use the whole `components/data` folder like before? if we rename the pdf file this code will break

.changeset/honest-dragons-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+Add folder selection and support more file types
```
marcusschiesser, line 5: does it work with express, fastapi and nextjs?

packages/create-llama/helpers/python.ts
```diff
@@ -177,17 +179,22 @@ export const installPythonTemplate = async ({
...
-      parents: true,
-      cwd: loaderPath,
-    });
+
+    const dataSourceType = dataSource?.type;
+    if (dataSourceType !== undefined && dataSourceType !== "none") {
+      let loaderPath =
+        dataSourceType === "folder"
+          ? path.join(compPath, "loaders", "python", "file")
+          : path.join(compPath, "loaders", "python", dataSourceType);
```
marcusschiesser, line 192: why we need a different loader for file or folder?

packages/create-llama/questions.ts
```diff
@@ -78,39 +103,46 @@ const getVectorDbChoices = (framework: TemplateFramework) => {
...
+        process.exit(1);
+      }
     }
-    return selectedFilePath;
+    return selectedPath;
   } catch (error) {
     console.log(
       red(
-        "Got error when trying to select file! Please try again or select other options.",
+        "Got error when trying to select local context data! Please try again or select other options.",
```
marcusschiesser, line 145:
```suggestion
        "Got an error when trying to select local context data! Please try again or select another data source option.",
```

packages/create-llama/questions.ts
```diff
@@ -373,108 +405,155 @@ export const askQuestions = async (
...
+            case "exampleFile":
+              program.engine = "context";
+              // example file is a context app with dataSource.type = file but has no config
+              program.dataSource = { type: "file", config: {} };
+              break;
+            case "localFile":
+              program.engine = "context";
+              program.dataSource.type = "file";
+              program.dataSource.config = {
+                path: await selectLocalContextData(program.dataSource),
```
marcusschiesser, line 473: i guess only the type is needed?
```suggestion
                path: await selectLocalContextData(program.dataSource.type),
```

packages/create-llama/questions.ts
```diff
@@ -373,108 +405,155 @@ export const askQuestions = async (
...
-          name: "baseUrl",
-          message: "Please provide base URL of the website:",
-          initial: "https://www.llamaindex.ai",
-        },
-        handlers,
-      );
-      try {
-        if (!baseUrl.includes("://")) {
-          baseUrl = `https://${baseUrl}`;
+      } else {
```
marcusschiesser, line 423: this code is getting quite complex, can we extract a function?

.changeset/honest-dragons-cough.md
leehuwuj: yes, it is

packages/create-llama/helpers/python.ts
leehuwuj: this code already uses the same loader for both file and folder, it just maps `folder` to `file`

packages/create-llama/helpers/index.ts
leehuwuj: because we provide an option as `An example PDF` so it should copy a file rather than a whole folder. If we change the file name, then the test will fail.

marcusschiesser: Can't we treat the option `An example PDF` so that we're copying the `components/data` folder? Then whatever is in this folder will be used

marcusschiesser (approved):

PR #490, opened 2024-01-31
feat: use pnpm pack for e2e

thucpn:


.github/workflows/e2e.yml
```diff
@@ -49,6 +49,12 @@ jobs:
       - name: Build create-llama
         run: pnpm run build
         working-directory: ./packages/create-llama
+      - name: Pack
+        run: pnpm pack
+        working-directory: ./packages/create-llama
+      - name: Extract Pack
+        run: tar -xvzf create-llama-0.0.21.tgz
```
marcusschiesser, line 56: 1. that will break if we update the version
2. better use temp dir
(see test.yml)

packages/create-llama/create-app.ts
```diff
@@ -100,7 +101,13 @@ export async function createApp({
...
     // copy readme for fullstack
     await fs.promises.copyFile(
-      path.join(__dirname, "..", "templates", "README-fullstack.md"),
+      path.join(
+        fileURLToPath(import.meta.url),
+        "..",
+        "..",
+        "templates",
+        "README-fullstack.md",
+      ),
```
marcusschiesser, line 110: Better to calculcate `templatesDir` once and use it everywhere
```suggestion
      path.join(
        templatesDir,
        "README-fullstack.md",
      ),
```

PR #498, opened 2024-01-31
feat: improve reader interfaces

himself65:


examples/readers/data/brk-2022.pdf
marcusschiesser, line 1: not sure it's a good idea to duplicate all our data - how could we avoid this?

marcusschiesser: @himself65 but i love the idea of grouping some examples and using them on stackblitz 👍

marcusschiesser (approved):
@himself65 thanks. looks good to me. please merge

PR #501, opened 2024-02-01
fix: update deprecated pnpx to pnpm dlx

TechPandaPro:
Hello!

This pull request updates two of the README files to use `pnpm dlx` instead of `pnpx`.

Due to the [deprecation/removal of `pnpx`](https://github.com/pnpm/pnpm/pull/3652), the current examples in the READMEs may not work for some users. There are also no docs for the `pnpx` command on pnpm.io now that it's removed. As of its deprecation in pnpm 6.x, it is now recommended to use [`pnpm dlx`](https://pnpm.io/cli/dlx) instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ main();
 Then you can run it using
 
 ```bash
-pnpx ts-node example.ts
+pnpm dlx ts-node example.ts
 ```
 
 ## Playground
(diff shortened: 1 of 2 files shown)


marcusschiesser (approved):
Thanks @TechPandaPro great catch!

PR #507, opened 2024-02-02
Fix typo

martimatix:
metadara -> metadata

Small change to prevent pinecone queries from crashing.


marcusschiesser (approved):
Thanks @martimatix great catch!

PR #509, opened 2024-02-02
feat(queryEngineTool): add query engine tool to agents

EmanuelCampos:
Depends on #416


packages/core/src/engines/query/SubQuestionQueryEngine.ts
```diff
@@ -122,13 +117,21 @@ export class SubQuestionQueryEngine implements BaseQueryEngine {
...
-      const response = await queryEngine.query({
+      const queryEngine = this.queryEngines.find(
+        (tool) => tool.metadata.name === subQ.toolName,
+      );
+
+      if (!queryEngine) {
+        return null;
+      }
+
+      const response = await queryEngine?.call?.({
```
marcusschiesser, line 129: `response` might be `undefined` - explicitly handling this case?

packages/core/src/tools/QueryEngineTool.ts
```diff
@@ -0,0 +1,56 @@
+// from typing import TYPE_CHECKING, Any, Optional
```
marcusschiesser, line 1: remove

packages/core/src/tools/QueryEngineTool.ts
```diff
@@ -0,0 +1,56 @@
...
+      type: "string",
+      description: "The query to search for",
+    },
+  },
+  required: ["query"],
+};
+
+export class QueryEngineTool implements BaseTool {
+  private _queryEngine: BaseQueryEngine;
+  private _metadata: ToolMetadata;
```
marcusschiesser, line 26:
```suggestion
  private queryEngine: BaseQueryEngine;
  private metadata: ToolMetadata;
```

packages/core/src/tools/QueryEngineTool.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  get metadata() {
+    return this._metadata;
+  }
+
+  async call(...args: any[]): Promise<any> {
+    let queryStr: string;
+
+    if (args && args.length > 0) {
+      queryStr = String(args[0].query);
```
marcusschiesser, line 45: Can we NOT use `any` for `args`? 
Looks like we're only using `query`

packages/core/src/tools/index.ts
```diff
@@ -1,2 +1,4 @@
+export * from "./QueryEngineTool";
 export * from "./functionTool";
 export * from "./types";
+export * from "./utils";
```
marcusschiesser, line 4: exporting utils necessary?

PR #519, opened 2024-02-06
Consistent Document IDs in NotionReader.ts

iansinnott:
## Problem

When using the NotionReader `Document` objects do not have the same ID as the Notion document they represent. This looks to simply be an oversight, rather than intended behavior. The Notion document ID is available.

## Solution

Use the Notion document ID as the ID of the `Document` object. This ensures consistent `id_`s for each ingestion run.


iansinnott:
I'm guessing the current behavior was not intended, but let me know if a random UUID is actually what is desired for Notion docs.

marcusschiesser (approved):
Thanks @iansinnott that looks great!

PR #522, opened 2024-02-07
fix: add handle error from template installation

leehuwuj:


packages/create-llama/helpers/python.ts
```diff
@@ -103,22 +103,19 @@ export const installPythonDependencies = (
...
-          "Poetry Installation",
-          `https://python-poetry.org/docs/#installation`,
-        )} to install poetry first, then install the dependencies manually.`,
-      ),
-    );
-    process.exit(1);
+      "Poetry Installation",
+      `https://python-poetry.org/docs/#installation`,
+    )} to install poetry first, then install the dependencies manually.`;
+    console.log(red(errorMessage));
```
marcusschiesser, line 117: Isn't the thrown error logged again? Wouldn't that result in logging the same message twice?

packages/create-llama/create-app.ts
```diff
@@ -88,20 +88,41 @@ export async function createApp({
...
     const frontendRoot = path.join(root, "frontend");
     await makeDir(frontendRoot);
-    await installTemplate({
-      ...args,
-      root: frontendRoot,
-      framework: "nextjs",
-      customApiPath: `http://localhost:${externalPort ?? 8000}/api/chat`,
-      backend: false,
-    });
+    try {
```
marcusschiesser, line 109: Why we still try the frontend installation if the backend has failed?

packages/create-llama/helpers/python.ts
leehuwuj: my bad, i added this one with my coroutine code to show the error log during installation and show it again in the final step to help user catch the error easier, but i did remove it so this one is duplicated now. Thanks for the catch!

packages/create-llama/create-app.ts
```diff
@@ -88,20 +88,41 @@ export async function createApp({
...
+        framework: "nextjs",
+        customApiPath: `http://localhost:${externalPort ?? 8000}/api/chat`,
+        backend: false,
+      });
+    } catch (error) {
+      successfulInstallation = false;
+      console.log(error);
+    }
+    if (!successfulInstallation) {
+      console.log(red("\nOne or more installations failed. Exiting...\n"));
```
marcusschiesser, line 122: Exists too early, all files must be generated first

packages/create-llama/create-app.ts
```diff
@@ -88,20 +88,41 @@ export async function createApp({
...
     await makeDir(backendRoot);
-    await installTemplate({ ...args, root: backendRoot, backend: true });
+    try {
+      await installTemplate({
+        ...args,
+        root: backendRoot,
+        backend: true,
+      });
+    } catch (error) {
+      successfulInstallation = false;
```
marcusschiesser, line 103: Here we have two errors:
1. Code generation fails
2. Install dependencies fails

If 1. I guess we don't want to continue installing the frontend?

packages/create-llama/create-app.ts
leehuwuj: agree, we should still generate all the files (docs, git,..) that user can reference if they were failed.

packages/create-llama/create-app.ts
leehuwuj: I thought about this, but our code currently is not easy to do that because: 

we do not separate code generation with dependency installation (for frontend and backend), but we can do that by introduce two types of error then handle it accordingly (but i think this one is not kind of our code pattern).
so i just consider the errors as one type and continue installing the frontend.

do you have any suggestion?

marcusschiesser: Makes sense. The problem here is that we might have an error that requires user interaction (i.e. installing poetry).
Can we easily show the reason for the error at the end (before the process.exit), e.g. by doing `installationError = error`?

leehuwuj: Nice suggestion, i'll add it

packages/create-llama/create-app.ts
```diff
@@ -87,21 +87,38 @@ export async function createApp({
...
     await makeDir(frontendRoot);
-    await installTemplate({
-      ...args,
-      root: frontendRoot,
-      framework: "nextjs",
-      customApiPath: `http://localhost:${externalPort ?? 8000}/api/chat`,
-      backend: false,
-    });
+    try {
+      await installTemplate({
```
marcusschiesser, line 110: what about the normal `installTemplate` case (frontend and backend are the same)?

marcusschiesser: seems to me if it throws an error they are handled one level above - but we should have the same error handling

leehuwuj: in that case the `frontend` will be false then it'll go through other branch, see:
https://github.com/run-llama/LlamaIndexTS/blob/0ac403106743ccb8c1baa60e6fe47de4871f78b9/packages/create-llama/create-app.ts#L128

the app the will be the same as is

.changeset/shaggy-otters-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+fix: continue install frontend template if the fastapi backend has failed
```
marcusschiesser, line 5:
```suggestion
fix: continue installing the frontend if the backend installation has failed
```

marcusschiesser (approved):

PR #523, opened 2024-02-07
Update NotionReader.ts to support loading from databases

iansinnott:
# Problem

Cannot load databases via the Notion reader. Without this change it is limited to notion pages. (Notion distinguishes between "page" and "database" objects).

# Solution

Utilize the existing functionality of the `notion-md-crawler` lib: https://github.com/TomPenguin/notion-md-crawler?tab=readme-ov-file#dbcrawler

# Details

The underlying Notion crawler library supports both pages and database, but does not automatically distinguish between them. 

This change allows specifying the type of data you want to load via the `loadData` method. 

This should be a non-breaking change, making the existing behavior the default.

### DIFF
--- a/packages/core/src/readers/NotionReader.ts
+++ b/packages/core/src/readers/NotionReader.ts
@@ -1,5 +1,12 @@
 import { Client } from "@notionhq/client";
-import { crawler, Crawler, Pages, pageToString } from "notion-md-crawler";
+import {
+  Crawler,
+  DatabaseCrawler,
+  Pages,
+  crawler,
+  dbCrawler,
+  pageToString,
+} from "notion-md-crawler";
 import { Document } from "../Node";
 import { BaseReader } from "./base";
 
@@ -25,6 +32,7 @@ type NotionReaderOptions = {
  */
 export class NotionReader implements BaseReader {
   private crawl: ReturnType<Crawler>;
+  private dbCrawl: ReturnType<DatabaseCrawler>;
 
   /**
    * Constructor for the NotionReader class
@@ -32,6 +40,7 @@ export class NotionReader implements BaseReader {
    */
   constructor({ client, serializers }: NotionReaderOptions) {
     this.crawl = crawler({ client, serializers });
+    this.dbCrawl = dbCrawler({ client, serializers });
   }
 
   /**
@@ -42,11 +51,7 @@ export class NotionReader implements BaseReader {
   toDocuments(pages: Pages): Document[] {
     return Object.values(pages).map((page) => {
       const text = pageToString(page);
-      return new Document({
-        id_: page.metadata.id, // Use the Notion-provided UUID for the document
-        text,
-        metadata: page.metadata,
-      });
+      return new Document({ text, metadata: page.metadata });
     });
   }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/readers/NotionReader.ts
```diff
@@ -42,11 +51,7 @@ export class NotionReader implements BaseReader {
   toDocuments(pages: Pages): Document[] {
     return Object.values(pages).map((page) => {
       const text = pageToString(page);
-      return new Document({
-        id_: page.metadata.id, // Use the Notion-provided UUID for the document
-        text,
-        metadata: page.metadata,
-      });
+      return new Document({ text, metadata: page.metadata });
```
marcusschiesser, line 54: why did you remove the `id_`?

packages/core/src/readers/NotionReader.ts
```diff
@@ -59,13 +64,32 @@ export class NotionReader implements BaseReader {
...
+   * @param {string} databaseId - The Notion database ID from which to start the crawl
+   * @returns {Promise<Pages>} A Promise that resolves to a Pages object(Convertible with the `toDocuments` method)
+   */
+  async loadDatabasePages(databaseId: string): Promise<Pages> {
+    return this.dbCrawl(databaseId);
+  }
+
   /**
    * Loads recursively Notion pages and converts them to an array of Document objects
    * @param {string} rootPageId - The root Notion page ID
```
marcusschiesser, line 78:
```suggestion
   * @param {string} id - The root Notion page ID (for mode "page") or the data base ID (for mode "database")
```

packages/core/src/readers/NotionReader.ts
```diff
@@ -59,13 +64,32 @@ export class NotionReader implements BaseReader {
...
+   * @param {Object} [options] - Options for loading data
+   * @param {"page" | "database"} [options.mode="page"] - The mode to use when loading data. "page" to load a page, "database" to load a database
    * @returns {Promise<Document[]>} A Promise that resolves to an array of Document objects
    */
-  async loadData(rootPageId: string): Promise<Document[]> {
-    const pages = await this.loadPages(rootPageId);
+  async loadData(
+    rootPageId: string,
```
marcusschiesser, line 84:
```suggestion
    id: string,
```

packages/core/src/readers/NotionReader.ts
```diff
@@ -59,13 +64,32 @@ export class NotionReader implements BaseReader {
     return this.crawl(rootPageId);
   }
 
+  /**
+   * Loads recursively the Notion database with the specified database ID.
+   * @param {string} databaseId - The Notion database ID from which to start the crawl
+   * @returns {Promise<Pages>} A Promise that resolves to a Pages object(Convertible with the `toDocuments` method)
+   */
+  async loadDatabasePages(databaseId: string): Promise<Pages> {
+    return this.dbCrawl(databaseId);
```
marcusschiesser, line 73: does this also return `Pages`?

iansinnott: Hm, I was giong to say yes as that's what the typescript types say in the editor. However, after digging in to the code to confirm I found that it will treat the page as a database if normal extraction fails:

https://github.com/TomPenguin/notion-md-crawler/blob/main/src/crawler.ts#L279-L280

So, it would seem this PR is unnecessary. It would be more efficient to avoid the work in the `try` branch but avoiding additional complexity is probably more important.

iansinnott:
> Hm, I was giong to say yes as that's what the typescript types say in the editor. However, after digging in to the code to confirm I found that it will treat the page as a database if normal extraction fails:
> 
> https://github.com/TomPenguin/notion-md-crawler/blob/main/src/crawler.ts#L279-L280
> 
> So, it would seem this PR is unnecessary. It would be more efficient to avoid the work in the `try` branch but avoiding additional complexity is probably more important.

https://github.com/run-llama/LlamaIndexTS/pull/523#discussion_r1480991225

Given that, i'll close this out for now.

PR #530, opened 2024-02-08
feat: Add reader for LlamaParse

marcusschiesser:


packages/core/src/readers/index.ts
```diff
@@ -3,6 +3,7 @@ export * from "./CSVReader";
 export * from "./DocxReader";
 export * from "./HTMLReader";
 export * from "./ImageReader";
+export * from "./LlamaParseReader";
```
marcusschiesser, line 6: @himself65 Contrary to the other readers, I didn't add an explicit export in package.json. Makes sense to me because a) it doesn't have any special dependencies and b) we probably always want to include this code

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!file.endsWith(".pdf")) {
+      throw new Error("Currently, only PDF files are supported.");
+    }
+
+    const metadata = { file_path: file };
+
+    // Load data, set the mime type
+    const data = await fs.readRawFile(file);
+    const body = new FormData();
+    body.set("file", new Blob([data], { type: "application/pdf" }), file);
```
marcusschiesser, line 49: @hexapode this code didn't work without explicitly setting the mime-type - maybe the API could guess the mime type?

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.apiKey = params.apiKey;
+  }
+
+  async loadData(
+    file: string,
+    fs: GenericFileSystem = defaultFS,
+  ): Promise<Document[]> {
+    if (!file.endsWith(".pdf")) {
+      throw new Error("Currently, only PDF files are supported.");
+    }
```
EmanuelCampos, line 42: Sometimes the file doesn't end with .pdf but is a pdf by the mime type, why not check the mime-type?

EmanuelCampos (approved):

packages/core/src/readers/LlamaParseReader.ts
himself65, line 42: No, use a file-type check instead. file ext is not file type guarantee

himself65 (changes requested):

packages/core/src/readers/LlamaParseReader.ts
marcusschiesser: @EmanuelCampos and @himself65 Okay, you guys convinced me :D ; mime-type check added. Currently, this makes more sense. My thinking was that LlamaParse will support multiple file types in the future. Then it makes sense that the client is just sending whatever file to the LlamaParse server, and LlamaParse is doing the mime-type check

PR #534, opened 2024-02-09
Add support for AstraDB in the create-llama cli

erichare:
Python code only, no typescript at the moment

### DIFF
--- a/packages/create-llama/templates/components/vectordbs/python/astra/constants.py
+++ b/packages/create-llama/templates/components/vectordbs/python/astra/constants.py
@@ -1,0 +1,3 @@
+DATA_DIR = "data"  # directory containing the documents to index
+CHUNK_SIZE = 1024
+CHUNK_OVERLAP = 20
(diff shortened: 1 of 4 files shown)


EmanuelCampos:
@erichare some tests are broken, can you check that and make sure it's following what create-llama expects?

erichare:
@EmanuelCampos what's interesting is from what i can tell, the tests are saying that an OpenAI Key is not being provided. i didnt think that anything i added specified something specific about OpenAI, like using it as the service context. So I can't quite make sense of this. Do you have any thoughts? I could be missing something completely.

marcusschiesser:
@erichare, yes, the e2e tests are not running for outside contributors. The reason is that this could leak the OpenAI key from our CI pipeline. But no worries; we'll use this PR as a draft for adding AstraDB to create-llama. 
Thanks a lot for your contribution!

Next steps:
1. adding TS code for using AstraDB vector store
2. modifying create-llama to add AstraDB as an option

marcusschiesser:
Will continue on this in https://github.com/run-llama/create-llama

marcusschiesser:
Just merged this to `main` of  https://github.com/run-llama/create-llama - will be in the next release

PR #543, opened 2024-02-13
feat: prompt mixin

EmanuelCampos:


packages/core/src/prompts/Mixin.ts
```diff
@@ -0,0 +1,99 @@
...
+    return {};
+  }
+
+  protected _getPromptModules(): Record<string, any> {
+    return {};
+  }
+
+  protected _updatePrompts(promptsDict: PromptsDict): void {
+    return;
+  }
```
himself65, line 98: use `abstract`?

EmanuelCampos: with abstract we will always need to implement all classes

apps/docs/docs/modules/prompt/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+// Query the index
+const queryEngine = index.asQueryEngine({ responseSynthesizer });
+
+const response = await queryEngine.query({
+  query: "What did the author do in college?",
+});
+```
+
+### 2. Customizing submodules prompt
```
yisding, line 50: Overall, good job on the documentation. This example here isn't customizing the submodules prompt though, but rather the textQATemplate

packages/core/src/QuestionGenerator.ts
```diff
@@ -7,22 +7,42 @@ import {
...
   }
 
+  protected _getPrompts(): { [x: string]: SubQuestionPrompt } {
+    return {
+      subQuestion: this.prompt,
+    };
+  }
+
+  protected _updatePrompts(promptsDict: {
+    [x: string]: SubQuestionPrompt;
```
yisding, line 39: Since you're looking for a specific prompt you can set the type variable name explicitly.

packages/core/src/engines/chat/ContextChatEngine.ts
```diff
@@ -43,6 +46,12 @@ export class ContextChatEngine implements ChatEngine {
     });
   }
 
+  protected _getPromptModules(): Record<string, ContextGenerator> {
```
yisding, line 49: Are we missing the update function here?

packages/core/src/engines/query/RouterQueryEngine.ts
```diff
@@ -31,8 +33,8 @@ async function combineResponses(
     console.log("Combining responses from multiple query engines.");
   }
 
-  const responseStrs = [];
-  const sourceNodes = [];
+  const responseStrs: string[] = [];
+  const sourceNodes: BaseNode[] = [];
```
yisding, line 37: Good, thanks for this fix.

packages/core/src/engines/query/RouterQueryEngine.ts
```diff
@@ -79,6 +83,13 @@ export class RouterQueryEngine implements BaseQueryEngine {
     this.verbose = init.verbose ?? false;
   }
 
+  _getPromptModules(): Record<string, any> {
```
yisding, line 86: Is there an update function missing here?

packages/core/src/prompts/Mixin.ts
```diff
@@ -0,0 +1,91 @@
+type PromptsDict = Record<string, any>;
+
+export class PromptMixin {
+  /**
+   * Validates the prompt keys and module keys
+   * @param promptsDict
+   * @param moduleDict
+   */
+  validatePrompts(
+    promptsDict: PromptsDict,
```
yisding, line 10: A little inconsistent here with the two types with one defined earlier and the other defined inline.

yisding (commented):
Good job on this! A few small fixes, but otherwise looks good.

apps/docs/docs/modules/prompt/index.md
EmanuelCampos: @yisding responseSynthesizer is a submodule from query engine

updating directly a module would be something like:

```
const treeSummarizer = new TreeSummarizer()
treeSummarize._updatePrompts()
```

PR #544, opened 2024-02-13
fix: some errors in available llms examples

buttercubz:
fix some bugs in the examples and add the step to get the retriever, and makes the example of using ollama clearer


EmanuelCampos (approved):
Great ones, thanks for the contribution!

PR #554, opened 2024-02-21
Add support for LlamaCloud

marcusschiesser:
Usage: see `examples/cloud`


packages/core/src/cloud/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function getClient({
+  apiKey,
+  baseUrl,
+}: ClientParams = {}): Promise<PlatformApiClient> {
+  // Get the environment variables or use defaults
+  baseUrl = baseUrl ?? process.env.LLAMA_CLOUD_BASE_URL ?? DEFAULT_BASE_URL;
+  apiKey = apiKey ?? process.env.LLAMA_CLOUD_API_KEY;
+
+  const { PlatformApiClient } = await import("@llamaindex/cloud");
```
marcusschiesser, line 12: @himself65, we need to do an `await import` here as @llamaindex/cloud is an ESM module.

Unfortunately, we're getting an `ERR_UNSUPPORTED_DIR_IMPORT` error (stack trace below) doing so. 
The `await import` is used a couple of times (e.g. for transformers.js), so it shouldn't be an issue with our setup but with the code generated by Fern.
Any idea? 

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@llamaindex+cloud@0.0.1/node_modules/@llamaindex/cloud/lib/api' is not supported resolving ES modules imported from /Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@llamaindex+cloud@0.0.1/node_modules/@llamaindex/cloud/lib/index.js
    at new NodeError (node:internal/errors:405:5)
    at finalizeResolution (node:internal/modules/esm/resolve:321:11)
    at moduleResolve (node:internal/modules/esm/resolve:980:10)
    at defaultResolve (node:internal/modules/esm/resolve:1193:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:403:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:372:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:249:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
    at link (node:internal/modules/esm/module_job:75:36) {
  url: 'file:///Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@llamaindex+cloud@0.0.1/node_modules/@llamaindex/cloud/lib/api',
  code: 'ERR_UNSUPPORTED_DIR_IMPORT'
```

packages/core/src/cloud/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { PlatformApiClient } from "@llamaindex/cloud";
```
marcusschiesser, line 1: NOTE: As we're using `await import` below, this import won't translate to a `require` call (it's just used for type checking by Typescript)

packages/core/src/cloud/LlamaCloudRetriever.ts
```diff
@@ -0,0 +1,85 @@
+import { PlatformApi, PlatformApiClient } from "@llamaindex/cloud";
```
marcusschiesser, line 1: NOTE: This import won't translate to a `require` call (it's just used for type checking by Typescript)

packages/core/src/cloud/LlamaCloudIndex.ts
```diff
@@ -0,0 +1,38 @@
+import { BaseRetriever } from "../Retriever";
+import { RetrieverQueryEngine } from "../engines/query/RetrieverQueryEngine";
+import { BaseNodePostprocessor } from "../postprocessors";
+import { BaseSynthesizer } from "../synthesizers";
+import { BaseQueryEngine } from "../types";
+import { LlamaCloudRetriever, RetrieveParams } from "./LlamaCloudRetriever";
+import { CloudConstructorParams } from "./types";
+
+export class LlamaCloudIndex {
```
marcusschiesser, line 9: **NOTE**: I intentionally didn't extend from `BaseIndex`. Reasons:
1. we don't need the complete implementation from `BaseIndex`
2. we could extract a common interface (containing `asRetriever` and `asQueryEngine`) but I doubt it will have a great value

packages/core/src/cloud/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ClientParams, DEFAULT_BASE_URL } from "./types";
+
+export async function getClient({
+  apiKey,
+  baseUrl,
+}: ClientParams = {}): Promise<PlatformApiClient> {
+  // Get the environment variables or use defaults
+  baseUrl = baseUrl ?? process.env.LLAMA_CLOUD_BASE_URL ?? DEFAULT_BASE_URL;
+  apiKey = apiKey ?? process.env.LLAMA_CLOUD_API_KEY;
```
EmanuelCampos, line 10: throw an error if no API key is set?

EmanuelCampos (commented):
just a nit, looks great

packages/core/src/cloud/utils.ts
himself65: no need, just runtime no auth error

packages/core/package.json
```diff
@@ -206,12 +212,12 @@
   "scripts": {
     "lint": "eslint .",
     "test": "jest",
-    "build": "rm -rf ./dist && NODE_OPTIONS=\"--max-old-space-size=8192\" bunchee",
+    "build": "rm -rf ./dist && NODE_OPTIONS=\"--max-old-space-size=16384\" bunchee",
```
himself65, line 215: Memory usage is too much now. I will try to create a manual rollup config

marcusschiesser: 8192 is not working, but 12288 is - so i keep that value for now

marcusschiesser: bunchee is great, but it doesn't seem mature enough yet

PR #555, opened 2024-02-21
feat: add pinecone support to create llama

thucpn:


packages/core/src/storage/vectorStore/PineconeVectorStore.ts
```diff
@@ -207,14 +208,10 @@ export class PineconeVectorStore implements VectorStore {
...
     let id: any = node.id_.length ? node.id_ : null;
-    let meta: any = node.metadata || {};
-    meta.create_date = new Date();
-    meta.text = node.getContent(MetadataMode.EMBED);
-
     return {
       id: id,
       values: node.getEmbedding(),
-      metadata: meta,
+      metadata: nodeToMetadata(node),
```
marcusschiesser, line 214: @thucpn you also need the inverse function (metadataToNode) for querying data

.changeset/pink-tools-look.md
```diff
@@ -0,0 +1,6 @@
+---
+"llamaindex": patch
+"create-llama": patch
```
marcusschiesser, line 3: should be one file per package

marcusschiesser (approved):

Issue #551, opened 2024-02-18
update create-llama to llama-index 0.10 +


ohdearquant:
`Llama - Index` had a huge upgrade, and the `create-llama` need to be updated as well. For example, `ServiceContext` is depreciated. 

```python
from llama_index import ServiceContext

from app.context import create_base_context
from app.engine.constants import CHUNK_SIZE, CHUNK_OVERLAP


def create_service_context():
    base = create_base_context()
    return ServiceContext.from_defaults(
        llm=base.llm,
        embed_model=base.embed_model,
        chunk_size=CHUNK_SIZE,
        chunk_overlap=CHUNK_OVERLAP,
    )
```

leehuwuj:
Hi @ohdearquant, Thanks for the request! I'm already on it and hope to deliver it soon.

PR #556, opened 2024-02-21
Upgrade llama-index version to v0.10+ for create-llama

leehuwuj:
### Changes:
- Upgraded llama-index to version 0.10.7.
- Replaced ServiceContext with global Settings.
- Modified llama-index package imports.

Fix: https://github.com/run-llama/LlamaIndexTS/issues/551


packages/create-llama/questions.ts
```diff
@@ -582,13 +582,16 @@ export const askQuestions = async (
...
-      const { tools } = await prompts({
+      const { toolsName } = await prompts({
         type: "multiselect",
-        name: "tools",
+        name: "toolsName",
         message:
           "Would you like to build an agent using tools? If so, select the tools here, otherwise just press enter",
         choices: toolChoices,
       });
+      const tools = toolsName?.map((tool: string) =>
```
marcusschiesser, line 592: @leehuwuj if you use the tool object here instead of just the name, how would you handle than passing the tools as cli parameter?

packages/create-llama/templates/components/engines/python/agent/__init__.py
```diff
@@ -44,7 +39,7 @@ def get_chat_engine():
     tools += ToolFactory.from_env()
 
     return create_agent_from_llm(
```
marcusschiesser, line 41: Better use `AgentRunner.from_llm` now and remove the factory (everywhere)
```suggestion
    return AgentRunner.from_llm(
```

packages/create-llama/templates/components/loaders/python/web/loader.py
```diff
@@ -1,5 +1,5 @@
 import os
-from llama_index import VectorStoreIndex, download_loader
+from llama_index.core.readers.download import download_loader
```
marcusschiesser, line 2: no package for `WholeSiteReader`?

packages/create-llama/templates/types/simple/fastapi/app/settings.py
```diff
@@ -0,0 +1,8 @@
+import os
+from llama_index.llms.openai import OpenAI
+from llama_index.core.settings import Settings
+
+
+def init_base_settings():
```
marcusschiesser, line 6: how can call this `init_settings` and override it if we select a vector store?

packages/create-llama/questions.ts
leehuwuj: Good catch, I forgot it! I think using the tool object is easier for handling later so let me add an object initialization for tools cli parameter.

packages/create-llama/templates/components/engines/python/agent/__init__.py
leehuwuj: Nice, so it's already there. Let me update this.

packages/create-llama/templates/components/loaders/python/web/loader.py
leehuwuj: We are using `download_loader`, that function changed behavior that now run the `pip install` command so the package will be installed automatically. 
But in newer commits, i also tested our addPoetryDependencies function for tools and vectordbs so let me consider moving this wholesitereader to that function instead of using the download one, it'll be clearer and more consistent with other integration

packages/create-llama/templates/types/simple/fastapi/app/settings.py
leehuwuj: We have another settings.py file in each vectordbs template code which calls the init_base_settings and overrides the config. Basically, i just use the Settings as ServiceContext to keep our code structure as it is.

packages/create-llama/templates/components/loaders/python/web/loader.py
leehuwuj: I just removed the download_loader to use our installation instead. The dependencies for integrations should be declared in the code if it's not included in default llama_index package. Can check them out in the `getAdditionalDependencies` function.

packages/create-llama/index.ts
```diff
@@ -153,7 +153,7 @@ const program = new Commander.Command(packageJson.name)
     "--tools <tools>",
     `
 
-  Specify the tools you want to use by providing a comma-separated list. For example, 'google_search,wikipedia'. Use 'none' to not using any tools.
+  Specify the tools you want to use by providing a comma-separated list. For example, 'wikipedia.WikipediaToolSpec,google.GoogleSearchToolSpec'. Use 'none' to not using any tools.
```
leehuwuj, line 156: this is considered as a breaking change when we update from previous version (google_search -> google.GoogleSearchToolSpec, wikipedia -> wikipedia.WikipediaToolSpec).

We can create a shortcut name for it to keep the same tool name as it is but use the too_package.TooSpecClassName convention is more predictable.

PR #558, opened 2024-02-22
refactor: simplify generated python code

marcusschiesser:


packages/create-llama/templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -22,6 +23,8 @@ class _ChatData(BaseModel):
 async def chat(
     request: Request,
     data: _ChatData,
+    # TODO: @lee ensure that init_settings is called before get_chat_engine
```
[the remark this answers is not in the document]
leehuwuj: How about we call init_settings when starting the app at the main.py file?

packages/create-llama/templates/components/vectordbs/python/none/constants.py
```diff
@@ -1 +1,2 @@
 STORAGE_DIR = "storage"  # directory to cache the generated index
+DATA_DIR = "data"  # directory to store the data
```
marcusschiesser, line 2: Why do we need it? Isn't it just used in the loader?

leehuwuj: Ah yes, then the other vector db should have the DATA_DIR too. I also think file loader can use the DATA_DIR and STORAGE_DIR from environment variable instead, then we can remove the constants.py for none vectordb.

PR #559, opened 2024-02-22
feat: add base evaluator and correctness evaluator

EmanuelCampos:


examples/evaluation/correctness.ts
```diff
@@ -0,0 +1,42 @@
...
+
+However, general relativity, published in 1915, extended these ideas to include the effects of magnetism. According to general relativity, gravity is not a force between masses but rather the result of the warping of space and time by magnetic fields generated by massive objects. Massive objects, such as planets and stars, create magnetic fields that cause a curvature in spacetime, and smaller objects follow curved paths in response to this magnetic curvature. This concept is often illustrated using the analogy of a heavy ball placed on a rubber sheet with magnets underneath, causing it to create a depression that other objects (representing smaller masses) naturally move towards due to magnetic attraction.
+`;
+
+  const result = await evaluator.evaluate(query, response, [], reference);
```
marcusschiesser, line 37: how about using a parameter object?
```suggestion
  const result = await evaluator.evaluate({query, response, reference});
```

examples/evaluation/correctness.ts
```diff
@@ -0,0 +1,42 @@
+import {
```
marcusschiesser, line 1: How about adding
1. README for the new `evaluation` folder
2. docs for using `CorrectnessEvaluator`
?

packages/core/src/evaluation/types.ts
```diff
@@ -0,0 +1,24 @@
+export interface EvaluationResult {
```
marcusschiesser, line 1: how about using type instead of interface?

packages/core/src/evaluation/Correctness.ts
```diff
@@ -0,0 +1,101 @@
...
+        role: "user",
+        content: defaultUserPrompt({
+          query,
+          generatedAnswer: response,
+          referenceAnswer: reference || "(NO REFERENCE ANSWER SUPPLIED)",
+        }),
+      },
+    ];
+
+    let evalResponse = await this.serviceContext.llm.chat({
```
marcusschiesser, line 76: const

packages/core/src/evaluation/Correctness.ts
```diff
@@ -0,0 +1,101 @@
...
+      messages,
+    });
+
+    let [score, reasoning] = this.parserFunction(evalResponse.message.content);
```
marcusschiesser, line 80:
```suggestion
    const [score, reasoning] = this.parserFunction(evalResponse.message.content);
```

packages/core/src/evaluation/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  evaluate(
+    query: string,
+    response: string,
+    contexts?: string[],
+    reference?: string,
+  ): Promise<EvaluationResult>;
+  evaluateResponse(
+    query: string,
+    response: Response,
+  ): Promise<EvaluationResult>;
```
marcusschiesser, line 23: looks to me like this is not needed (if `evaluate` would use a parameter object)

packages/core/src/evaluation/Correctness.ts
marcusschiesser: Note: added lint rule to prefer const to `main` - better rebase

packages/core/src/evaluation/types.ts
EmanuelCampos: @marcusschiesser was missing a few lines of code that make the DX better with evaluateResponse

EmanuelCampos: wondering about change BaseEvaluator to a class then we can have this evaluateResponse on the base class

himself65:
My recent commit might affect your code. Please add .js extensions for each import.

And read https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext-1 to know what's the reason.

If you have any question please ping me

PR #560, opened 2024-02-22
build: leave code as-is

himself65:
- Remove bundler, just compile code to cjs/esm
- Switch to ESM by default to support modern ES features like tree shake, subpath import, and strict ts check
- Use vitest

We don't bundle the code since it puts lots of effort, and I think bundling is the downstream issue (like next.js bundler or vite)

And we keep the file as is so that everyone can import it directly

Followup: https://github.com/run-llama/LlamaIndexTS/pull/560#issuecomment-1962115894

---

Thanks for some TypeScript QA from @Jack-Works


packages/core/package.json
```diff
@@ -211,13 +99,14 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "test": "jest",
-    "build": "rm -rf ./dist && NODE_OPTIONS=\"--max-old-space-size=12288\" bunchee",
+    "build": "rm -rf ./dist && pnpm run build:esm && pnpm run build:cjs && pnpm run build:type",
```
marcusschiesser, line 102: as you're using vitest now, can we still trigger it from the core package?

marcusschiesser: why did you change to vitest?

packages/core/package.json
```diff
@@ -211,13 +99,14 @@
...
+    "build:esm": "swc src -d dist --strip-leading-paths --config-file .swcrc",
```
marcusschiesser, line 103: If i read this correctly, you're using swc to generate ESM/CJS code and tsc to emit declarations only.

Why this combination? What are alternatives?

packages/core/package.json
```diff
@@ -211,13 +99,14 @@
...
+    "build:cjs": "swc src -d dist/cjs --strip-leading-paths --config-file .cjs.swcrc",
+    "build:type": "tsc -p tsconfig.json",
     "postbuild": "pnpm run copy && pnpm run modify-package-json",
     "copy": "cp -r package.json CHANGELOG.md ../../README.md ../../LICENSE examples src dist/",
     "modify-package-json": "node ./scripts/modify-package-json.mjs",
     "prepublish": "pnpm run modify-package-json && echo \"please cd ./dist and run pnpm publish\" && exit 1",
-    "dev": "NODE_OPTIONS=\"--max-old-space-size=16384\" bunchee -w",
```
marcusschiesser, line 220: what are we going to use for dev?

packages/core/package.json
```diff
@@ -36,169 +38,55 @@
...
-    "@types/jest": "^29.5.12",
     "@types/lodash": "^4.14.202",
     "@types/node": "^18.19.14",
     "@types/papaparse": "^5.3.14",
     "@types/pg": "^8.11.0",
-    "bunchee": "^4.4.6",
+    "concurrently": "^8.2.2",
     "edit-json-file": "^1.8.0",
     "madge": "^6.1.0",
+    "tsimp": "^2.0.11",
```
marcusschiesser, line 51: where is this used?

packages/core/scripts/modify-package-json.mjs
```diff
@@ -4,7 +4,9 @@
  * so that it can be published to npm.
```
marcusschiesser, line 4: is there an alternative tool dealing with this concern (generating the dist folder)? this script might become hard to maintain with time

examples/ollama.ts
```diff
@@ -1,4 +1,4 @@
-import { Ollama } from "llamaindex";
+import { Ollama } from "llamaindex/llm/ollama";
```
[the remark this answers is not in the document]
himself65: just try if work

packages/core/scripts/modify-package-json.mjs
```diff
@@ -23,3 +25,28 @@ import fs from "node:fs/promises";
...
+    const r = relative(filePath, envFileTarget).replace("../", "./");
+    const replacedContent = content.replaceAll(
+      'require("#llamaindex/env")',
+      `require("${r}")`,
+    );
+    await fs.writeFile(filePath, replacedContent, "utf8");
+  }
+}
+
+const esmModule = {
```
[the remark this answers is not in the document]
himself65: forget to change the name

packages/core/src/PromptHelper.ts
```diff
@@ -1,12 +1,12 @@
-import { globalsHelper } from "./GlobalsHelper";
-import { SimplePrompt } from "./Prompt";
-import { SentenceSplitter } from "./TextSplitter";
+import { globalsHelper } from "./GlobalsHelper.js";
```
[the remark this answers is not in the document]
himself65: No, this should be the correct way

packages/core/package.json
himself65: nope

packages/core/scripts/modify-package-json.mjs
himself65: I think we can remove this script? because I found that public inside dist is not very useful as people are not using very old node.js

packages/core/package.json
himself65: swc could also support polyfill and generate to different target if we want. TSC only generate based on tsconfig and   I don't think add multiple tsconfigs is a good choice

packages/core/package.json
himself65: Jest doesn't support ESM and needs lots of work on configuration. ALSO not support typescript natively

PR #583, opened 2024-02-27
feat: JSON Query Engine

EmanuelCampos:
missing:
- [x] add tests to output parser
- [ ] add to experiemental folder
- [ ] experiemental documentation


packages/core/src/engines/query/JSONQueryEngine.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (this.verbose) {
+      console.log(
+        `> JSONPath Instructions:\n\`\`\`\n${jsonPathResponseStr}\n\`\`\`\n`,
+      );
+    }
+
+    const jsonPathOutput = await this.outputProcessor({
+      llmOutput: jsonPathResponseStr.text,
+      jsonValue: this.jsonValue,
+      ...this.outputKwargs,
```
marcusschiesser, line 144: no kwargs in TS and type `OutputProcessor` does just support `llmOutput` and `jsonValue` as params

I would leave it out

EmanuelCampos:
Moved to #613

PR #584, opened 2024-02-28
fix: lazy import pg

himself65:


marcusschiesser:
@himself65 that's great!

Issue #563, opened 2024-02-24
Enable typescript-eslint rule


himself65:
I got 1k+ lint error after enable the strict lint rules

marcusschiesser:
@himself65 great idea!

PR #585, opened 2024-02-28
feat: improve type

himself65:
Related: https://github.com/run-llama/LlamaIndexTS/issues/563


marcusschiesser (commented):
@himself65 LGTM. What risks do you see? Some of my concerns:

1. Changing the vector DB code is risky (as they are not covered well by tests)
2. `jsonToNode` is using `type` - is your change still working?

himself65:
> Changing the vector DB code is risky (as they are not covered well by tests)

The main reason is I change `Record<string, any>` to `Record<string, unknown>.` I want to improve the type checking, step by step

himself65:
Let me test it in real production and see what it would affect

Issue #564, opened 2024-02-25
`PineconeVectorStore` support loading data from Python and reading from TS


david1542:
Hey everyone.

I'm using Pinecone as a vector DB for my RAG application. I've loaded some documents into it with python code, and now I'm trying to query the DB with TypeScript. However, the nodes in the query result always contains an empty "text" field, which prevents me from using the texts in my prompt.

Here is how I query the DB:
```ts
const embedModel = new OpenAIEmbedding({
  apiKey: process.env.OPENAI_API_KEY as string,
  model: "text-embedding-3-large",
  dimensions: 768,
});
const embedding = await embedModel.getTextEmbedding("my query");
const pcvs = new PineconeVectorStore({
  indexName: "my-index",
});
const result = await pcvs.query({
  queryEmbedding: embedding,
  similarityTopK: 5,
  mode: VectorStoreQueryMode.DEFAULT,
});

console.log(result.nodes); // Contains 5 items with scores but empty text field
```

Here is how I loaded them into the DB:
```python
pc = Pinecone(api_key=pinecone_api_key)
pc_index = pc.Index(name=organization_id)

vector_store = PineconeVectorStore(pinecone_index=pc_index)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
embed_model = OpenAIEmbedding(
    api_key=openai_api_key,
    model="text-embedding-3-large",
    dimensions=768,
)

VectorStoreIndex.from_documents(
    total_documents,
    vector_store=pc_index,
    show_progress=True,
    embed_model=embed_model,
    storage_context=storage_context,
)
```

Here is an image during a debugging session:
<img width="480" alt="image" src="https://github.com/run-llama/LlamaIndexTS/assets/9879252/8f28ce2c-affe-4cba-af6c-08a213c92e9c">

I think this is related to #552 but I'm not sure.

I'd would love to get some help on this :)

marcusschiesser:
@david1542 we just tested this with the new 0.1.19 release. Please have a try and reopen if necessary

Issue #552, opened 2024-02-20
PineconeVectorStore always returns 'Empty Response'


owencraston:
## What is the issue?
I am trying to use pinecone db along with llama index to create a RAG application. I was following [your example](https://github.com/run-llama/LlamaIndexTS/blob/main/examples/pinecone-vector-store/query.ts) closely but no matter what query I pass into my `BaseQueryEngine` I get the same response...

```
Response {
  metadata: {},
  response: 'Empty Response',
  sourceNodes: [
    Document {
      id_: '9dab9a39-c8fd-4c1c-a51a-33c5277c64fd',
      metadata: [Object],
      excludedEmbedMetadataKeys: [],
      excludedLlmMetadataKeys: [],
      relationships: {},
      hash: 'hasqQpeIQjqCSkn5snCmwKG2Ty2fM1hw4IAebcnoyO8=',
      text: '',
      embedding: [Array],
      textTemplate: '',
      metadataSeparator: '\n'
    },
    Document {
      id_: 'f76c78cd-84bb-4b81-826c-feb802171347',
      metadata: [Object],
      excludedEmbedMetadataKeys: [],
      excludedLlmMetadataKeys: [],
      relationships: {},
      hash: 'zzoar6NL5v1WZM0i05r/9iT7cR3REO+AWtGeNvxcwxk=',
      text: '',
      embedding: [Array],
      textTemplate: '',
      metadataSeparator: '\n'
    }
  ]
}
```

Here is the code for my next.js application.

```ts
import {
  PineconeVectorStore,
  serviceContextFromDefaults,
  SimpleNodeParser, SentenceSplitter, VectorStoreIndex,
  BaseQueryEngine,
} from "llamaindex";

async function initializeChatEngine(): Promise<BaseQueryEngine> {
  const pineconeVectorStore = new PineconeVectorStore();
  console.log("Pinecone Vector Store initialized successfully. ", await pineconeVectorStore.index());
  const serviceContext = serviceContextFromDefaults({
    nodeParser: new SimpleNodeParser({
      textSplitter: new SentenceSplitter(),
      includeMetadata: true,
      includePrevNextRel: true,
    }),
  });
  const index = await VectorStoreIndex.fromVectorStore(pineconeVectorStore, serviceContext);

  return index.asQueryEngine();
}


export async function getChatEngine(): Promise<BaseQueryEngine> {
  return initializeChatEngine();
}

```

this function then gets called here...

```ts
import { z } from "zod";
[...]

Issue #529, opened 2024-02-08
Add support for Pinecone namespaces


eburnette:
Pinecone takes a namespace (PINECONE_NAME_SPACE) that can either be empty (default) or a string. I didn't see a way in llamaindex.ts to pass that in when inserting or querying, and setting the environment variable had no effect, so it's currently always using the default namespace.

Documentation: https://docs.pinecone.io/docs/namespaces

abeauvois:
so a workaround could be:
```ts
    const pcvs = new PineconeVectorStore({ indexName: 'my-index' });

    const ctx = serviceContextFromDefaults();
    const index = await VectorStoreIndex.fromVectorStore(pcvs, ctx);

    // list all namespaces (the default one is '')
    await index.describeIndexStats();
    // Response: {
    //   namespaces: {
    //     '': { recordCount: 10 }
    //     foo: { recordCount: 2000 },
    //     bar: { recordCount: 2000 }
    //   },
    //   dimension: 1536,
    //   indexFullness: 0,
    //   totalRecordCount: 4010
    // }

    // Add a namespace arbitray named "ns-1" to our index
    index.namespace('ns-1');

    console.log(index.namespace('ns-1').target); // { index: 'my-index', namespace: 'ns-1', indexHostUrl: undefined }

    // Now perform index operations in the targeted index and namespace
    await index.fetch(['3']);

    await index
      .namespace('ns-1')
      .query({ topK: 3, vector: [0.22, 0.66] });
    // {
    //   matches: [
    //     {
    //       id: '556',
    //       score: 1.00000012,
    //       values: [],
    //       sparseValues: undefined,
    //       metadata: undefined
    //     },
    //     {
    //       id: '137',
    //       score: 1.00000012,
    //       values: [],
    //       sparseValues: undefined,
    //       metadata: undefined
    //     },
    //     {
    //       id: '129',
    //       score: 1.00000012,
    //       values: [],
    //       sparseValues: undefined,
    //       metadata: undefined
    //     }
    //   ],
    //   namespace: 'ns-1',
    //   usage: {
    //     readUnits: 5
    //   }
    // }

```

Should be necessary to improve the queryEngine params to accept the concept of namespaces? @marcusschiesser

david1542:
I also need this solution :) I'm currently using Pinecone's index object directly:
```ts
const embedModel = new OpenAIEmbedding({
  apiKey: process.env.OPENAI_API_KEY as string,
[...]

PR #586, opened 2024-02-28
feat: enhance pinecone usage

thucpn:
- support for Pinecone namespaces #529
- lazy load pinecone store
- fix empty text response: #552
- may fix issue #564


packages/core/src/storage/vectorStore/PineconeVectorStore.ts
```diff
@@ -148,10 +152,12 @@ export class PineconeVectorStore implements VectorStore {
     };
 
     const idx = await this.index();
-    const results = await idx.query(options);
+    const results = await idx.namespace(this.namespace).query(options);
```
marcusschiesser, line 155: also works with `this.namespace=''`?

marcusschiesser (approved):

PR #589, opened 2024-02-29
feat: add pipeline.register

marcusschiesser:


packages/core/src/cloud/config.ts
```diff
@@ -0,0 +1,74 @@
...
+  transformations?: TransformComponent[];
+  inputNodes?: BaseNode[];
+};
+
+function getTransformationConfig(
+  transformation: TransformComponent,
+): PlatformApi.ConfiguredTransformationItem {
+  if (transformation instanceof SimpleNodeParser) {
+    return {
+      configurableTransformationType: "SIMPLE_FILE_NODE_PARSER",
```
marcusschiesser, line 19: @himself65 have to use a string constant here as using the value from `PlatformApi.ConfigurableTransformationNames` would require a non-type import of `PlatformApi` which would be converted to a `require` call - any idea how to use this in a type safe manner with our current config?

packages/core/src/cloud/config.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (transformation instanceof OpenAIEmbedding) {
+    return {
+      configurableTransformationType: "OPENAI_EMBEDDING",
+      component: {
+        modelName: transformation.model,
+        embedBatchSize: transformation.embedBatchSize,
+        dimensions: transformation.dimensions,
+      },
+    };
+  }
```
marcusschiesser, line 35: @yisding @himself65 @EmanuelCampos, any alternative ideas for converting a transformation to its cloud config?

We could use a generic approach using `JSON.stringify` but then we might 
1. need to clean the serialized JSON afterwards AND
2. we might want to send parameters of sub-components to the cloud (e.g. chunksize of the textsplitter) which would be missing
- that's why I didn't do this

Adding a `toConfig` function to `TransformerComponent` is an alternative, but might be too much hassle.

packages/core/src/cloud/config.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+function getTransformationConfig(
+  transformation: TransformComponent,
+): PlatformApi.ConfiguredTransformationItem {
+  if (transformation instanceof SimpleNodeParser) {
+    return {
+      configurableTransformationType: "SENTENCE_AWARE_NODE_PARSER",
+      component: {
+        // TODO: API returns 422 if these parameters are included
```
marcusschiesser, line 21: @yisding this parameters are not used by the API

packages/core/src/cloud/config.ts
```diff
@@ -0,0 +1,81 @@
...
+        // includePrevNextRel: transformation.includePrevNextRel,
+      },
+    };
+  }
+  if (transformation instanceof OpenAIEmbedding) {
+    return {
+      configurableTransformationType: "OPENAI_EMBEDDING",
+      component: {
+        modelName: transformation.model,
+        apiKey: transformation.apiKey,
```
marcusschiesser, line 34: @yisding apiKey is not used by the API

yisding:
@marcusschiesser sorry for dropping the ball here. I'm going to merge this but we should get all of the other transformations in sooner than later.

PR #591, opened 2024-03-01
feat: Add use LlamaParse option to create-llama

leehuwuj:


packages/create-llama/helpers/python.ts
```diff
@@ -244,14 +245,25 @@ export const installPythonTemplate = async ({
...
-          : path.join(compPath, "loaders", "python", dataSourceType);
-      await copy("**", enginePath, {
-        parents: true,
-        cwd: loaderPath,
-      });
+      let loaderPath;
+      if (dataSourceType === "file" || dataSourceType === "folder") {
+        const loaderDir = path.join(compPath, "loaders", "python", "file");
+        const dataSourceConfig = dataSource?.config as FileSourceConfig;
+        if (dataSourceConfig.useLlamaParse) {
```
marcusschiesser, line 252: instead of deciding which single file to copy, can we use a dedicated folder `"llamaparse"` for LlamaParse?

So we would end up using;
`path.join(compPath, "loaders", "python", "llamaparse")`

Reason: We want to stick to our concept that we copy folders in create-llama

packages/create-llama/questions.ts
```diff
@@ -521,6 +526,63 @@ export const askQuestions = async (
...
+          : dataSourceConfig.path &&
+            path.extname(dataSourceConfig.path) === ".pdf");
+
+      // Ask if user wants to use LlamaParse
+      if (askingLlamaParse) {
+        const { useLlamaParse } = await prompts(
+          {
+            type: "toggle",
+            name: "useLlamaParse",
+            message: "Would you like to use LlamaParse?",
```
marcusschiesser, line 553:
```suggestion
            message: "Would you like to use LlamaParse (improved parser for RAG - requires API key)?",
```

packages/create-llama/helpers/python.ts
leehuwuj: Yes, let me update. I've considered llamaparse as a data source then no need to update this code but end-up with this because i wanted it's as an option for file/folder data source. 
Let me reimplement it.

marcusschiesser (approved):

PR #597, opened 2024-03-02
feat: OpenAI Agent Stream

EmanuelCampos:


examples/agent/stream_openai_agent.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // Create an OpenAIAgent with the function tools
+  const agent = new OpenAIAgent({
+    tools: [functionTool, functionTool2],
+    verbose: false,
+  });
+
+  const stream = await agent.streamChat({
+    message: "Divide 16 by 2 then add 20",
+  });
```
marcusschiesser, line 67: we shouldn't use a dedicated streamChat method (like we're doing in Python), but instead, we should override the chat method for streaming, so this should look like this:

```suggestion
  const stream = await agent.chat({
    message: "Divide 16 by 2 then add 20",
    stream: true
  });
```

PR #602, opened 2024-03-04
Feat: support agent in Typescript templates

thucpn:
✅ support stream chat in OpenAIAgentWorker
✅ update engine of NextJs streaming template to use QueryEngineTool and OpenAIAgent
✅ update route handler of NextJs streaming template to response stream
- add question to select other tools and load them in OpenAIAgent
- support Express streaming template to use OpenAIAgent
- support other vectordbs working with OpenAIAgent
- use React Server Component (from "ai/rsc") to make awesome UI

### DIFF
--- a/packages/create-llama/templates/types/streaming/nextjs/app/api/chat/route.ts
+++ b/packages/create-llama/templates/types/streaming/nextjs/app/api/chat/route.ts
@@ -1,5 +1,10 @@
 import { StreamingTextResponse } from "ai";
-import { ChatMessage, MessageContent, OpenAI } from "llamaindex";
+import {
+  ChatMessage,
+  ChatResponseMode,
+  MessageContent,
+  OpenAI,
+} from "llamaindex";
 import { NextRequest, NextResponse } from "next/server";
 import { createChatEngine } from "./engine";
 import { LlamaIndexStream } from "./llamaindex-stream";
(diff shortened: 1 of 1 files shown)


EmanuelCampos:
We have OpenAI streaming coming on this PR:

https://github.com/run-llama/LlamaIndexTS/pulls

But thanks for the contribution, if you prefer to remove the commits related to streaming then we can review the rest, sounds pretty good

packages/create-llama/templates/components/vectordbs/typescript/none/index.ts
```diff
@@ -32,13 +34,22 @@ async function getDataSource(llm: LLM) {
   });
 }
 
-export async function createChatEngine(llm: LLM) {
+export async function createChatEngine(llm: OpenAI) {
```
marcusschiesser, line 37: @thucpn if the user doesn't select a tool in (questions.ts) then we will still keep the `ContextChatEngine`, so we need two components: one using ContextChatEngine and one the OpenAIAgent

marcusschiesser: that's why Python has two chat engines, see https://github.com/run-llama/LlamaIndexTS/tree/main/packages/create-llama/templates/components/engines/python

PR #604, opened 2024-03-04
feat: Claude 3

EmanuelCampos:


packages/core/src/llm/LLM.ts
```diff
@@ -635,6 +632,7 @@ export class Anthropic extends BaseLLM {
   maxRetries: number;
   timeout?: number;
   session: AnthropicSession;
+  systemPrompt?: string;
```
yisding, line 635: We want to maintain the same API for other LLMs for systemPrompt, so we'll need to rejigger the logic for Anthropic.

packages/core/src/llm/LLM.ts
```diff
@@ -674,21 +673,13 @@ export class Anthropic extends BaseLLM {
     };
   }
 
-  mapMessagesToPrompt(messages: ChatMessage[]) {
```
yisding, line 677: We need to support both the "legacy" models as well as Claude 3, so need to think about how to make this work for both.

packages/core/src/llm/LLM.ts
```diff
@@ -698,49 +710,67 @@ export class Anthropic extends BaseLLM {
...
+      ...(systemPrompt && { system: systemPrompt }),
+    });
 
     let idx_counter: number = 0;
     for await (const part of stream) {
-      //TODO: LLM Stream Callback, pending re-work.
+      const content =
+        part.type === "content_block_delta" ? part.delta.text : "";
+
+      if (typeof content !== "string") continue;
```
himself65, line 770: https://docs.anthropic.com/claude/reference/messages-streaming#raw-http-stream-response

Only return content when `content_block_delta`

Don't emit empty content

packages/core/src/llm/LLM.ts
```diff
@@ -674,21 +687,20 @@ export class Anthropic extends BaseLLM {
...
+      return AVAILABLE_ANTHROPIC_MODELS_WITHOUT_DATE[model];
+    }
+    return model;
+  };
+
+  formatMessages(messages: ChatMessage[]) {
+    return messages.map((message) => {
+      return {
+        content: message.content,
+        role: message.role as "user" | "assistant",
```
himself65, line 701: This will throw error if there is non user or assistant type like, "system"

packages/core/src/llm/LLM.ts
```diff
@@ -613,10 +608,28 @@ If a question does not make any sense, or is not factually coherent, explain why
...
+
 export const ALL_AVAILABLE_ANTHROPIC_MODELS = {
-  // both models have 100k context window, see https://docs.anthropic.com/claude/reference/selecting-a-model
-  "claude-2": { contextWindow: 200000 },
-  "claude-instant-1": { contextWindow: 100000 },
+  ...ALL_AVAILABLE_ANTHROPIC_V2_MODELS,
+  ...ALL_AVAILABLE_V3_MODELS,
+};
+
+const AVAILABLE_ANTHROPIC_MODELS_WITHOUT_DATE: { [key: string]: string } = {
```
himself65, line 630: use `satisfies`

examples/anthropic_stream.ts
```diff
@@ -0,0 +1,23 @@
+import { Anthropic } from "llamaindex";
```
himself65, line 1: put them into a single directory

yisding (approved):
s. Looks good!

packages/core/src/llm/LLM.ts
```diff
@@ -613,10 +608,28 @@ If a question does not make any sense, or is not factually coherent, explain why
...
+  "claude-3-opus": "claude-3-opus-20240229",
```
himself65, line 631: `claude-3-opus-20240229` should be the correct model name, I would like to as-is

https://docs.anthropic.com/claude/reference/messages-streaming#streaming-with-sdks

packages/core/src/llm/LLM.ts
```diff
@@ -613,10 +608,28 @@ If a question does not make any sense, or is not factually coherent, explain why
   }
 }
 
+export const ALL_AVAILABLE_ANTHROPIC_V2_MODELS = {
+  "claude-2.1": {
+    contextWindow: 200000,
+  },
+  "claude-instant-1.2": {
```
himself65, line 615: v1.2 is not v2

himself65 (changes requested):

packages/core/src/llm/LLM.ts
yisding: @himself65 I asked Emanuel to add this feature. We're seeing other people do the same.

packages/core/src/llm/LLM.ts
yisding: Fair point, let's call this ANTHROPIC_LEGACY_MODELS then? And add claude-2.0 for completeness.

packages/core/src/llm/LLM.ts
EmanuelCampos: Currently Anthropic only support user or assistant, they don't support system or others

yisding: This is just a TS cast right?

But that's a good point that we should maybe check that all remaining messages are user or assistant, and not one of the other message types (and if they are throw an error)

But with the new changes, this function should currently be a noop.

examples/anthropic_stream.ts
yisding: Are you saying we should do a subdirectory level import? I think it would be a good idea to document, although not sure we want to make it standard across our examples just yet.

yisding:
Screenshot 2024-03-04 at 2 57 44 PM

Looks like all of the old models should support the new API. Let's test it but should be safe to use the new API exclusively.

PR #605, opened 2024-03-04
chore: bump version

himself65:


EmanuelCampos (changes requested):
For Anthropic we have https://github.com/run-llama/LlamaIndexTS/pull/604

PR #608, opened 2024-03-05
feat: add embedding model option to create-llama

leehuwuj:


packages/create-llama/questions.ts
```diff
@@ -419,6 +420,7 @@ export const askQuestions = async (
   if (!program.model) {
     if (ciInfo.isCI) {
       program.model = getPrefOrDefault("model");
+      program.embeddingModel = getPrefOrDefault("embeddingModel");
```
marcusschiesser, line 423: why do we need this line? it's set below again

marcusschiesser (commented):
Looks good, missing the typescript code to select the model

marcusschiesser (approved):

PR #613, opened 2024-03-06
feat: experimental package + json query engine

EmanuelCampos:
fixing types not being generated


marcusschiesser (commented):
Great job - looks good to me!

package.json
```diff
@@ -14,6 +14,7 @@
     "release": "pnpm run build:release && changeset publish",
     "new-llamaindex": "pnpm run build:release && changeset version --ignore create-llama",
     "new-create-llama": "pnpm run build:release && changeset version --ignore llamaindex --ignore @llamaindex/core-test",
+    "new-experimental": "pnpm run build:release && changeset version --ignore create-llama",
```
marcusschiesser, line 17: @EmanuelCampos this operation is then the same as `new-llamaindex`, you probably want to ignore more or leave it out (which would mean experimental versioning would depend on core)

packages/experimental/package.json
```diff
@@ -0,0 +1,70 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/run-llama/LlamaIndexTS.git",
+    "directory": "packages/experimental"
+  },
+  "scripts": {
+    "build": "rm -rf ./dist && pnpm run build:esm && pnpm run build:cjs && pnpm run build:type",
+    "build:esm": "swc src -d dist --strip-leading-paths --config-file .swcrc",
+    "build:cjs": "swc src -d dist/cjs --strip-leading-paths --config-file .cjs.swcrc",
```
himself65, line 50: https://github.com/run-llama/LlamaIndexTS/pull/620

package.json
EmanuelCampos: oh true hehe, mb

PR #614, opened 2024-03-06
fix: hardcode "en" as default language for llama-parse and use llama cloud key from env

leehuwuj:


packages/create-llama/helpers/index.ts
```diff
@@ -95,22 +95,34 @@ const createEnvLocalFile = async (
...
+  llamaCloudKey?: string,
+) {
   if (packageManager) {
     const runGenerate = `${cyan(
       framework === "fastapi"
         ? "poetry run python app/engine/generate.py"
         : `${packageManager} run generate`,
     )}`;
     const hasOpenAiKey = openAiKey || process.env["OPENAI_API_KEY"];
+    const hasLlamaCloudKey = (dataSource?.config as FileSourceConfig)
```
leehuwuj, line 114: I've considered naming it `checkLlamaCloudKey`, but `hasLlamaCloudKey` looks better in this context.

packages/create-llama/helpers/index.ts
```diff
@@ -95,22 +95,34 @@ const createEnvLocalFile = async (
...
+      ?.useLlamaParse
+      ? llamaCloudKey || process.env["LLAMA_CLOUD_API_KEY"]
+      : true;
     const hasVectorDb = vectorDb && vectorDb !== "none";
     if (framework === "fastapi") {
-      if (hasOpenAiKey && !hasVectorDb && isHavingPoetryLockFile()) {
+      if (
+        hasOpenAiKey &&
+        hasLlamaCloudKey &&
```
marcusschiesser, line 122: how about naming it like this?
```suggestion
        openAiKeyConfigured &&
        llamaCloudKeyConfigured &&
```

packages/create-llama/questions.ts
```diff
@@ -605,11 +612,8 @@ export const askQuestions = async (
           {
             type: "text",
             name: "llamaCloudKey",
-            message: "Please provide your LlamaIndex Cloud API key:",
-            validate: (value) =>
-              value
-                ? true
-                : "LlamaIndex Cloud API key is required. You can get it from: https://cloud.llamaindex.ai/api-key",
+            message:
+              "Please provide your LlamaIndex Cloud API key (leave blank to skip):",
```
marcusschiesser, line 616: do we add `https://cloud.llamaindex.ai/api-key` to the .env file?

packages/create-llama/questions.ts
```diff
@@ -222,11 +222,18 @@ export const askQuestions = async (
         ];
 
         const hasOpenAiKey = program.openAiKey || process.env["OPENAI_API_KEY"];
+        // If using LlamaParse, require LlamaCloud API key
+        const hasLlamaCloudKey = (
```
marcusschiesser, line 226: if you decide to change the var names, here too please

packages/create-llama/helpers/index.ts
leehuwuj: It's nice, thanks for the suggestion. I'm updating it, naming is hard :D

packages/create-llama/questions.ts
leehuwuj: I made an update to add a minor config guide to .env file. Now, if user left this option as blank then the .env will look like this:
```
# Please obtain the Llama Cloud API key from https://cloud.llamaindex.ai/api-key 
# and set it to the LLAMA_CLOUD_API_KEY variable below.
# LLAMA_CLOUD_API_KEY=
```

packages/create-llama/helpers/index.ts
```diff
@@ -54,6 +54,10 @@ const createEnvLocalFile = async (
 
   if (opts?.llamaCloudKey) {
     content += `LLAMA_CLOUD_API_KEY=${opts?.llamaCloudKey}\n`;
+  } else {
+    content += `# Please obtain the Llama Cloud API key from https://cloud.llamaindex.ai/api-key
```
marcusschiesser, line 58: this would copy this content all the time - even if using llamacloud is not configured, right?

leehuwuj: Thanks, it's a good catch.

marcusschiesser (approved):

PR #616, opened 2024-03-06
Use parameter object for retrieve function of Retriever 

marcusschiesser:
**Note**: As this is breaking, we're using a minor changeset
**Reason**: To align usage with `query` function of `QueryEngine`

This will align the LlamaCloud code, so we can do:

```
import { LlamaCloudIndex } from "llamaindex";

const index = new LlamaCloudIndex({
    name: "test",
    projectName: "default",
    apiKey: "llx-...",
  });
const nodes = await index.asRetriever().retrieve({ query });
const response = await index.asQueryEngine().query({ query });
```

### DIFF
--- a/.changeset/strange-waves-sit.md
+++ b/.changeset/strange-waves-sit.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": minor
+---
+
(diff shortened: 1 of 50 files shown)


yisding (approved):
Makes sense. It is a breaking change but better to do it sooner than later.

PR #618, opened 2024-03-07
Support NextJS edge runtime

marcusschiesser:
This PR is working with NextJS edge run-time (tested with Chat LlamaIndex (using `pnpm build && pnpm start` - `pnpm dev` doesn't work yet))

Changes:
1. Make core edge compatible
2. Removed non-edge compatible deps

Next steps:
Discuss how to deal with non-edge compatible deps - see comments


packages/core/src/readers/SimpleDirectoryReader.ts
```diff
@@ -37,10 +35,10 @@ export class TextFileReader implements BaseReader {
 
 export const FILE_EXT_TO_READER: Record<string, BaseReader> = {
   txt: new TextFileReader(),
-  pdf: new PDFReader(),
+  // pdf: new PDFReader(),
```
marcusschiesser, line 38: @himself65 I remember you removed the pdf.js based implementation - as far as I recall the reason was the bundle size, correct? But as pdf.js is working in the browser - should be possible to run it on edge runtime right?

packages/core/src/storage/index.ts
```diff
@@ -9,10 +9,6 @@ export { SimpleIndexStore } from "./indexStore/SimpleIndexStore.js";
 export * from "./indexStore/types.js";
 export { SimpleKVStore } from "./kvStore/SimpleKVStore.js";
 export * from "./kvStore/types.js";
-export { AstraDBVectorStore } from "./vectorStore/AstraDBVectorStore.js";
-export { ChromaVectorStore } from "./vectorStore/ChromaVectorStore.js";
-export { MongoDBAtlasVectorSearch } from "./vectorStore/MongoDBAtlasVectorStore.js";
-export { PGVectorStore } from "./vectorStore/PGVectorStore.js";
 export { PineconeVectorStore } from "./vectorStore/PineconeVectorStore.js";
```
marcusschiesser, line 12: @himself65 
Suggestion: 
Remove all the vector stores here (except SimpleVectorStore) and the user has to import the vector store that he needs
If we do this what are the problems? e.g. how do we make it run with a CJS project?

packages/core/src/readers/SimpleDirectoryReader.ts
```diff
@@ -37,10 +35,10 @@ export class TextFileReader implements BaseReader {
 
 export const FILE_EXT_TO_READER: Record<string, BaseReader> = {
```
marcusschiesser, line 36: @himself65 

**Suggestion**, we include a minimal `FILE_EXT_TO_READER` as default (without the readers that are not working with edge) and make it easy for the dev to add whatever readers that they need, e.g.:

To use all readers (importing `ALL_READERS` would import the non-edge readers):
```
const documents = await new SimpleDirectoryReader({
    readers: ALL,
  }).loadData({
    directoryPath: "node_modules/llamaindex/examples",
  })
```

or to use specific readers
```
const documents = await new SimpleDirectoryReader({
    readers: {
    "pdf": PDFReader
    },
  }).loadData({
    directoryPath: "node_modules/llamaindex/examples",
  })
```

packages/core/src/readers/index.ts
```diff
@@ -1,12 +1,10 @@
 export * from "./AssemblyAIReader.js";
 export * from "./CSVReader.js";
```
marcusschiesser, line 2: removed here the readers that are not working with edge

packages/core/src/readers/SimpleDirectoryReader.ts
himself65: no, there is no directory in edge runtime, you cannot access file system

himself65 (commented):
My suggestion would be change "index.edge-ligtht.js" and only export core module. 

But no matter how, this would need more effort because need to change many path import

himself65:
> But no matter how, this would need more effort because we need to change many path import

For example, some files would indirectly import vectorStore.js

```
index.edge-light.js -> ... -> verctorStore.js
```

packages/core/src/storage/index.ts
himself65: same in cjs module `require(llamaindex/path/to/vectorStore)`

packages/core/src/readers/SimpleDirectoryReader.ts
marcusschiesser: AFAIK you have read access to the file system, but not write access

packages/core/src/readers/SimpleDirectoryReader.ts
himself65: Pdfjs is too large not possible in edge runtime

packages/core/package.json
```diff
@@ -54,6 +54,18 @@
...
+        "types": "./dist/type/readers/index.d.ts",
+        "default": "./dist/readers/index.edge-light.js"
+      }
+    },
+    "./storage": {
+      "edge-light": {
+        "types": "./dist/type/storage/index.d.ts",
+        "default": "./dist/storage/index.edge-light.js"
+      }
+    },
```
marcusschiesser, line 68: @himself65 I added edge-specific code, which is transpiled but not used. My nextjs project build still shows this import trace:
```
Import trace for requested module:
../LlamaIndexTS/node_modules/.pnpm/mammoth@1.6.0/node_modules/mammoth/lib/index.js
../LlamaIndexTS/packages/core/dist/readers/DocxReader.js
../LlamaIndexTS/packages/core/dist/readers/index.js
../LlamaIndexTS/packages/core/dist/index.js
```
any idea?

himself65: yes because you didn't trigger the "subpath exports". You have to manually import from `import 'llamaindex/storage'` for every import path. otherwise it not working

himself65: So that's why I suggest splitting them into `@llamaindex/xxx`.

himself65: Another idea is to use [subpath import](https://nodejs.org/api/packages.html#subpath-imports) but it only works in the nodejs latest version

packages/edge/package.json
```diff
@@ -0,0 +1,80 @@
...
+  },
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "types": "./dist/type/index.edge.d.ts",
+  "main": "./dist/index.edge.js",
+  "exports": {
+    "./readers/SimpleDirectoryReader": {
+      "import": {
+        "types": "./dist/type/readers/SimpleDirectoryReader.edge.d.ts",
```
marcusschiesser, line 51: We could call this EdgeDirectoryReader then we won't need this edge specific export

packages/edge/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@llamaindex/edge",
+  "version": "0.1.21",
+  "license": "MIT",
+  "type": "module",
+  "dependencies": {
+    "@anthropic-ai/sdk": "^0.15.0",
```
marcusschiesser, line 7: best we generate this file from core/package.json and copy its dependencies and version

packages/core/src/index.ts
```diff
@@ -1,33 +1,3 @@
...
-export * from "./extractors/index.js";
-export * from "./indices/index.js";
-export * from "./ingestion/index.js";
-export * from "./llm/index.js";
-export * from "./nodeParsers/index.js";
-export * from "./objects/index.js";
-export * from "./postprocessors/index.js";
-export * from "./prompts/index.js";
+export * from "./index.edge.js";
 export * from "./readers/index.js";
```
marcusschiesser, line 2: here's the trick: edge is neither importing readers nor storage

himself65 (changes requested):
seems not working

himself65:
> seems not working

working, but some deps seem not correct

packages/edge/package.json
```diff
@@ -0,0 +1,80 @@
+{
```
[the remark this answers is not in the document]
marcusschiesser: yes, idea is to just copy the build from core and then use different exports in this package

packages/core/e2e/test-edge-runtime/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "llamaindex": "workspace:*",
```
marcusschiesser, line 12: should use `@llamaindex/edge` package

packages/core/package.json
```diff
@@ -60,14 +60,12 @@
     ".": {
       "import": {
         "types": "./dist/type/index.d.ts",
+        "edge-light": "./dist/index.edge.js",
```
marcusschiesser, line 63: @himself65 the idea is that `@llamaindex/edge` is working with edge, but `llamaindex` is not - that's why i removed `edge-light` from the core package

PR #619, opened 2024-03-07
feat: Add start in VSCode option and support python for dev container

leehuwuj:


packages/create-llama/helpers/devcontainer.ts
```diff
@@ -0,0 +1,44 @@
+import fs from "fs";
+import path from "path";
+import { TemplateFramework } from "./types";
+
```
marcusschiesser, line 4: i think you can directly import the JSON like this:
```suggestion
import devcontainerJson from "../templates/devcontainer.json"
```

packages/create-llama/templates/devcontainer.json
```diff
@@ -27,5 +27,8 @@
       }
     }
   },
+  "containerEnv": {
+    "POETRY_VIRTUALENVS_CREATE": "false"
```
marcusschiesser, line 31: great idea!

packages/create-llama/helpers/devcontainer.ts
marcusschiesser: but then you would make to need a copy of the object, probably not worth it

marcusschiesser (approved):

PR #621, opened 2024-03-07
Feat: add tools module

thucpn:


packages/tools/assembly/query-engine-tool/index.ts
```diff
@@ -0,0 +1,36 @@
+import {
```
marcusschiesser, line 1: @thucpn you can leave out the query engine tool - we won't convert this to WASM (as it has to many dependencies from the core framework)

packages/tools/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "node tests",
+    "asbuild:debug": "asc assembly/index.ts --target debug",
+    "asbuild:release": "asc assembly/index.ts --target release",
+    "asbuild": "npm run asbuild:debug && npm run asbuild:release",
+    "start": "npx serve ."
+  },
+  "author": "",
+  "license": "ISC",
```
marcusschiesser, line 14: should be same as core package

packages/tools/index.ts
```diff
@@ -0,0 +1,52 @@
+import asc from "assemblyscript/bin/asc";
+import fs from "fs";
+import { BaseTool } from "./assembly/base";
```
marcusschiesser, line 3: is it ok to import assemblyscript in TS?

packages/tools/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      },
+      {
+        id: "wiki-tool",
+        name: "Wikipedia Summary",
+      },
+    ];
+  }
+
+  async downloadTool(toolId: string, storageDir: string = "build") {
+    const { error, stdout, stderr } = await asc.main([
```
marcusschiesser, line 31: isn't that starting the compiler?
the build process should compile the `wasm` files for each tool and then the file should be just loaded here

packages/tools/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+type ToolInfo = {
+  id: string;
+  name: string;
+};
+
+export interface IToolFactory {
+  getToolList(): ToolInfo[];
+  downloadTool(toolId: string, storageDir?: string): void;
+  loadTool(filePath: string): Promise<BaseTool>;
```
marcusschiesser, line 13: good idea to have a factory! How to simplify it's usage for the caller?
```suggestion
  getToolList(): ToolInfo[];
  loadTool(toolId: string): Promise<BaseTool>;
```

marcusschiesser: but as a user of the tools package, I would probably prefer to directly use the tool (the factory is still good for enumeration and dynamically creating a tool), this means we need a wrapper for each tool that I can use like this:

`new WikipediaTool()`

packages/tools/package.json
```diff
@@ -0,0 +1,59 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/run-llama/LlamaIndexTS.git",
+    "directory": "packages/tools"
+  },
+  "scripts": {
+    "build": "rm -rf ./dist && pnpm run build:esm && pnpm run build:cjs && pnpm run build:type && pnpm run build:wasm",
+    "build:esm": "swc src -d dist --strip-leading-paths --config-file .swcrc",
+    "build:cjs": "swc src -d dist/cjs --strip-leading-paths --config-file .cjs.swcrc",
```
EmanuelCampos, line 55: #620

packages/tools/assembly/base.ts
```diff
@@ -0,0 +1,76 @@
...
+  constructor(type: string, description: string | null = null) {
+    this.type = type;
+    this.description = description;
+  }
+}
+
+// Because AssemblyScript does not support Record<string, ToolParameterProperty> yet,
+// we have to use an array of key-value pairs instead.
+// When loading the metadata in application, we will convert
+// the array ToolParameterPropertyRecord[] to Record<string, ToolParameterProperty>.
```
marcusschiesser, line 14: AssemblyScript is supporting `Map`s (see https://www.assemblyscript.org/stdlib/map.html) - how about using them?

packages/tools/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+  type: string | "object";
+  properties: Record<string, { type: string; description?: string }>;
+  required?: string[];
+};
+
+interface ToolMetadata {
+  description: string;
+  name: string;
+  parameters?: ToolParameters;
+  argsKwargs?: Record<string, any>;
```
marcusschiesser, line 15: `argsKwargs` is currently not used by the agent - I would remove this complexity from the tools package for the start

packages/tools/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Convert asm string to ts string
+   * Convert asm array to ts array
+   * Convert the properties from an Array to a Record<string, { type: string; description?: string }>
+   * Convert the argsKwargs from an Array to a Record<string, any>
+   */
+  private static wasmInstanceToConfigs(wasmInstance: any): BaseTool {
+    const { __pin, __unpin, __getString, __getArray, __newString } =
+      wasmInstance.exports;
+
+    const getObjectByAddress = (address: any, classWrapper: any) => {
```
marcusschiesser, line 35: can we extract her reusable transform operations (e.g. Map to Record)? I guess this would make the code easier to read

packages/tools/assembly/base.ts
marcusschiesser: ok, then better document this in the code.

marcusschiesser (approved):
Looks great!  Thanks @thucpn

PR #630, opened 2024-03-11
feat: Add more. env variables to config host, port, llm and embedding

leehuwuj:


packages/create-llama/helpers/index.ts
```diff
@@ -7,6 +7,7 @@ import { cyan } from "picocolors";
 
 import { COMMUNITY_OWNER, COMMUNITY_REPO } from "./constant";
 import { templatesDir } from "./dir";
+import { createEnvLocalFile } from "./env-variables";
```
marcusschiesser, line 10: good to factor out!

packages/create-llama/helpers/env-variables.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+  console.log("\nUsing OpenAI model: ", model, "\n");
+
+  if (opts?.openAiKey) {
+    content += `OPENAI_API_KEY=${opts?.openAiKey}\n`;
+  }
+
+  if (opts?.framework === "fastapi") {
+    //  Default environment variables for FastAPI
+    const defaultEnvs = [
```
marcusschiesser, line 54: great idea with this defaultEnvs - how about we put all variables in there first and then render the string?
from what I see, only conditional rendering is missing, we could do this in two ways:

1. add a `render` flag to each entry, e.g.:
```js
[
{
        render: opts?.framework === "nextjs",
        name: "NEXT_PUBLIC_MODEL",
        value: opts?.model ?? "gpt-3.5-turbo",
}
...
 ```
 or decide to add the entry in the array literal:
 ```js
[
   ...opts?.framework === "nextjs" ? [{
        name: "NEXT_PUBLIC_MODEL",
        value: opts?.model ?? "gpt-3.5-turbo",
    }] : []
...
 ```
 
I guess we could then separate building the structure and rendering it. Also we could make two defaultEnvs: one for python and one for TS (and therefore separate each environment then nicely)

File: packages/create-llama/templates/components/engines/python/agent/__init__.py [hunk 3]
```diff
@@ -6,11 +7,13 @@
 
 
 def get_chat_engine():
+    system_prompt = os.getenv("SYSTEM_PROMPT")
+    top_k = os.getenv("TOP_K")
```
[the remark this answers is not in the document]
leehuwuj: - `system_prompt` default is None so it's no affect.
- `top_k`: I could not identify the default value (could be 2). If its value is None then the default behavior may change. So we should set a default value for it in the code, i just updated it with top_k = 3 as default.

packages/create-llama/helpers/env-variables.ts
leehuwuj: Thanks for the suggestion, it's great to apply to all our default env variables.

PR #636, opened 2024-03-12
fix: add missing check env logic

leehuwuj:


packages/create-llama/helpers/env-variables.ts
```diff
@@ -20,9 +20,11 @@ const renderEnvVar = (envVars: EnvVar[]): string => {
       (env.description
         ? `# ${env.description.replaceAll("\n", "\n# ")}\n`
         : "") +
-      (env.name && env.value
-        ? `${env.name}=${env.value}\n\n`
-        : `# ${env.name}=\n\n`),
+      (env.name
```
leehuwuj, line 23: here, we allow empty input `{}` but won't render it. Just for easier generating the env variables.

marcusschiesser (approved):

PR #638, opened 2024-03-12
fix: add missing env value; improve docs and error message

leehuwuj:


packages/create-llama/templates/types/simple/fastapi/README-template.md
```diff
@@ -44,7 +46,7 @@ Open [http://localhost:8000/docs](http://localhost:8000/docs) with your browser
 The API allows CORS for all origins to simplify development. You can change this behavior by setting the `ENVIRONMENT` environment variable to `prod`:
 
 ```
-ENVIRONMENT=prod uvicorn main:app
+ENVIRONMENT=prod python main.py
```
leehuwuj, line 49: because `uvicorn` cli does not respect the configured value in main.py so the host and port config won't work if user changed in .env.

marcusschiesser (approved):

PR #645, opened 2024-03-14
feat: Add auto create collection and node metadata for Milvus vector store

leehuwuj:


packages/core/src/storage/vectorStore/MilvusVectorStore.ts
```diff
@@ -12,16 +13,18 @@ import type {
   VectorStoreQuery,
   VectorStoreQueryResult,
 } from "./types.js";
+import { nodeToMetadata } from "./utils.js";
```
marcusschiesser, line 16: If you use `nodeToMetadata`, you also have to use `metadataDictToNode` see e.g. astradb

packages/core/src/storage/vectorStore/MilvusVectorStore.ts
```diff
@@ -71,13 +76,54 @@ export class MilvusVectorStore implements VectorStore {
...
+      await this.createCollection();
+    }
+
     await this.milvusClient.loadCollectionSync({
-      collection_name: collection,
+      collection_name: this.collection,
     });
 
-    this.collection = collection;
+    this.collection = this.collection;
```
marcusschiesser, line 126: no-op?

packages/core/src/storage/vectorStore/MilvusVectorStore.ts
```diff
@@ -71,13 +76,54 @@ export class MilvusVectorStore implements VectorStore {
...
+        },
+      ],
+    });
+    await this.milvusClient.createIndex({
+      collection_name: this.collection,
+      field_name: this.embeddingKey,
+    });
+  }
+
+  public async connect(): Promise<void> {
```
marcusschiesser, line 111: It's better DX if the user doesn't have to call this `connect` before using the store.
Instead the store can do this check itself.
To not add unnecessary latency, this check should be only done once (lazily)
```suggestion
  private async ensureCollection(): Promise<void> {
```

packages/core/src/storage/vectorStore/MilvusVectorStore.ts
```diff
@@ -103,36 +104,35 @@ export class MilvusVectorStore implements VectorStore {
...
-    // Check collection exists
-    const isCollectionExist = await this.milvusClient.hasCollection({
-      collection_name: this.collection,
-    });
-    if (!isCollectionExist.value) {
-      await this.createCollection();
+      await this.milvusClient.loadCollectionSync({
+        collection_name: this.collectionName,
+      });
+      this.collection = this.collectionName;
```
marcusschiesser, line 127: it's a bit confusing that `this.collection` is a `string` - doesn't a `collectionInitialized` work?

Issue #639, opened 2024-03-12
SimpleNodeParser should produce TextNodes with deterministic IDs


g12i:
When initializing the `VectorStoreIndex` with the default service context, it utilizes the `SimpleNodeParser` that divides documents into a sequence of nodes. However, with each call, new nodes are generated with random UUIDs associated. In my view, it would be preferable for this process to be idempotent. Assigning node IDs derived from the SOURCE ID and text hash should be sufficient.

This way, if I call `index.insert` with the same origin Document it won't clutter my storage with ghost nodes.

```diff
// packages/core/src/nodeParsers/utils.ts

  textSplits.forEach((textSplit) => {
    const node = new TextNode({
+     id_: `${document.id_}/${quickHash(textSplit)}`,
      text: textSplit,
      metadata: includeMetadata ? _.cloneDeep(document.metadata) : {},
      excludedEmbedMetadataKeys: _.cloneDeep(
        document.excludedEmbedMetadataKeys,
      ),
      excludedLlmMetadataKeys: _.cloneDeep(document.excludedLlmMetadataKeys),
    });
    node.relationships[NodeRelationship.SOURCE] = document.asRelatedNodeInfo();
    nodes.push(node);
  });
  
+ function quickHash(str: string): string {
+   let hash = 0;
+   for (let i = 0, len = str.length; i < len; i++) {
+     const chr = str.charCodeAt(i);
+     hash = (hash << 5) - hash + chr;
+     hash |= 0; // Convert to 32bit integer
+   }
+   return String(hash);
+ }

```

marcusschiesser:
@g12i Thanks a lot for pointing this out. I think we can not generally add this, as there are different use cases to cover. For example, your `quickHash` value would be the same if the metadata is changing.

But after calling the `SimpleNodeParser`, the split `TextNodes` should have the same `hash` value (if metadata and content are the same).

Are you currently calling `VectorStoreIndex.insertNodes` to add new nodes? This method does not check the `hash` value. 

Have you tried using our [IngestionPipelines](https://ts.llamaindex.ai/modules/ingestion_pipeline/#connecting-to-vector-databases) - they are handling `hash` values.

Alternatively, we could also add an `idFunc` to the `SimpleNodeParser` that generates an ID, and you could pass in your code a customized ID function.

Please let me know what you think, and I will be happy to solve this problem.

g12i:
Hi @marcusschiesser , thanks for your response. 

[...]

PR #646, opened 2024-03-14
feat: Add doc store strategies to VectorStoreIndex.fromDocuments

marcusschiesser:
- Adding the same doc store strategies from IngestionPipeline to VectorStoreIndex.fromDocuments
- Prevents duplicate issues like https://github.com/run-llama/LlamaIndexTS/issues/639

### DIFF
--- a/packages/core/src/indices/vectorStore/index.ts
+++ b/packages/core/src/indices/vectorStore/index.ts
@@ -230,10 +230,12 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
       docStore,
       args.vectorStore,
     );
-    args.nodes = await runTransformations(documents, [
-      ...(docStoreStrategy ? [docStoreStrategy] : []),
-      args.serviceContext.nodeParser,
-    ]);
+    args.nodes = await runTransformations(
+      documents,
+      [args.serviceContext.nodeParser],
+      {},
+      { docStoreStrategy },
+    );
     if (args.logProgress) {
       console.log("Finished parsing documents.");
     }
--- a/packages/core/src/indices/vectorStore/index.ts
+++ b/packages/core/src/indices/vectorStore/index.ts
@@ -214,7 +214,12 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
       docStoreStrategy?: DocStoreStrategy;
     } = {},
   ): Promise<VectorStoreIndex> {
-    args.docStoreStrategy = args.docStoreStrategy ?? DocStoreStrategy.UPSERTS;
+    args.docStoreStrategy =
+      args.docStoreStrategy ??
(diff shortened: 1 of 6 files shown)


packages/core/src/indices/BaseIndex.ts
```diff
@@ -91,10 +95,24 @@ export abstract class BaseIndex<T> {
...
+          options.docStoreStrategy,
+          this.docStore,
+          this.vectorStore,
+        )
+      : undefined;
     const nodes = await runTransformations(
       [document],
-      [this.serviceContext.nodeParser],
+      [
+        ...(docStoreStrategy ? [docStoreStrategy] : []),
```
EmanuelCampos, line 113:
```suggestion
        ...(docStoreStrategy && [docStoreStrategy])
```

I think this way is less prone to errors

and why it's an array? not the transformer directly?

marcusschiesser: That's not doing the same thing. The idea is to prepend `docStoreStrategy` to the array literal only if it defined

EmanuelCampos: right, but the && operator will only pass if `docStoreStrategy ` is true, if not will not prepend, on the current way it prepend an empty array (which I think is unnecessary)

EmanuelCampos: oh that's right, the spread will let it empty

packages/core/src/indices/vectorStore/index.ts
```diff
@@ -213,21 +207,28 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
...
-    args: VectorIndexOptions = {},
+    args: VectorIndexOptions & {
+      docStoreStrategy?: DocStoreStrategy;
+    } = {},
   ): Promise<VectorStoreIndex> {
+    args.docStoreStrategy = args.docStoreStrategy ?? DocStoreStrategy.UPSERTS;
     args.storageContext =
       args.storageContext ?? (await storageContextFromDefaults({}));
     args.serviceContext = args.serviceContext ?? serviceContextFromDefaults({});
     const docStore = args.storageContext.docStore;
```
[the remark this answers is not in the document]
logan-markewich: But it will only continue to work if the user persists the docstore? I'm not sure if llamaindexTS has any remote docstore integrations, if not, thay will probably be the next thing people want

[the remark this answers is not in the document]
marcusschiesser: llamaindexTS does not have remote docstore integration yet. Yes for it to work the user would need to persist the docstore (same as for the ingestion pipeline)

yisding:
Note to self: want to review this.

marcusschiesser:
documented doc strategies first in Python: https://github.com/run-llama/llama_index/pull/12457

PR #648, opened 2024-03-15
feat: add wikipedia tool

thucpn:


EmanuelCampos (commented):
what about an example file with agent or any other use case?

marcusschiesser (commented):
@thucpn looks great. please add an example to `/examples/agent` as @EmanuelCampos suggested

marcusschiesser (approved):

PR #650, opened 2024-03-15
[docs] missing await in sample

pelikhan:


EmanuelCampos (approved):
thanks for the contribution!

PR #654, opened 2024-03-18
[WIP]: PromptTemplate

EmanuelCampos:
DRAFT, not ready for use

missing:
- refactor missing modules
- refactor existent prompts
- tests with refine template
- fix circular dependency
- refactor partial format + prompt repack


packages/core/src/test.ts
EmanuelCampos, line 1: this file will be removed, just an example to test fast for now

packages/core/src/test.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  ChatPromptTemplate,
+  PromptTemplate,
+  SimpleDirectoryReader,
+  VectorStoreIndex,
+  type ChatMessage,
+} from "./index.js";
+
+// Define a default prompt template
+const defaultTextQaPromptTemplate = () => `Context information is below.
```
marcusschiesser, line 10: does this need to be function?
```suggestion
const defaultTextQaPromptTemplate = "Context information is below.
```

packages/core/src/test.ts
```diff
@@ -0,0 +1,71 @@
...
+  const retriever = await index.asRetriever({});
+  retriever.similarityTopK = 10;
+
+  const queryEngine = index.asQueryEngine({
+    retriever,
+  });
+
+  // Update the prompts
+  queryEngine.updatePrompts({
+    "responseSynthesizer:textQATemplate": chatMessageTemplate,
```
marcusschiesser, line 56: can we directly pass the template?
```suggestion
    "responseSynthesizer:textQATemplate": [
  {
    content: "Always answer the question, even if you don't know the answer.",
    role: "system",
  },
  {
    content: defaultTextQaPromptTemplate(),
    role: "user",
  },
],
```
e.g. `updatePrompts` could generate a `ChatPromptTemplate` object if needed

marcusschiesser, line 56: How about adding a convenience function for the user that is updating the most likely prompt (here `responseSynthesizer:textQATemplate`):
```suggestion
  queryEngine.updatePrompt(chatMessageTemplate)
```

EmanuelCampos:
closing, due to several changes on how it works, moving to https://github.com/run-llama/LlamaIndexTS/pull/665

@marcusschiesser thanks for the comments, appreciate your review on the new PR

PR #656, opened 2024-03-19
feat: add tidb vector store

Mini256:


packages/core/src/storage/vectorStore/TiDBVectorStore.ts
```diff
@@ -0,0 +1,322 @@
...
+
+        // All good?  Keep the connection reference
+        this.db = db;
+      } catch (err: any) {
+        console.error(err);
+        return Promise.reject(err);
+      }
+    }
+
+    return Promise.resolve(this.db);
```
marcusschiesser, line 108:
```suggestion
        throw err;
      }
    }

    return this.db;
```

packages/core/src/storage/vectorStore/TiDBVectorStore.ts
```diff
@@ -0,0 +1,322 @@
...
+    namespace?: string;
+    tableName?: string;
+    dimensions?: number;
+    poolOptions?: PoolOptions;
+    client?: mysql.Pool;
+  }) {
+    this.tableName = config?.tableName ?? TIDB_VECTOR_TABLE;
+    this.namespace = config?.namespace ?? "";
+    this.dimensions = config?.dimensions ?? 1536;
+    this.poolOptions = config?.poolOptions ?? {};
```
marcusschiesser, line 55: you can use `Object.assign` to copy the config to this (then you don't need to duplicate your default values)

packages/core/src/storage/vectorStore/TiDBVectorStore.ts
```diff
@@ -0,0 +1,322 @@
...
+   */
+  async delete(refDocId: string, deleteKwargs?: any): Promise<void> {
+    const namespaceCriteria = this.namespace.length ? "AND namespace = $2" : "";
+    const sql: string = `DELETE FROM ${this.tableName} WHERE id = $1 ${namespaceCriteria}`;
+    const db = await this.getDb();
+    const params = this.namespace.length
+      ? [refDocId, this.namespace]
+      : [refDocId];
+    await db.query(sql, params);
+    return Promise.resolve();
```
marcusschiesser, line 227: no-op

packages/core/src/storage/vectorStore/TiDBVectorStore.ts
```diff
@@ -0,0 +1,322 @@
...
+  document: string;
+  metadata: Metadata;
+  embeddings: number[];
+  score: number;
+}
+
+/**
+ * Provides support for writing and querying vector data in TiDB.
+ */
+export class TiDBVectorStore implements VectorStore {
```
marcusschiesser, line 27: did you check if the data generated is compatible with the python version (see https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-tidbvector/llama_index/vector_stores/tidbvector/base.py) ?

packages/core/src/storage/vectorStore/TiDBVectorStore.ts
```diff
@@ -0,0 +1,322 @@
...
+    return Promise.resolve();
+  }
+
+  /**
+   * Query the vector store for the closest matching data to the query embeddings
+   * @param query The VectorStoreQuery to be used
+   * @param options Required by VectorStore interface.  Currently ignored.
+   * @returns Zero or more Document instances with data from the vector store.
+   */
+  async query(
```
marcusschiesser, line 236: We're using metadataDictToNode and nodeToMetadata from "./utils.js" to de/serialize the node in the metadata. You can reference the other vector store implementations to have a look into this.

marcusschiesser (commented):
@Mini256 looking great! Thanks for your contribution. Please have a look at my comments and if you have time it would be great if you could add an example to the `/examples` folder

packages/core/src/storage/vectorStore/TiDBVectorStore.ts
Mini256: grest, I will make some adjustments to be consistent with the Python version.

himself65 (commented):
any update on this?

himself65:
Closing as stale

Issue #660, opened 2024-03-21
Claude 3 ReAct Support


oberlinstands:
I see in the docs Claude 3 Opus is not officially supported in the ReAct agent. Curious to see if this is on the roadmap, if someone is working on it, or what I could do to get started by opening up a PR?

marcusschiesser:
Thanks, @oberlinstands. That's a great idea and we're always happy for any PR!
Using the ReAct Agent should be possible with any LLM, including Claude 3. 

I just created a PR with an example https://github.com/run-llama/LlamaIndexTS/pull/661 - you might want to start from that.

Running it shows up the following error:
```
Thought: I need to use the divideNumbers tool to divide 16 by 2, then use the sumNumbers tool to add 20 to the result.
Action: divideNumbers
Action Input: {"a":16,"b":2}

Observation: 8
/Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@anthropic-ai+sdk@0.18.0/node_modules/@anthropic-ai/sdk/src/error.ts:58
      return new BadRequestError(status, error, message, headers);
             ^
BadRequestError: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: roles must alternate between \"user\" and \"assistant\", but found multiple \"user\" roles in a row"}}
    at Function.generate (/Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@anthropic-ai+sdk@0.18.0/node_modules/@anthropic-ai/sdk/src/error.ts:58:14)
    at Anthropic.makeStatusError (/Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@anthropic-ai+sdk@0.18.0/node_modules/@anthropic-ai/sdk/src/core.ts:383:21)
    at Anthropic.makeRequest (/Users/marcus/code/llamaindex/LlamaIndexTS/node_modules/.pnpm/@anthropic-ai+sdk@0.18.0/node_modules/@anthropic-ai/sdk/src/core.ts:446:24)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Anthropic.chat (/Users/marcus/code/llamaindex/LlamaIndexTS/packages/core/dist/cjs/llm/LLM.js:596:26)
    at async ReActAgentWorker._runStep (/Users/marcus/code/llamaindex/LlamaIndexTS/packages/core/dist/cjs/agent/react/worker.js:227:30)
[...]

PR #661, opened 2024-03-22
feat: use claude3 with react agent

marcusschiesser:
Updated react example to use Claude3.

To run:
```shell
cd examples
pnpm ts-node agent/react_agent.ts
```


EmanuelCampos:
Solves #660

PR #663, opened 2024-03-22
feat: add tool factory

thucpn:


packages/core/src/tools/ToolFactory.ts
```diff
@@ -0,0 +1,33 @@
+import type { BaseTool } from "../types.js";
+import { WikipediaTool } from "./WikipediaTool.js";
+
+enum ExternalTool {
```
marcusschiesser, line 4: How about just 'Tools'

marcusschiesser (approved):

PR #668, opened 2024-03-26
feat: add global settings

EmanuelCampos:
Closes: https://github.com/run-llama/LlamaIndexTS/issues/671

- now settings are configurable through
```ts
import { Settings } from "llamaindex"

Settings.llm = new OpenAI({ model: "gpt-4" });
```
- settings are stored on nodejs global and on window. for browser
- serviceContextFromDefaults is being deprecated
- we still accept serviceContext for backwards compatibility, but not using more as defult


packages/core/src/Retriever.ts
```diff
@@ -13,5 +13,5 @@ export type RetrieveParams = {
  */
 export interface BaseRetriever {
   retrieve(params: RetrieveParams): Promise<NodeWithScore[]>;
-  getServiceContext(): ServiceContext;
+  getServiceContext(): ServiceContext | undefined;
```
marcusschiesser, line 16: how about removing this? my guess is this is not used very often

packages/core/src/engines/query/RouterQueryEngine.ts
```diff
@@ -55,7 +55,7 @@ async function combineResponses(
  * A query engine that uses multiple query engines and selects the best one.
  */
 export class RouterQueryEngine extends PromptMixin implements BaseQueryEngine {
-  serviceContext: ServiceContext;
+  serviceContext?: ServiceContext;
```
marcusschiesser, line 58: how about removing the serviceContext from classes (but keeping it from the init parameter)?

my guess is that users are never reading the serviceContext and if they are using it, they are passing it as init parameter (that's what all our example code does)

packages/core/src/evaluation/Correctness.ts
```diff
@@ -24,17 +24,19 @@ type CorrectnessParams = {
 
 /** Correctness Evaluator */
 export class CorrectnessEvaluator extends PromptMixin implements BaseEvaluator {
-  private serviceContext: ServiceContext;
+  private serviceContext?: ServiceContext;
```
marcusschiesser, line 27: not needed (as state is stored in llm)

packages/core/src/Settings.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  set prompt(prompt: PromptConfig) {
+    this._prompt = prompt;
+  }
+}
+
+// Determine the global object based on the environment
+const globalObject: any =
+  typeof window !== "undefined"
+    ? window
```
himself65, line 130: AND, use `globalThis,` do not do this since there is no forward compatibility considerations
AND, don't use `any` if possible

```ts
declare global {
  // var is necessary, not const or let
  var Settings: xxx;
}

globalThis.Settings = xxx;
```

himself65: BTW, I want to use AsyncLocalSotrage here, which will have more features for async context.

https://nodejs.org/docs/latest/api/async_context.html#class-asynclocalstorage

P.S. Even though this is node.js API, but deno, bun, and vercel edge runtime all support this

examples/qdrantdb/preFilters.ts
```diff
@@ -4,12 +4,21 @@ import {
...
+// Update callback manager
+Settings.callbackManager = new CallbackManager({
+  onRetrieve: (data) => {
+    console.log(
+      "The retrieved nodes are:",
+      data.nodes.map((node) => node.node.getContent(MetadataMode.NONE)),
+    );
+  },
+});
+
```
marcusschiesser, line 21: @himself65 just saw that you removed `runWithCallbackManager` here ;)
How about his simplification:
```suggestion
Settings.callbacks = {
  onRetrieve: (data) => {
    console.log(
      "The retrieved nodes are:",
      data.nodes.map((node) => node.node.getContent(MetadataMode.NONE)),
    );
  },
};

```

examples/recipes/cost-analysis.ts
```diff
@@ -7,8 +7,7 @@ const llm = new OpenAI({
 
 let tokenCount = 0;
 
-// @todo: use GlobalSetting in the future
-getCurrentCallbackManager().addHandlers("llm-start", (event) => {
+Settings.callbackManager.addHandlers("llm-start", (event) => {
```
marcusschiesser, line 10: similar here:
```suggestion
Settings.callbacks.add("llm-start", (event) => {
```

marcusschiesser (commented):
@himself65 @EmanuelCampos This looks really cool

himself65 (approved):
LGTM

himself65 (approved):
LGTM

Issue #671, opened 2024-03-28
Async Local Storage for global Settings


EmanuelCampos:
https://nodejs.org/docs/latest/api/async_context.html#class-asynclocalstorage

P.S. Even though this is node.js API, but deno, bun, and vercel edge runtime all support this

_Originally posted by @himself65 in https://github.com/run-llama/LlamaIndexTS/pull/668#discussion_r1542166013_

himself65:
My expected API would be like

```ts
import { Settings } from 'llamaindex'

Settings.runWithLLM(async () => {
  // will use this llm instance
}, llm)
```

himself65:
Some upstream issue: https://github.com/denoland/deno/pull/23175

PR #675, opened 2024-03-29
feat: improve callback manager

himself65:
As we talked about before

Also, use AsyncLocalStorage for better API and code style.

Related: https://github.com/run-llama/LlamaIndexTS/issues/671, https://github.com/run-llama/LlamaIndexTS/pull/668#discussion_r1542166013

In the future, we should put more things into global settings and async local storage instead of using `new XXX({ ... })` (everything into parameter when creation), which worse understand and debug

Downstream: https://github.com/run-llama/LlamaIndexTS/pull/667

---

There should only be 2 ways to change such variables in the future (like callback manager, or llm in a service context)

1. AsyncLocalStorage

```ts
import { runWith } from 'llamaindex'
await runWith(myllm, () => {})
```

2. Global Settings

```ts
import { Settings } from 'llamaindex'

Settings.llm = myllm
```

---

More context about AsyncLocalStorage

- TC39 Proposal: https://github.com/tc39/proposal-async-context
- https://nodejs.org/api/async_context.html


packages/core/src/callbacks/CallbackManager.ts
```diff
@@ -60,25 +67,94 @@ export interface RetrievalCallbackResponse extends BaseCallbackResponse {
...
+    handler: EventHandler<LlamaIndexEventMaps[K]>,
+  ) {
+    if (!this.#handlers.has(event)) {
+      this.#handlers.set(event, []);
+    }
+    this.#handlers.get(event)!.push(
+      // @ts-expect-error
+      handler,
+    );
+  }
```
himself65, line 121: WIP

examples/multimodal/rag.ts
```diff
@@ -38,17 +39,18 @@ async function main() {
...
   const index = await createIndex(serviceContext);
 
   const queryEngine = index.asQueryEngine({
     responseSynthesizer: new MultiModalResponseSynthesizer({ serviceContext }),
     retriever: index.asRetriever({ similarityTopK: 3, imageSimilarityTopK: 1 }),
   });
-  const result = await queryEngine.query({
-    query: "Tell me more about Vincent van Gogh's famous paintings",
-  });
+  const result = await runWithCallbackManager(callbackManager, () =>
```
yisding, line 49: Is this runWithCallbackManager the only way to get this functionality?

I think for the most natural DevEx the user should do something like registerCallbackManager one, and then everything should just automatically flow to that.

yisding: Almost like a `window.addEventListener` (actually, that makes me wonder if that would be an acceptable interface to hook into.

himself65: As mentioned in the description, you could do `Settings.callbackManager = new CallbackManager` if you want

himself65: `runWith` is more like fine-grained tracking situation.

For example for tracking LLM cost for a single call, you call use `runWithCallbackManager(myManager, () => llm.query())`, then you could get usage for this single call.

But you want to do some global injection, I think we should suggest user to use Settings API

yisding: OK let's make that the default in our examples. I'm guessing most beginner users will only need one CallbackManager for one specific thing (i.e. logging or Weights and Biases). Btw, should we make that an array? Or is there an easy way to chain CallbackManagers?

himself65: I'm making it chainable now; see https://github.com/run-llama/LlamaIndexTS/pull/675#discussion_r1544860405

I'm imaging that you can do like 

```
cb
  .addHandler('chat-start', () => {})
  .addHandler('chat-start', () => {})
  .addHandler('chat-end', () => {})
```

examples/qdrantdb/preFilters.ts
```diff
@@ -36,19 +37,21 @@ async function main() {
...
-        callbackManager: new CallbackManager({
-          onRetrieve: (data) => {
-            console.log(
-              "The retrieved nodes are:",
-              data.nodes.map((node) => node.node.getContent(MetadataMode.NONE)),
-            );
-          },
-        }),
+    const index = await runWithCallbackManager(
+      new CallbackManager({
```
marcusschiesser, line 41: @himself65 as we're changing this anyhow, how about simplifying it to:
```suggestion
const index = await runWithCallbacks({
```

Issue #664, opened 2024-03-22
Calculate the cost or tokens for each question


jung-han:
Hi llamaindex ts. Thank you for making a great tool. 
As a developer who primarily uses TS, this is of great help.

I have one question. It's about cost.
Using `ContextChatEngine` and `serviceContextFromDefaults`, I created a bot that asks information questions in RAG style. I want to know how many tokens and costs each input and output cost when making this query.
In the case of langchain, it seems that you can find out the prompt and output token through callback. 
I think there is a TokenCountingHandler in llamaindex.. (https://docs.llamaindex.ai/en/stable/examples/callbacks/TokenCountingHandler/)

But I wonder if there is a similar method in llamaIndex ts. If not, how should I calculate it?

marcusschiesser:
Thanks for your suggestion, we have a PR open working on it: https://github.com/run-llama/LlamaIndexTS/pull/270 - will be available soon. @himself65 as you've been working on it last time, should I continue?

himself65:
We have implemented global settings (thanks, @EmanuelCampos ) and improved the callback manager; you can add this small change to your code and track the token used, I'm not adding `Usage` class because you can do it now by adding only few code and with your own logic.

```ts
import { encodingForModel } from "js-tiktoken";
import { Settings } from "llamaindex";

const encoding = encodingForModel("gpt-4-0125-preview");

let tokenCount = 0;

Settings.callbackManager.on("llm-start", (event) => {
  const { messages } = event.detail.payload;
  tokenCount += messages.reduce((count, message) => {
    return count + encoding.encode(message.content).length;
  }, 0);
  console.log("Token count:", tokenCount);
  // https://openai.com/pricing
  // $10.00 / 1M tokens
  console.log(`Price: $${(tokenCount / 1_000_000) * 10}`);
});
Settings.callbackManager.on("llm-end", (event) => {
  const { response } = event.detail.payload;
  tokenCount += encoding.encode(response.message.content).length;
  console.log("Token count:", tokenCount);
  // https://openai.com/pricing
  // $30.00 / 1M tokens
  console.log(`Price: $${(tokenCount / 1_000_000) * 30}`);
});
```

Document: https://ts.llamaindex.ai/recipes/cost-analysis

himself65:
https://github.com/run-llama/LlamaIndexTS/pull/707

PR #679, opened 2024-04-01
refactor: remove `llm.tokens` api

himself65:
Closes: https://github.com/run-llama/LlamaIndexTS/pull/667, https://github.com/run-llama/LlamaIndexTS/issues/664

I'm removing this API since most of the LLM classes don't have an accurate function (only OpenAI supports it), and there are runtime errors.

But I put code into `examples/recipes/cost-analysis.ts` to let users know how to implement it by theirselves.

### DIFF
--- a/packages/core/src/ChatHistory.ts
+++ b/packages/core/src/ChatHistory.ts
@@ -1,7 +1,8 @@
-import { OpenAI } from "./llm/LLM.js";
-import type { ChatMessage, LLM, MessageType } from "./llm/types.js";
+import { globalsHelper } from "./GlobalsHelper.js";
 import type { SummaryPrompt } from "./Prompt.js";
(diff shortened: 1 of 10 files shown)


packages/core/src/ChatHistory.ts
```diff
@@ -178,7 +187,10 @@ export class SummaryChatHistory extends ChatHistory {
     const requestMessages = this.calcCurrentRequestMessages(transientMessages);
 
     // get tokens of current request messages and the transient messages
-    const tokens = this.llm.tokens(requestMessages);
+    const tokens = requestMessages.reduce(
```
marcusschiesser, line 190: @himself65 this is just calculating the tokens of the `message.content` but not the used tokens for a message prompt like in the old code

PR #685, opened 2024-04-02
Fixed postgres for "type"="module", ie es - should now work for both this and commonjs

WarlaxZ:
Spotted an issue that meant it wouldn't work and fixed it up.

### DIFF
--- a/packages/core/src/storage/vectorStore/PGVectorStore.ts
+++ b/packages/core/src/storage/vectorStore/PGVectorStore.ts
@@ -82,7 +82,7 @@ export class PGVectorStore implements VectorStore {
   private async getDb(): Promise<pg.Client> {
     if (!this.db) {
       try {
-        const pg = await import('pg');
+        const pg = await import("pg");
         const { Client } = pg.default ? pg.default : pg;
 
         const { registerType } = await import("pgvector/pg");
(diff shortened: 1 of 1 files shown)


packages/core/src/storage/vectorStore/PGVectorStore.ts
```diff
@@ -82,7 +82,9 @@ export class PGVectorStore implements VectorStore {
   private async getDb(): Promise<pg.Client> {
     if (!this.db) {
       try {
-        const { Client } = await import("pg");
+        const pg = await import('pg');
+        const { Client } = pg.default ? pg.default : pg;
```
bennyschmidt, line 86: I can confirm this when trying locally, it's due to the use of `await import` so it can be supported in both ESM/non-ESM 

```suggestion
        const { Client } = pg.default || pg;
```

bennyschmidt (approved):

packages/core/src/storage/vectorStore/PGVectorStore.ts
bennyschmidt: Just needs that prettier step to merge

himself65 (approved):
Thanks for helping out

PR #689, opened 2024-04-03
fix: use LLM's context window to specify agent's token limit

marcusschiesser:
I found this issue while testing the `WikipediaTool,` which quickly fills up the agent's memory (see `examples/agent/wiki.ts`).
The default token limit of `ChatMemoryBuffer` is too small for LLMs with larger context windows. 
So it's a good idea to use one that's specific to the selected LLM.


EmanuelCampos (approved):
sweet, just push a changeset

PR #694, opened 2024-04-05
refactor: improve agent type

himself65:


packages/core/src/agent/react/worker.ts
```diff
@@ -75,15 +77,18 @@ export class ReActAgentWorker implements AgentWorker {
     maxInteractions,
     reactChatFormatter,
     outputParser,
-    verbose,
     toolRetriever,
   }: ReActAgentWorkerParams) {
-    this.llm = llm ?? new OpenAI({ model: "gpt-3.5-turbo-0613" });
+    this.llm =
+      llm ?? isFunctionCallingModel(Settings.llm.metadata.model)
```
marcusschiesser, line 83: react pattern doesn't need function call model

PR #699, opened 2024-04-09
fix: remove optional type on BaseTool

himself65:


himself65:
See https://github.com/run-llama/LlamaIndexTS/pull/709

Issue #692, opened 2024-04-05
unifed logic for function call / agent


himself65:
Since gpt3.5/4 and Claude-3 are all support function calls, they have different way to use it. It's necessary to use a better way to handle it on the LlamaindexTS side to make user easy to switch without overhead.

```ts
type MessageRole = 'user' | 'ai' | 'system'
type MessageContentPart = MessageContentTextPart | MessageContentImagePart // ...
type Message = {
  role: MessageRole
  content: string | MessageContentPart[]
}

interface MessageInterop<LLMMessage> {
  toLLMMessage(messages: Message[]): LLMMessage[]
  fromAIMessage(message: LLMMessage[]): Message[]
}
interface LLM<LLMMessage> extends MessageInterop<LLMMessage> {
  // ...
}

type Tool<Input> = {
  handler: (input: Input) => string | Promise<string>
  metadata: {
    name: string
    // ... there should be json schema so that it could support zod
  }
}

type LLMFunctionCallEvent = LLMBaseEvent<'llm-function-call', {
  tool: Tool; // <-- should be BaseTool<Input>
  input: unknown // <-- should be Input
}>
```

PR #701, opened 2024-04-10
refactor: chat message type

himself65:
Step 2 of improving Agent

### DIFF
--- a/packages/core/src/agent/openai/worker.ts
+++ b/packages/core/src/agent/openai/worker.ts
@@ -58,9 +58,9 @@ async function callFunction(
 
   return [
     {
-      content: String(output),
+      content: `${output}`,
       role: "tool",
-      additionalKwargs: {
+      options: {
         name,
         tool_call_id: id_,
       },
@@ -138,7 +138,8 @@ export class OpenAIAgentWorker
       return null;
     }
 
-    return chatHistory[chatHistory.length - 1].additionalKwargs?.toolCalls;
+    // fixme
+    return chatHistory[chatHistory.length - 1].options?.toolCalls as any;
   }
 
   private _getLlmChatParams(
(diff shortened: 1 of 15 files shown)


packages/core/src/agent/openai/worker.ts
```diff
@@ -58,9 +58,9 @@ async function callFunction(
 
   return [
     {
-      content: String(output),
+      content: `${output}`,
```
himself65, line 61: nit: output should be string, instead of `any`

packages/core/src/agent/openai/worker.ts
```diff
@@ -138,7 +138,8 @@ export class OpenAIAgentWorker
       return null;
     }
 
-    return chatHistory[chatHistory.length - 1].additionalKwargs?.toolCalls;
+    // fixme
+    return chatHistory[chatHistory.length - 1].options?.toolCalls as any;
```
himself65, line 142: nit: missing type in `chatHistory`

packages/core/src/agent/openai/worker.ts
```diff
@@ -58,9 +58,9 @@ async function callFunction(
 
   return [
     {
-      content: String(output),
+      content: `${output}`,
       role: "tool",
-      additionalKwargs: {
```
himself65, line 63: rename `kwargs` to `options`. It looks like the way of python. And I don't think there should be compatibility on this, instead should use some JS standard

packages/core/src/llm/LLM.ts
```diff
@@ -273,6 +273,7 @@ If a question does not make any sense, or is not factually coherent, explain why
       replicateOptions,
     );
     return {
+      raw: response,
```
himself65, line 276: leave raw for every response to make it easy to debug

PR #702, opened 2024-04-10
feat: use query bundle

himself65:
Upstream: https://github.com/run-llama/LlamaIndexTS/pull/704

Align some modules


packages/core/e2e/node/snapshot/llm.snap
```diff
@@ -0,0 +1,67 @@
...
+          "content": "Hello! How can I assist you today?",
+          "role": "assistant",
+          "options": {}
+        }
+      }
+    },
+    {
+      "id": "de46b84e-7345-430f-b8fa-423354b630c9",
+      "response": {
+        "raw": {},
```
himself65, line 59: We found a bug here, `raw` should be something actually

marcusschiesser (approved):
@himself65 looking great!

PR #704, opened 2024-04-10
test: coverage on mock mode

himself65:
Current e2e logic has two modes: mock and real

In the mock mode, we will try to combine the callback manager with node ESM hook, compare each input, and give the output. To make sure there is the least regression on codespace.

In the real mode, you can work with real AI


packages/core/e2e/node/snapshot/llm.snap
```diff
@@ -0,0 +1,67 @@
...
+          "content": "Hello! How can I assist you today?",
+          "role": "assistant",
+          "options": {}
+        }
+      }
+    },
+    {
+      "id": "de46b84e-7345-430f-b8fa-423354b630c9",
+      "response": {
+        "raw": {},
```
himself65, line 59: this is a bug

PR #706, opened 2024-04-10
fix: agent streaming with new OpenAI models

marcusschiesser:
Fixes streaming with OpenAIAgent using the latest OpenAI models. 

Fixes:
1. prepend content from the first chunk that was already read
2. skip parts that don't have any content
3. support multiple tool calls in one LLM request


packages/core/src/agent/openai/worker.ts
```diff
@@ -214,7 +214,19 @@ export class OpenAIAgentWorker
...
 
-    return new StreamingAgentChatResponse(newStream, task.extraState.sources);
+    // prepend content from first chunk that was already read
+    const prependedStream = (async function* () {
+      yield new Response(content);
+
+      for await (const response of newStream) {
+        yield response;
+      }
+    })();
```
marcusschiesser, line 224: @himself65 that's pretty ugly - anything more elegant?

himself65: updated

himself65: https://github.com/run-llama/LlamaIndexTS/pull/706/commits/0bdc6d838a654fa6e038f336a9ee9a0cc75125fa

PR #707, opened 2024-04-10
feat: add `llm-stream` event

himself65:
Related: https://github.com/run-llama/LlamaIndexTS/issues/664


packages/core/src/llm/utils.ts
```diff
@@ -105,6 +106,13 @@ export function wrapLLMEvent(
               ...chunk.options,
             };
           }
+          getCallbackManager().dispatchEvent("llm-stream", {
+            payload: {
+              id,
+              chunk,
+            },
+          });
+          finalResponse.raw.push(chunk);
```
himself65, line 115: call before the `yield` could be more sense

PR #708, opened 2024-04-10
Fix protobuf

yisding:
https://github.com/run-llama/LlamaIndexTS/security/dependabot/24


package.json
```diff
@@ -27,13 +27,14 @@
...
-    "typescript": "^5.4.4"
+    "typescript": "^5.4.5"
   },
   "packageManager": "pnpm@8.15.6+sha256.01c01eeb990e379b31ef19c03e9d06a14afa5250b82e81303f88721c99ff2e6f",
   "pnpm": {
     "overrides": {
       "trim": "1.0.1",
-      "@babel/traverse": "7.23.2"
+      "@babel/traverse": "7.23.2",
+      "protobufjs": "7.2.6"
```
yisding, line 37: @himself65 quick question: so while putting this here will update our pnpm-lock.yaml but I don't think it'll do anything for the consumers of our library correct? Would putting a npm overrides section in the package.json for packages/core help in that regard (for every package manager) or does every consumer of the library need to do this themselves? https://docs.npmjs.com/cli/v10/configuring-npm/package-json#overrides Or could we just add the new protobufjs version as a dependency?

himself65: overrides is for application use cases for example you are building a web app but find something bug in modules. There's no way to modify user's node_modules but npm:version

himself65: there's a solution but very hacky: https://github.com/dai-shi/waku/pull/136/files#diff-792d288469e94f9b513d387c9f83324d9cd9fc8a492534b8da4626795627e746R62

yisding: OK I'm going to merge this to get rid of the Github complaint, and hopefully transformers.js updates their dependencies and makes this not necessary. But if they don't in say, a month, let's come back and figure out what to do (write a PR/fork it/do the hack from your link)

PR #709, opened 2024-04-10
feat: improve base tool type

himself65:


packages/core/src/types.ts
```diff
@@ -30,14 +31,46 @@ export interface BaseQueryEngine {
...
-  call?: (...args: any[]) => any;
-  metadata: ToolMetadata;
+export interface BaseTool<Input = any> {
+  /**
+   * This could be undefined if the implementation is not provided,
+   *  which might be the case when communicating with a llm.
+   *
+   * @return string - the output of the tool, should be string in any case for LLM input.
+   */
+  call?: (input: Input) => string | Promise<string>;
```
himself65, line 65: talked with @logan-markewich, baseTool.call should return a string

himself65: also the input should be Input, which follow JSON schema

packages/core/tests/tools/Tools.test.ts
```diff
@@ -1,46 +0,0 @@
...
-    b: {
-      type: "number",
-      description: "The second number",
-    },
-  },
-  required: ["a", "b"],
-};
-
-describe("Tools", () => {
-  it("should be able to call a tool with a common JSON", async () => {
```
himself65, line 25: outdated, deleting

packages/core/tests/agent/OpenAIAgent.test.ts
```diff
@@ -1,61 +0,0 @@
-import { Settings } from "llamaindex";
```
himself65, line 1: removing mock logics in vitest

PR #710, opened 2024-04-11
fix: agent stream

himself65:


packages/core/src/agent/openai/worker.ts
```diff
@@ -212,14 +215,16 @@ export class OpenAIAgentWorker
...
       let content = "";
-      return pipeline(
-        finalStream.pipeThrough<Response>({
+      return new StreamingAgentChatResponse(
+        responseStream.pipeThrough<Response>({
           readable: new ReadableStream({
             async start(controller) {
-              for await (const chunk of finalStream) {
+              for await (const chunk of chunkStream) {
                 controller.enqueue(new Response(chunk.delta));
```
himself65, line 225: here is a pitfall that you cannot read and pipethough in the same time

PR #712, opened 2024-04-11
test: add openai agent stream chat

himself65:


himself65:
I think agent stream takes longer time than I expected

PR #716, opened 2024-04-14
chore: put eslint into top level

himself65:
First is to remove module package, but put lint rule into top-level, Second there's no `create-llama` here, so I removed some next eslint rules


marcusschiesser:
Nice cleanup, we can also turn on more of these @typescript-eslint rules if needed

PR #719, opened 2024-04-15
feat: init anthropic agent (part 2)

himself65:

BREAKING CHANGE:

Remove the `tool` | `function` type in MessageType; please use `assistant` instead. This is because these two types are only available for OpenAI. OpenAI deprecates the function type, and we support the Claude 3 tool call.

`toolCall` and `toolUse` will be the two options in messages; if `toolCall` is provided, it means **server** (which is AI) asks to call a function. If `toolUse` is provided, it means **client**(the User) used a tool and got a response. This  type of design follows Claude 3 (not OpenAI who puts function call into a new message role) since it will simplify the server-client model design and less mental burden


himself65 (commented):
I will remove `PlaceholderRecord` next day, seems like it's too strict for both side

PR #721, opened 2024-04-15
refactor: VectorStoreIndex: use TransformerComponent to calc embeddings 

marcusschiesser:
Refactor VectorStoreIndex: use `TransformerComponent` to calc embeddings (for image nodes and non-image nodes)


himself65 (commented):
You might need to provide a mock llm in fixtures

PR #718, opened 2024-04-14
Toolretriever for Agent OpenAI broken

Wesselvdp:
Hi, I ran into some issues with the multi documents example. It seems like the toolRetriever for an OpenAI agent is broken. 

Since the constructor has a default value with `tools = []` on line 93, checking if tools are an array seems unnecessary. Additionally, the if block on line 112 was always true. This resulted in the `toolRetriever` never working. 

Also, are you hiring? 🚀

Issue #557, opened 2024-02-21
Cannot provide system prompt with `OpenAIAgent`


Aryik:
If you create an `OpenAIAgent` like this, `llamaindex` will throw an exception: 
```
  const agent = new OpenAIAgent({
    tools: [],
    verbose: true,
    systemPrompt,
  });
```

This is the exception: 
```
Error: Cannot provide both systemPrompt and prefixMessages
    at new OpenAIAgent (node_modules/llamaindex/index.js:3953:23)
```

This is because in the constructor, you're doing `prefixMessages = prefixMessages || []`, and then you are throwing an exception if `systemPrompt` is passed in and `prefixMessages` is truthy: https://github.com/run-llama/LlamaIndexTS/blob/d4eda9f3965969d89d396e26d04fcaf2529f9431/packages/core/src/agent/openai/base.ts#L44

`prefixMessages` will always be truthy because of the `|| []`.

marcusschiesser:
@EmanuelCampos can you have a look please?

rawwerks:
related, but for queryEngine: https://github.com/run-llama/LlamaIndexTS/issues/606

PR #728, opened 2024-04-17
refactor: full typed & iterator of agent worker/runner (part 3)

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/692, https://github.com/run-llama/LlamaIndexTS/issues/557

Follow the original design doc: https://github.com/run-llama/llama_index/blob/5a6ffe32faa75db0b4737d1e7a85e6fe4afe94af/docs/module_guides/deploying/agents/agent_runner.md.

- Async Iterable task
- Remove complex & duplicated LLM-translated code (I spent 2 days to realize that what the original design means and how to reduce old TS code)

## Checklist
- [x] migrate OpenAIAgent
- [x] migrate ReActAgent
- [x] ReActAgent streaming
- [x] Test coverage
- [x] Object Retrieval
  - [x] Test case
  - [x] [Example](https://github.com/run-llama/LlamaIndexTS/pull/718#issuecomment-2056851997)

### Nice to have
- [ ] Mistral AI Agent 

In future step, we are going to support multi-model in agent


marcusschiesser (commented):
looks very good! will openai streaming also work?

packages/core/src/callbacks/CallbackManager.ts
```diff
@@ -205,10 +207,8 @@ export class CallbackManager implements CallbackManagerMethods {
     if (!handlers) {
       return;
     }
-    queueMicrotask(() => {
```
himself65, line 208: if put into microtask, the message might be modified

himself65: but I think it better put inside microtask, let me add this back later

packages/core/e2e/node/claude.e2e.ts
```diff
@@ -73,9 +74,8 @@ await test("anthropic agent", async (t) => {
     const result = await agent.chat({
       message: "What is the weather in San Francisco?",
     });
-    consola.debug("response:", result.response);
-    ok(typeof result.response === "string");
-    ok(result.response.includes("35"));
```
himself65, line 78: bring API back later

himself65 (commented):
getTools should be called in runtime

himself65:
Let me do the rest part today

himself65 (commented):
I will update document

apps/docs/docs/modules/agent/multi_document_agent.mdx
himself65, line 1: document are outdated, will update in the future

packages/core/src/callbacks/CallbackManager.ts
```diff
@@ -205,9 +212,10 @@ export class CallbackManager implements CallbackManagerMethods {
     if (!handlers) {
       return;
     }
+    const clone = structuredClone(detail);
```
marcusschiesser, line 215: can we use structuredClone now? i remember you shouldn't because of node 16 compatibility?

apps/docs/docs/modules/agent/query_engine_tool.mdx
particledream, line 1: @himself65 I'm having a hard time understanding why these files were deleted.
image

I see that you replaced with a single [example](apps/docs/docs/examples/agent.mdx). 

Where is now the documentation for [QueryEngineTool](https://github.com/run-llama/LlamaIndexTS/blob/98a2b4a5471a7492c4c7788003596555e52a9fe2/apps/docs/docs/modules/agent/query_engine_tool.mdx) and others that were previously under "Agents"?

I see all the examples are still [available as code](https://github.com/run-llama/LlamaIndexTS/tree/cfb90f7666d6ad4a92ceaf2a7df36f46d36330e8/examples/agent) in the `examples/agent` dir, but no longer in the MD documentation.

particledream: I just scanned the code again. I believe you plan on adding back after you've had a chance to update the documentation to reflect your most recent refactor.

himself65: Yeah the type was broken so I temporary removed it I will add back soon

examples/agent/react_agent.ts
```diff
@@ -62,18 +62,18 @@ async function main() {
   });
 
   // Create an ReActAgent with the function tools
-  const agent = new ReActAgent({
+  const agent = new ReACTAgent({
```
particledream, line 65: @himself65 I feel like going from `ReActAgent` to `ReACTAgent` would be considered a breaking change. Given llamaindex is still not v1, and if semantic versioning is being used, possibly a minor bump would have been warranted instead of a patch.

himself65: Thanks for reviewing, this is unintended. It's okay to revert the naming back

packages/core/src/callbacks/CallbackManager.ts
himself65: I'm not sure for now; I think we could use new APIs for the new module. And we could drop node16 support in 0.3.x

himself65: /cc @yisding

PR #729, opened 2024-04-17
fix: ensure to use build for examples

marcusschiesser:


marcusschiesser:
@yisding this fixes the build problem. Issue was: examples have been type-checked with `latest` llamaindex release in CI - but that didn't work as the example code already references new code from the local build. This PR ensures that we're using the local build.

> Note: we can't set llamaindex version in examples to `workspace:*`: this would use the local build for `pnpm` but won't work for just using `npm`

PR #734, opened 2024-04-19
feat: support jina ai embedding and reranker

ZiniuYu:
Support for Jina AI Embedding and Reranker via API:

- Integrated Jina AI's embedding and reranking features through the API.
  - Ensured compatibility and optimized performance for real-time data processing.
- Added documentation to guide users on how to utilize the new Jina AI embedding and reranking features.


apps/docs/docs/modules/node_postprocessors/jinaai_reranker.md
```diff
@@ -0,0 +1,71 @@
...
+retriever.similarityTopK = 5;
+```
+
+## Create a new instance of the JinaAIReranker class
+
+Then you can create a new instance of the `JinaAIReranker` class and pass in the number of results you want to return.
+The Jina AI Reranker API key is set in the `JINAAI_API_KEY` environment variable.
+
+```bash
+export JINA_API_KEY=<YOUR API KEY>
```
DresAaron, line 52: `export JINAAI_API_KEY` or `JINA_API_KEY`?

DresAaron (changes requested):

packages/core/src/postprocessors/rerankers/JinaAIReranker.ts
```diff
@@ -0,0 +1,89 @@
...
+  relevance_score: number;
+}
+
+export class JinaAIReranker implements BaseNodePostprocessor {
+  model: string = "jina-reranker-v1-base-en";
+  topN?: number;
+  apiKey?: string = undefined;
+
+  constructor(init?: Partial<JinaAIReranker>) {
+    this.topN = init?.topN;
```
Mini256, line 20:
```suggestion
    this.topN = init?.topN ?? 2;
```

Should we add default value for the topN parameter, like python version?

https://github.com/run-llama/llama_index/blob/aa13d47444692faa06b5753b7451b1920837b29c/llama-index-integrations/postprocessor/llama-index-postprocessor-jinaai-rerank/llama_index/postprocessor/jinaai_rerank/base.py#L26

ZiniuYu: Thanks for the comment! In fact we do not require topN in the api, the default value will be len(docs) if it's not provided. I will discuss whether we want a good value for it with my colleagues.

marcusschiesser (approved):

marcusschiesser:
@ZiniuYu Looks great, thanks. Can you call `pnpx changeset` and add a changeset? Then your change will end up in our changelog.md file

ZiniuYu:
@marcusschiesser Thanks for reviewing! The changeset is added

PR #744, opened 2024-04-22
refactor: use official ollama sdk

himself65:
v0.5.0 support CJS format


packages/core/src/llm/ollama.ts
```diff
@@ -211,4 +175,105 @@ export class Ollama extends BaseEmbedding implements LLM {
   async getQueryEmbedding(query: string): Promise<number[]> {
     return this.getEmbedding(query);
   }
+
+  // ollama specific methods, inherited from the ollama library
+  static async list() {
```
marcusschiesser, line 180: why do we add these delegates? can't the user just directly call the `ollama` package if he needs to use this functionality?
I guess there's a reason why, but then a comment would be helpful :)

PR #746, opened 2024-04-22
fix: agent class should implement ChatEngine interface

himself65:


himself65 (commented):
I noticed that there might has some type missing in @types/node. Let me see if I can do some contribution on upstream

PR #747, opened 2024-04-22
feat: use blob image in clip example

thucpn:


himself65 (commented):
could you please add a test case?

examples/multimodal/clip.ts
```diff
@@ -1,5 +1,16 @@
+import * as fs from "fs";
```
himself65, line 1: use { defaultFS } from `@llamaindex/env`

himself65: we are aiming to support multiple js runtime, so it's necessary to wrap system API into another package

PR #751, opened 2024-04-22
feat: overide generate hash function for image document

thucpn:


packages/core/src/Node.ts
```diff
@@ -326,6 +326,16 @@ export class ImageNode<T extends Metadata = Metadata> extends TextNode<T> {
     const absPath = path.resolve(this.id_);
     return new URL(`file://${absPath}`);
   }
+
+  generateHash() {
+    const hashFunction = createSHA256();
```
marcusschiesser, line 331: how does this work if two images have the same id but different data in the blob?
can you add a unit test for this and make sure this works

PR #752, opened 2024-04-22
test: add edge-runtime in vitest

himself65:


himself65:
it cannot test the real next.js environment, closing

PR #754, opened 2024-04-22
feat: http(s) support for file readers

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/discussions/495

### DIFF
--- a/packages/core/src/readers/ImageReader.ts
+++ b/packages/core/src/readers/ImageReader.ts
@@ -16,11 +16,11 @@ export class ImageReader implements FileReader {
    * @returns Promise<Document[]> A Promise object, eventually yielding zero or one ImageDocument of the specified file.
    */
   async loadData(
-    file: string,
+    file: string | URL,
     fs: GenericFileSystem = defaultFS,
   ): Promise<Document[]> {
     const dataBuffer = await fs.readRawFile(file);
     const blob = new Blob([dataBuffer]);
-    return [new ImageDocument({ image: blob, id_: file })];
+    return [new ImageDocument({ image: blob, id_: `${file}` })];
   }
 }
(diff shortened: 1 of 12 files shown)


packages/core/src/readers/ImageReader.ts
```diff
@@ -16,11 +16,11 @@ export class ImageReader implements FileReader {
...
    */
   async loadData(
-    file: string,
+    file: string | URL,
     fs: GenericFileSystem = defaultFS,
   ): Promise<Document[]> {
     const dataBuffer = await fs.readRawFile(file);
     const blob = new Blob([dataBuffer]);
-    return [new ImageDocument({ image: blob, id_: file })];
+    return [new ImageDocument({ image: blob, id_: `${file}` })];
```
marcusschiesser, line 24: if `file` is an `URL` then you can also use
`[new ImageDocument({ image: file, id_: `${file}` })];`
because `image` can be `URL`
(I guess this only makes sense for URLs with `http(s)://` prefix as we might send the image's URL to a LLM)

himself65: The file system also can read URLs if it's `file:` schema

like 
```ts
fs.readFile(new URL('file:/path/to/file'))
```

So, it could be error if pass file URL or non-public URL to LLM

marcusschiesser: Yes, you're right. I forgot that an `http(s)://` URL can still be non-public, and we can't find out whether it is public or not, so it's better to use `Blob` as a general case.
It might be worth adding a dedicated function for reading a public URL though

himself65: so there are many cases:
1. string
  1.1 base64 ✅
  1.2 http(s) ✅
  1.3 others
2. URL
  2.1. image URL
    2.1.1 nonpublic
    2.1.2 public ✅
  2.2 non-image URL
3. blob
  3.1 image blob (png/jpg...) ✅
  3.2 non-image blob (docx, pdf...)

PR #757, opened 2024-04-23
fix: content type

himself65:


himself65:
looks not perfect. closed

PR #760, opened 2024-04-24
fix: use Array.prototype.flat instead of Array.prototype.push(...spread)

ezirmusitua:
if `reader.loadData` return large amount of documents, the `IngestionPipeline.prepareInput` will throw "RangeError: Maximum call stack size exceeded", use Array.concat to avoid it.


himself65 (approved):
LGTM

I put some contexts here: https://stackoverflow.com/questions/61740599/rangeerror-maximum-call-stack-size-exceeded-with-array-push

PR #768, opened 2024-04-25
fix: use sha256 hash value as the `Document.id_` in `MarkdownReader`

ezirmusitua:
`MarkdownReader` uses `randomUUID()` as the `Document.id_`, which invalidate `DocStoreStrategy.UPSERTS` and `DocStoreStrategy.UPSERTS_AND_DELETE`, check [here]https://github.com/run-llama/LlamaIndexTS/blob/74686f5776d4707815e1f3f761dabc5393b29779/packages/core/src/ingestion/strategies/classify.ts#L13)


packages/core/src/readers/MarkdownReader.ts
```diff
@@ -95,15 +95,15 @@ export class MarkdownReader implements FileReader {
...
+      const sha256 = createSHA256()
       if (header) {
-        results.push(
-          new Document({
-            text: `\n\n${header}\n${value}`,
-          }),
-        );
+        const text = `\n\n${header}\n${value}`;
+        sha256.update(text);
+        results.push(new Document({ text, id_: sha256.digest() }));
```
marcusschiesser, line 103: @ezirmusitua great catch! The other readers are using `file` as `id_`. How about doing the same here (and appending a suffix as it seems more than one document is generated per file here)

ezirmusitua: In my project i use pdf and markdown documents, so i use sha256 digest, may be we should use `file` in `PDFReader` too?

https://github.com/run-llama/LlamaIndexTS/blob/ca348a657053b8f3d3427029d9dfcf343ed03fe7/packages/core/src/readers/PDFReader.ts#L21

himself65: Hash content may be costly if it's enormous. So maybe filePath or something?

marcusschiesser: @himself65 yes, that's what I mean: using `file` as ID. @ezirmusitua feel free to send a PR - otherwise, we will for sure change this in an upcoming release

himself65:
Could you please fix the lint issue?

marcusschiesser (approved):
Thanks a lot @ezirmusitua - that looks good!

PR #769, opened 2024-04-25
feat: init gemini llm

thucpn:
Follow logic from llama-index repo:
https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py

### DIFF
--- a/packages/core/src/llm/gemini.ts
+++ b/packages/core/src/llm/gemini.ts
@@ -137,7 +137,7 @@ export class GeminiSessionStore {
 /**
  * Helper class providing utility functions for Gemini
  */
-export class GeminiHelper {
+class GeminiHelper {
   // Gemini only has user and model roles. Put the rest in user role.
   public static readonly ROLES_TO_GEMINI: Record<
     MessageType,
--- a/packages/core/src/llm/gemini.ts
+++ b/packages/core/src/llm/gemini.ts
@@ -5,13 +5,12 @@ import {
   type Part,
 } from "@google/generative-ai";
 import { getEnv } from "@llamaindex/env";
-import { BaseEmbedding } from "../embeddings/types.js";
+import { ToolCallLLM } from "./base.js";
 import type {
   ChatMessage,
   ChatResponse,
   ChatResponseChunk,
   CompletionResponse,
-  LLM,
   LLMChatParamsNonStreaming,
   LLMChatParamsStreaming,
   LLMCompletionParamsNonStreaming,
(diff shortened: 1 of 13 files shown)


yisding (approved):
This looks good thanks! Let's also add Gemini-1.5 because I know people will want to experiment with the long context windows even though it's still in beta.

packages/core/src/llm/gemini.ts
```diff
@@ -0,0 +1,381 @@
...
+
+/**
+ * Gemini Session to manage the connection to the Gemini API
+ */
+export class GeminiSession {
+  gemini: GoogleGenerativeAI;
+
+  constructor(options: GeminiSessionOptions) {
+    if (!options.apiKey) {
+      options.apiKey = getEnv("ANTHROPIC_API_KEY");
```
marcusschiesser: Wrong key name?

packages/core/src/llm/gemini.ts
```diff
@@ -0,0 +1,381 @@
...
+    const result = await client.generateContent(
+      GeminiHelper.messageContentToGeminiParts(prompt),
+    );
+    return {
+      text: result.response.text(),
+      raw: result.response,
+    };
+  }
+
+  private async getEmbedding(prompt: string): Promise<number[]> {
```
marcusschiesser, line 368: Better seperate embedding and LLM model usage in two classes (like the other LLM implementations except Ollama)

packages/core/src/llm/gemini.ts
```diff
@@ -0,0 +1,381 @@
...
+      this.sessions.push({ session, options });
+    }
+    return session;
+  }
+}
+
+/**
+ * Helper class providing utility functions for Gemini
+ */
+export class GeminiHelper {
```
marcusschiesser, line 141: Needs to be exported?

packages/core/src/llm/gemini.ts
```diff
@@ -0,0 +1,381 @@
...
+      throw new Error("Set Google API Key in GOOGLE_API_KEY env variable");
+    }
+    this.gemini = new GoogleGenerativeAI(options.apiKey);
+  }
+}
+
+/**
+ * Gemini Session Store to manage the current Gemini sessions
+ */
+export class GeminiSessionStore {
```
marcusschiesser, line 113: Do we need multiple sessions?

marcusschiesser (commented):
@thucpn please add also to docs with a short example

marcusschiesser (approved):
Thanks @thucpn that looks great!

PR #775, opened 2024-04-26
feat(core): 0.3.0 release

himself65:
## Updates

- add support for Cloudflare(maybe WinterJS and others also support, but I didn't test). Better support for next.js (you don't need to install `@llamaindex/edge`)
- set default node.js version to `20` as it's a LITS; we might no longer consider 16 support but still support 18~22

## Checklist

- [x] Update blog
  - [x] README.md
  - [x] Agent
  - [x] Function Tool
  - [x] TypeScript Support
- [x] Test Production
  - Edge runtime
  - Cloudflare worker
- [x] Agent examples
  - [x] with Cloudflare Worker
  - [x] with Next.js + Vercel AI RSC
  - [x] with Waku (vite)

## QA part

(asked by @marcusschiesser)

### Q: AsyncIterator could be transformed to ReadableStream easily 

Yes, ReadableStream is async iterable by native, and all of the async iterator could be translated to ReadableStream in a few lines. (for example https://github.com/vercel/ai/blob/73356a9d46f1d271e3278cb92bd8ebe92a6a058d/packages/core/streams/ai-stream.ts#L284)

But this does not mean we should still return only an async iterator. Readable Stream could bring more features to the developer side.

1. tee, pipeTo, pipeThough, and Stream helpers, which are already existing useful APIs
2. cloneable, a stream could be cloned which would benefit whose wanna re-use a response

```ts
const r = new ReadableStream({
	start: (controller) => {
		controller.enqueue('1');
		controller.enqueue('2');
		controller.enqueue('3');
		controller.close();
	},
});

const [a, b] = r.tee()

for await (const v of a) {
	console.log('v', v);
}

const cp = structuredClone(b, {
	transfer: [b],
})

for await (const v of cp) {
	console.log('v', v);
}
// outputs:
// v 1
// v 2
// v 3
// v 1
// v 2
// v 3
```

### Is this a breaking change?

No, Readable Stream is already async iterable (which was async iterable in the same). But we announce this as new version to show that with Readable could do more things

### What about other old APIs that use AsyncIterator?

Personally, I think we should change them, too, but this is not an emergency and not have extra benefit, but only code style changes which I think would benefit for reading?

```ts
// old
function f() {
  return (function*(){})() // this is ugly to return a generator
}
// new 
function g() {
  return new ReadableStream({
    start(controller) {controller.end()}
  })
}
```


packages/core/e2e/examples/cloudflare-worker-agent/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    const textEncoder = new TextEncoder();
+    const response = responseStream.pipeThrough<Uint8Array>(
+      // @ts-expect-error: see https://github.com/cloudflare/workerd/issues/2067
+      new TransformStream({
+        transform: (chunk, controller) => {
+          controller.enqueue(textEncoder.encode(chunk.response.delta));
+        },
+      }),
+    );
+    // @ts-expect-error: see https://github.com/cloudflare/workerd/issues/2067
```
himself65, line 26: tip: https://github.com/cloudflare/workerd/issues/2067

packages/core/e2e/examples/cloudflare-worker-agent/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "devDependencies": {
+    "@cloudflare/vitest-pool-workers": "^0.2.3",
+    "@cloudflare/workers-types": "^4.20240423.0",
+    "typescript": "^5.4.5",
+    "vitest": "1.3.0",
+    "wrangler": "^3.52.0"
+  },
+  "dependencies": {
+    "llamaindex": "workspace:*"
```
marcusschiesser, line 22: might be a bit confusing that cloudflare is using `llamaindex` and for edge we have a dedicated package. 
Technically, we don't have a reason for a dedicated edge package; the idea was that we could track usage separately.

himself65: I think this use same package name will be better, It's not a suggeseted way to seperate pacakges

himself65 (commented):
tomorrow I will release a beta version and try to bump create-llama and chat-llamaindex

marcusschiesser:
@himself65 for create-llama you might want to use this PR: https://github.com/run-llama/create-llama/pull/63 as @thucpn already updated some code for your agents there

PR #777, opened 2024-04-27
fix(core): polyfill Web Stream APIs

himself65:
feedback from: https://discord.com/channels/1059199217496772688/1133167189860565033/1233430729879453738

Reason is lower version typescript doesn't have built-in async iterator support for Readable Stream

This PR will restrict the type check by set `types` to empty array, and remove "DOM" on the top config to make sure each sub-module uses correct types


himself65 (commented):
find a bug when running streaming mode, fixing

examples/agent/step_wise_openai.ts
```diff
@@ -69,11 +70,24 @@ async function main() {
...
     console.log(`======== OUTPUT ==========`);
-    console.log(stepOutput.output.message.content);
+    const output = stepOutput.output;
+    if (output instanceof ReadableStream) {
+      for await (const chunk of output) {
+        process.stdout.write(chunk.delta);
+      }
+    } else {
+      console.log(output);
+    }
```
himself65, line 80: This seems ugly to use, but since this not for user side API, I think it's okay for now

Issue #794, opened 2024-05-02
streaming support for ReActAgent


tom-leamon:
Streaming for OpenAIAgent was added by #788. Attempting to use AnthropicAgent results in the error: 

`Error: Anthropic does not support streaming` 

This Error message is a little dubious, because it's clear Anthropic does actually support streaming, so I assume this to be a limitation of the current implementation of agent streaming, in which case the error should be: 

`Error: LlamaIndex does not yet support streaming for AnthropicAgent`

Additionally, I attempted to use ReActAgent streaming with Llama3-70B on Groq. The call seemed to just freeze and never return any results. 

Non-streaming works properly for both AnthropicAgent and ReActAgent. I would like to clarify if these are just limitations of the current agent streaming abilities, or if it's actually true that this is not possible with Anthropic.

himself65:
1. Anthroipc doesn't support streaming https://docs.anthropic.com/claude/docs/tool-use

himself65:
Thanks for report, seems like there is a bug in ReAct Agent that will exit silently

himself65:
Seems like node.js pitfall or bug

PR #798, opened 2024-05-02
fix: streaming for ReAct Agent

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/794


pnpm-lock.yaml
himself65, line 1: Why do we have a 20K diff for the lock file? Seems like it upgraded

PR #801, opened 2024-05-03
feat: add top k to `asQueryEngine`

yisding:


packages/core/src/indices/vectorStore/index.ts
```diff
@@ -279,18 +279,31 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
     return new VectorIndexRetriever({ index: this, ...options });
   }
 
+  /**
+   * Create a RetrieverQueryEngine.
+   * similarityTopK is only used if no existing retriever is provided.
+   * @param options
+   * @returns
```
himself65, line 286:
```suggestion
```

himself65: since it describes nothing

himself65 (approved):

PR #803, opened 2024-05-03
feat: add `with` helper apis

himself65:
Suggested by @marcusschiesser


examples/Settings.ts
```diff
@@ -26,4 +33,13 @@ async function main() {
   console.log(response.toString());
 }
 
-main().catch(console.error);
+withCallbacks(
```
marcusschiesser, line 36: @himself65 very cool - seems also like we can chain it, right? 
Then in this file we could do e.g.:

```suggestion
withLLM(new OpenAI({ model: "gpt-4" }),
withCallbacks(
```

PR #808, opened 2024-05-04
feat: LlamaParseReader: update supported file types, add support for array of file Paths, add support for directory Paths + example

InsightByAI:
Should be mostly up to date with the python version now.

Main thing still missing is get_images function, will look at that next.


InsightByAI:
I tried adding implementing it more with SimpleDirectoryReader by also defining numWorkers and adding parallel processing, but I had issues with it not respecting the numWorkers from LlamaParseReader, so I added directory reading straight into LlamaParseReader.

If someone wants to check out parallel processing in SimpleDirectoryReader and maybe has an idea how to respect the number of "workers": 
[Commit](https://github.com/run-llama/LlamaIndexTS/commit/9b734a4398fe0f10e0a7856a7e34423f6b0d471f)

InsightByAI:
Something causing a Connect Timeout Error at loadFile and loadData, gonna check that out

packages/core/src/readers/type.ts
```diff
@@ -15,7 +15,19 @@ export interface FileReader extends BaseReader {
...
+/**
+ * A reader takes single and multiple file paths as well as a directory Path and imports data into an array of Document objects.
+ */
+export interface MultiReader extends BaseReader {
+  loadData(filePath: string, fs?: CompleteFileSystem): Promise<Document[]>;
+  loadData(filePaths: string[], fs?: CompleteFileSystem): Promise<Document[][]>;
+  loadData(
+    directoryPath: string,
+    fs?: CompleteFileSystem,
+  ): Promise<Document[][]>;
```
himself65, line 27: Should we support directory input?

InsightByAI: I'm gonna check it out again and see. I'll probably have some time in 2-3 days.

The Connect Timeout Error seems to be kind of random. Sometimes it works as expected, sometimes it times out, so I'll recheck everything anyway.

marcusschiesser: yes, the SimpleDirectoryReader already supports loading multiple files from a directory.

I think we have three concerns:
1. Walking the filesystem (already done by the SimpleDirectoryReader)
2. Reading a single file (done by the reader)
3. Loading files in parallel (new concern in this PR)

If we handle this new concern by a single reader (here LlamaParseReader), then we can only use it with this reader. 

How about adding `numWorkers` to SimpleDirectoryReader instead?

@InsightByAI if you like you can split your PR in two, the first one just adding the missing filetypes

packages/core/src/readers/utils.ts
```diff
@@ -0,0 +1,105 @@
+// For LlamaParseReader.ts
+
+export const SupportedFileTypes: string[] = [
```
marcusschiesser, line 3: this is related to llamaparse right? better add it then to the llama parse file

packages/core/src/readers/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  ".pdf",
```
marcusschiesser, line 4: @hexapode is there a way to get this list of supported filetypes somehow dynamically?

examples/readers/src/llamaparse_2.ts
```diff
@@ -1,26 +1,31 @@
...
     resultType: "markdown",
     language: "en",
+    numWorkers: 3, //Load files in batches of 2
     parsingInstruction:
-      "The provided document is a manga comic book. Most pages do NOT have title. It does not contain tables. Try to reconstruct the dialogue happening in a cohesive way. Output any math equation in LATEX markdown (between $$)",
+      "The provided documents are datasheets and Quick-Installation-Guides for Solplanet's Ai-LB series of batteries. They contain tables and graphics. There is also a lot of technical information. The goal is to extract and structure the knowledge in a coherent way",
   });
-  const documents = await reader.loadData("../data/manga.pdf"); // The manga.pdf in the data folder is just a copy of the TOS, due to copyright laws. You have to place your own. I used "The Manga Guide to Calculus" by Hiroyuki Kojima
+  // Can either accept a single file path an array[] of file paths or a directory path
+  const documents = await reader.loadData("../data/LlamaParseData");
```
marcusschiesser, line 13: can we just use all the files in the existing data folder?
```suggestion
  const documents = await reader.loadData("../data");
```
we can add new files, but then we need to take care about the license

packages/core/src/readers/type.ts
InsightByAI: @marcusschiesser The issue for me with adding `numWorkers` to SimpleDirectoryReader was that it would load in batches just fine, but it would load the next batch right after finishing the 1st without waiting for the parse results.

I think the options would be to either:
1. have SimpleDirectoryReader wait for a finish signal from the Parser
2. have SimpleDirectoryReader pass on a list of all filePaths and let the parallel loading be handled by the Parser

I concentrated only on LlamaParse because I think the requirements for `numWorkers` in most other readers is quite different, since in LlamaParse the "workers" kind of refers to the max amount of files the LlamaParse API can accept at once, which is limited to 10.

If I have powerful enough hardware, I could probably use a 100 or more "workers" to load a 100 files parallel with PDFReader. 
Not sure if a `numWorkers` that limits one reader to 10 workers and has to wait for results from an API endpoint, while allowing another one to use unlimited workers and only be limited by hardware speed would be easy to implement/work that well? 

I'm going to check how python implements this. The Llama-Parse standalone can accept arrays of files but not a directory afaik, so I'll check how that works with an equivalent llama-index python directory Reader.

I'll split the PR later. Thanks for the elaborate review!

examples/readers/src/llamaparse_2.ts
InsightByAI: Yeah sure, could also just split the brk-2022.pdf file into 6 parts and put them into a separate folder to showcase parallel processing if that makes things easier.

packages/core/src/readers/type.ts
marcusschiesser: Can you re-add `numWorkers` to SimpleDirectorReader and add a test case for the problem? Then we can have a look at it

PR #817, opened 2024-05-07
fix: PGVector similarity score

JT-Dev-215:
Similarity score should be 1 minus the resulting value from the PG cosine distance. This change now matches with the Python package.


himself65 (approved):
https://github.com/run-llama/llama_index/blob/2d3b7186318b36298543c749510abdc2640be7cb/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py

PR #820, opened 2024-05-08
fix: max tokens

himself65:
Fixes: https://discord.com/channels/1059199217496772688/1133167189860565033/1237394506274439268

### DIFF
--- a/packages/core/src/ChatHistory.ts
+++ b/packages/core/src/ChatHistory.ts
@@ -65,6 +65,13 @@ export class SimpleChatHistory extends ChatHistory {
   }
 }
 
+export type SummaryChatHistoryOptions = {
+  /**
+   * some LLM instance (like Ollama) might doesn't have a maxTokens metadata, so you should set it manually
+   */
+  maxTokens: number;
+};
+
 export class SummaryChatHistory extends ChatHistory {
   /**
    * Tokenizer function that converts text to tokens,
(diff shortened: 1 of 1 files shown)


packages/core/src/ChatHistory.ts
```diff
@@ -77,20 +84,23 @@ export class SummaryChatHistory extends ChatHistory {
...
+  constructor(
+    init?: Partial<SummaryChatHistory> & Partial<SummaryChatHistoryOptions>,
+  ) {
     super();
     this.messages = init?.messages ?? [];
     this.messagesBefore = this.messages.length;
     this.summaryPrompt = init?.summaryPrompt ?? defaultSummaryPrompt;
     this.llm = init?.llm ?? new OpenAI();
-    if (!this.llm.metadata.maxTokens) {
+    if (!this.llm.metadata.maxTokens || !init?.maxTokens) {
```
himself65, line 97: Should we use `Infinity` if it's undefined?

marcusschiesser: the code below won't work if maxTokens is not specified, so I think it's ok, to throw this error, it's basically a configuration error.

this class also needs a tokenizer to work correctly (and for ollama it could be any LLM with any tokenizer). 

So I think we should set `tokenizer` to `globalsHelper.defaultTokenizer.encode` only if the LLM is OpenAI and throw an error if `tokenizer` is not explicitly specified otherwise

PR #821, opened 2024-05-08
feat: bind embedding models to vec stores and use vector store map

marcusschiesser:
- bind `embedModel` to vector stores
- use multiple vector stores and split them by the modality type that they should store
- add `vectorStores` to storage context to define vector store per modality


packages/core/src/indices/vectorStore/index.ts
```diff
@@ -231,10 +212,12 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
     }
 
     // use doc store strategy to avoid duplicates
+    // TODO: use doc strategy with multiple vector stores
```
marcusschiesser, line 215:
```suggestion
```

packages/core/src/indices/vectorStore/index.ts
```diff
@@ -248,20 +231,30 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
     return await this.init(args);
   }
 
-  static async fromVectorStore(
-    vectorStore: VectorStore,
-    serviceContext?: ServiceContext,
-    imageVectorStore?: VectorStore,
-  ) {
-    if (!vectorStore.storesText) {
+  static async fromVectorStore({
```
marcusschiesser, line 234: maybe better add `fromVectorStores`?

packages/core/src/indices/vectorStore/index.ts
```diff
@@ -415,47 +402,35 @@ export class VectorIndexRetriever implements BaseRetriever {
...
     query: string,
+    vectorStore: VectorStore,
     preFilters?: MetadataFilters,
   ): Promise<NodeWithScore[]> {
     const options = {};
     const q = await this.buildVectorStoreQuery(
-      this.index.embedModel,
+      vectorStore.embedModel,
       query,
+      // TODO: use different topK for image and text
```
marcusschiesser, line 429: TODO

packages/core/src/storage/StorageContext.ts
```diff
@@ -53,20 +52,19 @@ export async function storageContextFromDefaults({
...
-          path.join(persistDir, DEFAULT_IMAGE_VECTOR_NAMESPACE),
+    if (!(ObjectType.TEXT in vectorStores)) {
+      vectorStores[ObjectType.TEXT] =
+        vectorStore ??
+        ((await SimpleVectorStore.fromPersistDir(
+          persistDir,
           fs,
-        )
-      : imageVectorStore;
+        )) as unknown as VectorStore);
```
marcusschiesser, line 61: @himself65 this is not very nice. How can I make `SimpleVectorStore.fromPersistDir` return a `VectorStore` given the fact that it's build by the mixin (see source of SimpleVectorStore)?

examples/multimodal/load.ts
```diff
@@ -25,12 +27,16 @@ async function generateDatasource() {
     const documents = await new SimpleDirectoryReader().loadData({
       directoryPath: path.join("multimodal", "data"),
     });
-    const storageContext = await storageContextFromDefaults({
-      persistDir: "storage",
-      storeImages: true,
+    const clipVectorStore = new SimpleVectorStore({
+      embedModel: new ClipEmbedding(),
```
marcusschiesser, line 31: @himself65 `pnpm type-check` produces here:
```
Object literal may only specify known properties, and 'embedModel' does not exist in type '{ data?: SimpleVectorStoreData | undefined; }'.
```
doesn't make a lot of sense to me as `SimpleVectorStore` is produced by a mixin and the mixin's constructor allows any

examples/multimodal/rag.ts
```diff
@@ -41,7 +41,7 @@ async function main() {
 
   const queryEngine = index.asQueryEngine({
     responseSynthesizer: new MultiModalResponseSynthesizer(),
-    retriever: index.asRetriever({ similarityTopK: 3, imageSimilarityTopK: 1 }),
+    retriever: index.asRetriever({ topK: { text: 3, image: 1 } }),
```
marcusschiesser, line 44: should probably better use uppercase here
```suggestion
    retriever: index.asRetriever({ topK: { TEXT: 3, IMAGE: 1 } }),
```

packages/core/src/ingestion/IngestionPipeline.ts
```diff
@@ -176,3 +189,31 @@ export class IngestionPipeline {
...
+    newIds: string[],
+    nodes: BaseNode<Metadata>[],
+    vectorStore: VectorStore,
+  ) => Promise<void>,
+) {
+  const nodeMap = splitNodesByType(nodes);
+  for (const type in nodeMap) {
+    const nodes = nodeMap[type as ModalityType];
+    if (nodes) {
+      console.log(type);
```
marcusschiesser, line 206: remove

examples/multimodal/load.ts
himself65: I added some changes, I think type here is indeed wrong

.github/workflows/test.yml
```diff
@@ -75,9 +75,6 @@ jobs:
         working-directory: ./examples
       - name: Run Type Check
         run: pnpm run type-check
-      - name: Run Circular Dependency Check
-        run: pnpm run circular-check
```
marcusschiesser, line 79: @himself65 adding the embedModel to the vector store base class added a lot of circular dependencies which seem to be hard to resolve.

I checked that `madge` is checking circular deps on a file basis, not on object level. Seems too strict to me - why did you add it?

himself65: No, check circular deps on file base makes sense. It will cause incorrect behavior for bundler like using Next.js
And there is no meaning to check on object level

himself65: Dont remove this

himself65: This is a very common bug, see https://stackoverflow.com/a/44727578

PR #822, opened 2024-05-08
fix: Gemini text chat - prevent sending broken messageContent and history

parhammmm:
Gemini was completely broken for text chat and was getting confused when prompted with an empty context.

It is still broken for vision. Currently the `mimeType` is incorrect and `inlineData` isn't taken into account. Will submit a separate PR to address.

### DIFF
--- a/packages/core/src/Prompt.ts
+++ b/packages/core/src/Prompt.ts
@@ -371,6 +371,7 @@ export function messagesToHistoryStr(messages: ChatMessage[]) {
 }
 
 export const defaultContextSystemPrompt = ({ context = "" }) => {
+  if (!context) return "";
   return `Context information is below.
 ---------------------
 ${context}
--- a/.changeset/sharp-knives-ring.md
+++ b/.changeset/sharp-knives-ring.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+Improve Gemini message and context preparation
--- a/packages/core/src/llm/gemini.ts
+++ b/packages/core/src/llm/gemini.ts
@@ -139,7 +139,7 @@ class GeminiHelper {
   > = {
     user: "user",
     system: "user",
-    assistant: "user",
+    assistant: "model",
     memory: "user",
   };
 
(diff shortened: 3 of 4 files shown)


packages/core/src/Prompt.ts
```diff
@@ -371,6 +371,7 @@ export function messagesToHistoryStr(messages: ChatMessage[]) {
 }
 
 export const defaultContextSystemPrompt = ({ context = "" }) => {
+  if (!context) return "";
```
marcusschiesser, line 374: @parhammmm looks like a great idea, but why did you add it to this PR? did you run in a specific error condition? If so, can you send a new PR for this?

marcusschiesser (commented):
@parhammmm looks great - do you have a test case that failed with the old implementation? if so, can you add it to 'examples'?

packages/core/src/Prompt.ts
parhammmm: @marcusschiesser cheers! I put it here since it was quite Gemini specific.`Context information is below.` would force Gemini to respond with `I cannot answer the question because the provided context does not mention ...` I can make a new PR if you still prefer it that way

parhammmm:
> @parhammmm looks great - do you have a test case that failed with the old implementation? if so, can you add it to 'examples'?

@marcusschiesser 

The test case would be the following, I'm not sure if you want the following in the `examples` as it requires a fake response from the llm? Let me know if that's ok and I'll add it or not 

```ts
import { Gemini, GEMINI_MODEL } from "llamaindex";

(async () => {
  if (!process.env.GOOGLE_API_KEY) {
    throw new Error("Please set the GOOGLE_API_KEY environment variable.");
  }
  const gemini = new Gemini({
    model: GEMINI_MODEL.GEMINI_PRO,
  });
  const result = await gemini.chat({
    messages: [
      { content: "You want to talk in rhymes.", role: "system" },
      {
        content: [{ type: "text", text: "Who is woodchuck?" }],
        role: "user",
      },
      {
        content: [{ type: "text", text: "Woodchuck is a Bird" }],
        role: "assistant",
      },
      {
        content:
          "How much wood would a woodchuck chuck if a woodchuck could chuck wood?",
        role: "user",
      },
    ],
  });
  console.log(result);
})();
```

There were two issues:
1. role = `assistant` was being mapped to Gemini's `user`, causing the entire history to be merged into one 
2. when `MessageContent` was of the type `MessageContentDetail[]` the message sent would be:

```ts
[
  {
    text: 'You want to talk in rhymes.\n' +
      '[object Object]\n' +
      '[object Object]\n' +
      'How much wood would a woodchuck chuck if a woodchuck could chuck wood?'
  }
]
```

parhammmm:
@marcusschiesser I also have the Gemini Vertex implementation ready for PR after this is merged in

marcusschiesser:
Thanks a lot @parhammmm - just tested it. Great work!

marcusschiesser:
Looking forward to your Vertex implementation

PR #824, opened 2024-05-09
feat: add nextjs plugin for `llamaindex`

himself65:
We are maintaining next.js plugin for polyfill the upstream dependencies.

Other solution will be:

1. Fork & maintain transformer package by ourselves
2. Let user do this

Fixes: https://discord.com/channels/1059199217496772688/1133167189860565033/1237677833627635793


packages/core/e2e/examples/nextjs-edge-runtime/src/utils/llm.ts
```diff
@@ -1,9 +1,18 @@
 "use server";
 // test runtime
+import { AutoTokenizer } from "@xenova/transformers";
```
himself65, line 3: we cannot directly import, will cause bundler error because they cannot find .node file

marcusschiesser (approved):
great!

PR #827, opened 2024-05-09
feat: send retrieve start and end events

marcusschiesser:


packages/core/src/indices/vectorStore/index.ts
```diff
@@ -411,18 +406,34 @@ export class VectorIndexRetriever implements BaseRetriever {
     this.imageSimilarityTopK = imageSimilarityTopK ?? DEFAULT_SIMILARITY_TOP_K;
   }
 
+  @wrapEventCaller
   async retrieve({
     query,
     preFilters,
   }: RetrieveParams): Promise<NodeWithScore[]> {
+    Settings.callbackManager.dispatchEvent("retrieve-start", {
```
himself65, line 414: use `getCallbackManager` here, import from internal

PR #838, opened 2024-05-15
feat: Gemini Access via Vertex AI

parhammmm:
@marcusschiesser here's the Vertex implementation. It's a little different than the others because with vertex there isn't any api key auth. 

To authenticate for local development:

```bash
npm install @google-cloud/vertexai
gcloud auth application-default login
```

For production the preferred method is via a service account, more details: https://cloud.google.com/docs/authentication/ . The `GeminiVertexSession` accepts all the params supported by `import { VertexAI } from "@google-cloud/vertexai";`. e.g.
      
```ts
new GeminiVertexSession({
  location: "",
  project: "",
  googleAuthOptions: {...},
})
```

Google Auth Options takes all the values from `GoogleAuthOptions` in https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/googleauth.ts incase someone wants to directly pass their own `keyFile` or it's values via `credentials`

Once logged in you can test via: 

```bash
export GOOGLE_VERTEX_PROJECT="" GOOGLE_VERTEX_LOCATION="us-central1" && npx ts-node ./gemini/chatVertex.ts
```

Some notes:

The `@google-cloud/vertexai` sdk is lagging behind `@google/generative-ai`, main differences affecting this PR:

- embeddings aren't supported.  I've opened an issue https://github.com/googleapis/nodejs-vertexai/issues/340
- vertexai doesn't have `EnhancedGenerateContentResponse` yet so added the missing pieces here too


parhammmm:
Not sure I quite follow why the example checks are failing, would you have any ideas?

marcusschiesser:
@parhammmm you have to run `pnpm type-check` to get more details about what is failing

packages/core/src/llm/gemini.ts
```diff
@@ -98,32 +203,137 @@ export class GeminiSession {
...
     options: GeminiSessionOptions;
   }> = [];
 
+  private static getSessionId(options: GeminiSessionOptions): string {
+    if (options.backend === GEMINI_BACKENDS.GOOGLE)
+      return options?.apiKey ?? "";
+    if (options.backend === GEMINI_BACKENDS.VERTEX)
+      return `${options.location}/${options.project}`;
+
+    return "";
```
marcusschiesser, line 311: i guess that's not something that should ever happen - throw an exception?

marcusschiesser (commented):
Looks great thanks @parhammmm! Please fix the builld, then we can merge it

packages/core/src/llm/gemini.ts
```diff
@@ -98,32 +203,137 @@ export class GeminiSession {
...
+    return streamConverter(result.stream, (response) => ({
+      text: this.getResponseText(response),
+      raw: response,
+    }));
+  }
+}
+
+/* To use Google's Vertex AI backend, it doesn't use api key authentication.
+ *
+ * To authenticate for local development:
```
marcusschiesser, line 236: can you update https://github.com/run-llama/LlamaIndexTS/blob/main/apps/docs/docs/modules/llms/available_llms/gemini.md and add it there?

packages/core/src/llm/gemini.ts
parhammmm: makes sense, sure!

packages/core/src/llm/gemini.ts
parhammmm: will do!

parhammmm:
@marcusschiesser I found the issue; it's because Vertex can only be used via the backend and here webpack is attempting to build `https://github.com/googleapis/google-auth-library-nodejs` for the browser. 

Have you dealt with such cases before?

parhammmm:
@marcusschiesser  done. I had to restructure the code a little bit to make sure node specific code doesn't break the browser/cloudflare builds.... on to bedrock implementation (pretty close)!

parhammmm:
@marcusschiesser @himself65 how do I use exported items from `packages/core/src/index.ts` seems like they aren't actually exported during the build... I've tried with `HuggingFaceEmbedding` (and `GeminiVertexSession`) and in nextjs I get:

`Attempted import error: 'HuggingFaceEmbedding' is not exported from 'llamaindex' (imported as 'HuggingFaceEmbedding').`

packages/core/src/index.ts
```diff
@@ -10,3 +10,6 @@ export {
   HuggingFaceEmbedding,
   HuggingFaceEmbeddingModelType,
 } from "./embeddings/HuggingFaceEmbedding.js";
+
+export { type VertexGeminiSessionOptions } from "./llm/gemini/types.js";
```
marcusschiesser, line 14: @parhammmm the exports that you add here are only used for NodeJS - so this change means VertexGemini does only work with NodeJS (not edge for example) - if that' ok, nothing to do - otherwise you can send a PR and move this to `src/llm/index.ts`

marcusschiesser (approved):
@parhammmm looks great thanks!

Issue #836, opened 2024-05-14
[Bug]: VectorStoreIndex.fromDocuments with .CSV file does not create OpenAPI embeddings


reidperyam:
### Bug Description

The following code fails to generate a vectorStorageIndex from a simple .CSV file (see attached MOCK_DATA.csv).

```
async function generateDatasource() {
  console.log(`Generating generateDatasource...`);
  console.log(`embedModel`,        Settings.embedModel);
  console.log(`STORAGE_CACHE_DIR`, STORAGE_CACHE_DIR);

  // Split documents, create embeddings and store them in the storage context
  const ms = await getRuntime(async () => {
    console.log(`Generating serviceContextFromDefaults...`);
    const serviceContext = await serviceContextFromDefaults({
      embedModel:    Settings.embedModel,
      chunkSize:     Number(process.env.CHUNK_SIZE),
      chunkOverlap:  Number(process.env.CHUNK_OVERLAP),
    });
    console.log(`Generating storageContextFromDefaults...`);
    const storageContext = await storageContextFromDefaults({
      persistDir: STORAGE_CACHE_DIR,
    });
    console.log(`getting docutments...`);
    const documents = await getDocuments();
    documents.forEach((doc) => {
      console.log(`document loaded: (${doc.id_})`);
    });
    console.log(`Generating VectorStoreIndex.fromDocuments...`);
    **[MOCK_DATA.csv](https://github.com/run-llama/llama_index/files/15310670/MOCK_DATA.csv)**
  });
  console.log(`Storage context successfully generated in ${ms / 1000}s.`);
}
```

### Version

0.2.10

### Steps to Reproduce

clone the following github repo:

https://github.com/reidperyam/practical-star/tree/master

see README.md for repro which I will copy here:

First, populate .env with OPENAI_API_KEY

install the dependencies:

npm install
Second, generate the embeddings of the documents in the ./data directory:

npm run generate

EXPECTED RESULT:

  generated cache/doc_store.json
  generated cache/index_store.json
  generated cache/vector_store.json

ACTUAL RESULT:

[MOCK_DATA.csv](https://github.com/run-llama/llama_index/files/15310681/MOCK_DATA.csv)
Error generating text embeddings: [see Relevant Logs/Tracbacks added here

### Relevant Logs/Tracbacks

```shell
[...]

PR #843, opened 2024-05-16
fix: truncate long document

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/836

There are some possible ways to solve that issue.

1. force split text right before embedding

This is not what the user expected, and must lose much of infos

2. Split documents inside of each loader.

Good, but it still will overflow the chunk size since metadata may overflow the chunk size.

### DIFF
--- a/packages/core/src/Node.ts
+++ b/packages/core/src/Node.ts
@@ -1,5 +1,6 @@
 import { createSHA256, path, randomUUID } from "@llamaindex/env";
 import _ from "lodash";
+import { Settings } from "./Settings.js";
 
 export enum NodeRelationship {
   SOURCE = "SOURCE",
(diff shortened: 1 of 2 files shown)


himself65 (commented):
Some ideas about the new design of Node classes

1. move BaseNode to interface
2. getContent to be value getter
3. metadata to be Map<string, JSONValue>
4. hidden text property
5. auto-memo id

```ts

interface BaseNode {
  readonly id: string;
  content: string;
}

type JSONObject = { [key: string]: JSONValue };
type JSONArray = JSONValue[];
type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
export type Metadata = Map<string, JSONValue>;

export const Settings = {
  chunkSize: undefined as number | undefined,
};

function chunkSizeCheck(
  originalGetter: () => string,
  context: ClassGetterDecoratorContext,
) {
  return function (this: Document) {
    const content = originalGetter.call(this);
    if (Settings.chunkSize !== undefined) {
      if (content.length > Settings.chunkSize) {
        console.warn(
          `Document (${this.id.substring(
            0,
            8,
          )}) is larger than chunk size: ${content.length}`,
        );
        console.warn(`Truncating content...`);
        console.warn("If you want to disable this warning:");
        console.warn("  1. Set Settings.chunkSize = undefined");
        console.warn("  2. Set Settings.chunkSize to a larger value");
        console.warn(
          "  3. Change the way of splitting content into different chunks",
        );
        return content.slice(0, Settings.chunkSize);
      }
    }
    return content;
  };
}

export class Document implements BaseNode {
  get id(): string {
    return Document.hash(this.text, this.metadata);
  }

  @chunkSizeCheck
  get content(): string {
    const leading = JSON.stringify([...this.metadata.entries()]);
    return `${leading ? `${leading}\n` : ""}${this.text}`;
  }

  constructor(
    private text: string,
    public metadata: Metadata = new Map(),
  ) {}

  static hash = memoize(
    (content: string, metadata: Metadata) => {
      return createHash("sha256")
        .update(content)
        .update(JSON.stringify([...metadata.entries()]))
        .digest("hex");
    },
    { cacheKey: JSON.stringify },
  );
}
```

himself65:
See https://github.com/run-llama/LlamaIndexTS/pull/1065

PR #846, opened 2024-05-16
fix: allow custom task query for agents

thucpn:


packages/core/src/agent/base.ts
```diff
@@ -287,6 +287,12 @@ export abstract class AgentRunner<
     return task.context.toolCallCount < MAX_TOOL_CALLS;
   }
 
+  protected computedMessageQuery = (
+    message: MessageContent,
+  ): MessageContent => {
+    return extractText(message);
+  };
```
himself65, line 294: This is not necessary since each LLM already does `extractText` if not supported

marcusschiesser (approved):
@himself65 LGTM - this is fixing that we can't send messages with images to agents (this use case is not working with OpenAIAgent: https://x.com/MarcusSchiesser/status/1790731811741409302)

himself65 (approved):

PR #849, opened 2024-05-16
chore: use `unpdf`

himself65:
move `pdfjs` back, but use universal build to make sure it can be run under Cloudflare


himself65:
I remove `node-canvas` deps, seems like we don't need this for non-image

packages/core/src/readers/PDFReader.ts
```diff
@@ -1,33 +1,35 @@
...
 import { FileReader } from "./type.js";
 
 /**
  * Read the text of a PDF
  */
 export class PDFReader extends FileReader {
-  async loadDataAsContent(fileContent: Buffer): Promise<Document[]> {
-    const pages = await readPDF(fileContent);
-    return pages.map((text, page) => {
+  async loadData(file: string): Promise<Document[]> {
```
marcusschiesser, line 9: @himself65 don't override `loadData` from `FileReader` - that way you're losing the unified metadata. Better make `loadData` `final` in `FileReader`?

Also `FileReader` is using `loadDataAsContent(content: Buffer)` instead of `loadDataAsContent(content: Uint8Array)` - how do you suggest dealing with this?

PR #855, opened 2024-05-17
feat: allow adding system prompt to chat engine

thucpn:


packages/core/src/engines/chat/ContextChatEngine.ts
```diff
@@ -71,7 +73,7 @@ export class ContextChatEngine extends PromptMixin implements ChatEngine {
       message,
       chatHistory,
     );
-
+    this.addSystemPromptToMessages(requestMessages.messages);
```
marcusschiesser, line 76: 1. i think we can just add the system prompt in this class, no need to change Mixin
2. check the behavior with python, but I think we have to always set the system prompt as first system prompt (even if already one is specified)

marcusschiesser: @thucpn about 2: great catch! didn't know that the system prompt is just prepended. Please test my commit, I passed the whole message to the requestMessage function - that's something we need to do as it ensures that we don't exceed the context length of the LLM.

PR #866, opened 2024-05-21
Fix/Agent llm initialization

HenryHengZJ:
Bug: When initializing agent with `llm` like:
```
new OpenAIAgent({
  tools,
  llm: new OpenAI({ apiKey: 'abc', temperature: 0.6 }),  // custom openai config
  chatHistory,
})
```
Error `Set OpenAI Key in OPENAI_API_KEY env variable` is thrown.

Fix: Syntax for 
```
params.llm ?? Settings.llm instanceof OpenAI ? (Settings.llm as OpenAI) : new OpenAI()
```
needs additional bracket like:
```
params.llm ?? (Settings.llm instanceof OpenAI ? (Settings.llm as OpenAI) : new OpenAI())
```
otherwise everytime `new OpenAI()` will be executed, and if no env variable is set, error is thrown

### DIFF
--- a/packages/core/src/agent/anthropic.ts
+++ b/packages/core/src/agent/anthropic.ts
@@ -39,9 +39,9 @@ export class AnthropicAgent extends AgentRunner<Anthropic> {
   constructor(params: AnthropicAgentParams) {
     super({
       llm:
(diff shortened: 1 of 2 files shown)


yisding (approved):
Thank you for the fix!

PR #868, opened 2024-05-22
fix: setDocumentHash should be async

pserrer1:
Even if used as an async function in multiple places the current implementation of setDocumentHash isn't declared that way.

This PR fixes this issue

### DIFF
--- a/.changeset/twelve-bottles-wait.md
+++ b/.changeset/twelve-bottles-wait.md
@@ -1,0 +1,6 @@
+---
+"llamaindex": minor
+"docs": minor
+---
+
+setDocumentHash should be async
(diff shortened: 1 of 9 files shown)


himself65 (commented):
What’s the use case of this

pserrer1:
> What’s the use case of this

1. This is the only function of the BaseDocumentStore which isn't async
2. Having a remote and not some in memory document store , this operation will always be async
3. Looking at the KVDocumentStore - even there the implementation of setDocumentHash is async and therefore should be awaited

himself65 (approved):

.changeset/twelve-bottles-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+"llamaindex": minor
+"docs": minor
```
marcusschiesser, line 3: @himself65 i wouldn't use minor

himself65: didn't see that, let me change

Issue #871, opened 2024-05-22
MongoDBAtlasVectorSearch delete only deletes one item


pserrer1:
Currently the MongoDBAtlasVectorSearch.delete function only deletes one item. This seems wrong for me as there could be multiple items having the same `ref_doc_id` i.e. if the ingestion pipeline splits a node into multiple ones. Other stores have the same issue - or even a totally different behaviour (Astra uses the node.id_ instead of the ref for deletion).

I'm now kind of unsure if I have a wrong understanding of how this should work (only store documents without ingestion..?) or if the delete logic is wrong across all vector stores.

This is how I would change the MongoDBAtlasVectorStore:

```
  async delete(refDocId: string, deleteOptions?: any): Promise<void> {
    // changed from deleteOne to deleteMany
    await this.collection.deleteMany(
      {
        [`${this.metadataKey}.ref_doc_id`]: refDocId,
      },
      deleteOptions,
    );
  }
```

justinmann:
This should fix it: https://github.com/run-llama/LlamaIndexTS/pull/887

AVtheking:
if this is fixed we can close this issue @marcusschiesser ?

PR #872, opened 2024-05-22
fix: mongodb atlas delete function to delete all ref doc from vector …

pserrer1:
…store

See https://github.com/run-llama/LlamaIndexTS/issues/871


marcusschiesser:
already merged

PR #886, opened 2024-05-24
feat: use images in context chat engine

marcusschiesser:
- adds support to use retrieved images in context chat engine
- adds example using it `examples/multimodal/context.ts`
- add streaming to multi modal response synthesizer

### DIFF
--- a/packages/core/src/engines/chat/DefaultContextGenerator.ts
+++ b/packages/core/src/engines/chat/DefaultContextGenerator.ts
@@ -1,10 +1,11 @@
-import type { NodeWithScore, TextNode } from "../../Node.js";
+import { type NodeWithScore } from "../../Node.js";
 import type { ContextSystemPrompt } from "../../Prompt.js";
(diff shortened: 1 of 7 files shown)


packages/core/src/engines/chat/DefaultContextGenerator.ts
```diff
@@ -68,15 +76,66 @@ export class DefaultContextGenerator
...
+    }
+
     return {
       message: {
-        content: this.contextSystemPrompt({
-          context: nodes.map((r) => (r.node as TextNode).text).join("\n\n"),
-        }),
-        role: "system",
+        content,
+        role: "user", // TODO: images in "system" messages don't work
```
marcusschiesser, line 98: @logan-markewich @yisding system messages can't have images in OpenAI (see https://platform.openai.com/docs/api-reference/chat/create)
This means we can't add retrieved images to the context.
However the example (`examples/multimodal/context.ts`) works nicely with generating a user message as context.
Any ideas how to handle this?

PR #891, opened 2024-05-28
fix: incorrect hash because of missing params in decorator

pserrer1:
The current `chunkSizeCheck` decorator don't pass the params to the getContent function and therefore always calculates the node hash without using the metadata information.

This PR fixes this bug


yisding (approved):
Thanks!

.changeset/early-spoons-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"llamaindex": minor
```
marcusschiesser, line 2: @yisding changed this to patch (i assume we don't want a minor release for this change)

PR #893, opened 2024-05-29
feat: add numCandidates setting to MongoDBAtlasVectorStore for tuning queries

pserrer1:
When requesting the MongoDB Atlas store, the result could be inaccurate - especially when only requesting a small amount of items. To tune the result, this PR adds the option to set the `numCandiates` vectorSearch option

### DIFF
--- a/.changeset/angry-eagles-dress.md
+++ b/.changeset/angry-eagles-dress.md
@@ -1,5 +1,5 @@
 ---
-"llamaindex": minor
+"llamaindex": patch
 ---
 
 feat: add numCandidates setting to MongoDBAtlasVectorStore for tuning queries
--- a/.changeset/angry-eagles-dress.md
+++ b/.changeset/angry-eagles-dress.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": minor
+---
+
+feat: add numCandidates setting to MongoDBAtlasVectorStore for tuning queries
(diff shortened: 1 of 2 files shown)


.changeset/angry-eagles-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"llamaindex": minor
```
marcusschiesser, line 2:
```suggestion
"llamaindex": patch
```

packages/core/src/storage/vectorStore/MongoDBAtlasVectorStore.ts
```diff
@@ -119,7 +121,7 @@ export class MongoDBAtlasVectorSearch
     const params: any = {
       queryVector: query.queryEmbedding,
       path: this.embeddingKey,
-      numCandidates: query.similarityTopK * 10,
+      numCandidates: this.numCandidates ?? query.similarityTopK * 10,
```
marcusschiesser, line 124: why is using `query.similarityTopK` not sufficient? 
(I also wonder why it is `*10` here...)

pserrer1: MongoDB Atlas using ANN (https://arxiv.org/abs/1603.09320) for the vector search. If you have a small number for the `similarityTopK`, the candiates they use for the lookup might be to small to include all relevant chunks.

marcusschiesser: @pserrer1 got it. thanks a lot for the context. I see. So the upper bound `* 10` is incorrect for your use case

Here's the documentation:
image

They recommend a max of 20, does this work for you:

```
limit: query.similarityTopK,
numCandidates: query.similarityTopK * 20
```

pserrer1: Not sure if we find a solution that fits all. I updated the file with js docs

marcusschiesser (approved):

PR #901, opened 2024-06-04
fix: calling tools with large inputs

marcusschiesser:
Problem: A partial JSON is returned by the LLM of the agent if the `maxTokens` parameter is too small (Tested with OpenAI's GPT4-Turbo). 

Changes:
- Show an error message if an agent tool is called with a partial JSON. Hint the user to increase `maxTokens`
- Don't stop agent execution

### DIFF
--- a/packages/core/src/agent/utils.ts
+++ b/packages/core/src/agent/utils.ts
@@ -16,10 +16,25 @@ export async function callTool(
   toolCall: ToolCall | PartialToolCall,
   logger: Logger,
 ): Promise<ToolOutput> {
-  const input: JSONObject =
-    typeof toolCall.input === "string"
-      ? JSON.parse(toolCall.input)
-      : toolCall.input;
+  let input: JSONObject;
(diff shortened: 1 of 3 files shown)


packages/core/src/agent/utils.ts
```diff
@@ -16,10 +16,25 @@ export async function callTool(
...
+  let input: JSONObject;
+  if (typeof toolCall.input === "string") {
+    try {
+      input = JSON.parse(toolCall.input);
+    } catch (e) {
+      const output = `Tool ${toolCall.name} can't be called. Input is not a valid JSON object.`;
+      logger.error(
+        `${output} Try increasing the maxTokens parameter of your LLM. Invalid Input: ${toolCall.input}`,
+      );
+      return {
```
marcusschiesser, line 28: @himself65 Makes sense to me not to throw the error (as it would stop the whole agent) and instead return `isError: true`. But in my tests the agent continuously tries then to call the tool again. Another issue?

PR #904, opened 2024-06-05
fix: move paragraph separation back to two newlines

yisding:


packages/core/src/TextSplitter.ts
```diff
@@ -55,10 +53,9 @@ export function cjkSentenceTokenizer(sentence: string): string[] {
   return slist.filter((s) => s.length > 0);
 }
 
-export const defaultParagraphSeparator = EOL + EOL + EOL;
-
-// In theory there's also Mac style \r only, but it's pre-OSX and I don't think
-// many documents will use it.
+// NOTE this is different from Python because in Python a lot of the heavy lifting for paragraph separation
+// is already handled by NLTK, but we don't have that option here.
+export const defaultParagraphSeparator = /\n\n|\r\n\r\n|\r\r/g;
```
yisding, line 58: This way we decouple the document encoding from the OS that the user is currently running on.

himself65 (approved):

packages/core/src/TextSplitter.ts
```diff
@@ -145,7 +142,11 @@ export class SentenceSplitter {
         paragraphSplits[idx] = [
           paragraphSplits[idx],
           paragraphSplits[idx + 1],
-        ].join(this.paragraphSeparator);
+        ].join(
+          typeof this.paragraphSeparator === "string"
+            ? this.paragraphSeparator
+            : "\n\n",
```
marcusschiesser, line 148: `\n\n` works for all envs?

packages/core/src/TextSplitter.ts
```diff
@@ -55,10 +53,9 @@ export function cjkSentenceTokenizer(sentence: string): string[] {
   return slist.filter((s) => s.length > 0);
 }
 
-export const defaultParagraphSeparator = EOL + EOL + EOL;
```
marcusschiesser, line 58: wasn't the problem that we're not using `export const defaultParagraphSeparator = EOL + EOL`?

PR #883, opened 2024-05-24
feat: add parallel processing to SimpleDirectoryReader

KindOfAScam:
#830 Messed up the rebase, so here we go again, hopefully functional now.

PR #908, opened 2024-06-05
feat: add parallel processing to SimpleDirectoryReader

KindOfAScam:
@marcusschiesser as requested. #883 
The reader examples work fine, parallel processing in the `llamaparse-dir` example also works as expected, so not sure what's causing `ERR_INVALID_URL_SCHEME` with the other examples. 

Looking at the error message, it might be related to filePath being pushed to a queue, therefore not being as freely available as before the PR to other use cases? Looking into it


marcusschiesser:
@KindOfAScam Thanks for re-opening. The problem is that p-limit is only working with ESM see https://github.com/sindresorhus/p-limit/releases/tag/v5.0.0 - let me have a look

packages/env/src/index.ts
```diff
@@ -35,6 +36,11 @@ export function createSHA256(): SHA256 {
   };
 }
 
+export async function pLimit(numWorkers: number): Promise<LimitFunction> {
+  const { default: pLimit } = await import("p-limit");
```
marcusschiesser, line 40: @KindOfAScam the await import makes it work with CJS

marcusschiesser:
Thanks @KindOfAScam I fixed it and merged it again

KindOfAScam:
awesome thanks! @marcusschiesser 
The way you used utils.ts is similar to python llama-index [run_jobs](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/async_utils.py#L114) 

Possible suggestion: Make it a global helper function to be used throughout for any parallel jobs like in the python version?

marcusschiesser:
@KindOfAScam yes, that's definitely something we need to do, I also think the `pLimit` function can be simplified for our use case a lot. I just made it work for now. We're happy for another PR :D

PR #909, opened 2024-06-05
feat: chat engine based on agents

parhammmm:
@marcusschiesser this one is the other engine we talked about. 

Includes a fix for chat history in agents

```ts
  const agent = new OpenAIAgent({
    llm,
    tools: [...],
    systemPrompt: "..."
  });
  const engine = new AgentChatEngine({ agent });
  const result = await engine.chat({
    message: [
      {
        type: "text",
        text: "hi",
      },
    ],
    stream: false,
  });

```

### DIFF
--- a/packages/core/src/agent/base.ts
+++ b/packages/core/src/agent/base.ts
@@ -1,4 +1,5 @@
 import { ReadableStream, TransformStream, randomUUID } from "@llamaindex/env";
+import { ChatHistory } from "../ChatHistory.js";
 import { Settings } from "../Settings.js";
 import {
   type ChatEngine,
@@ -285,8 +286,9 @@ export abstract class AgentRunner<
     message: MessageContent,
     stream: boolean = false,
     verbose: boolean | undefined = undefined,
+    chatHistory?: ChatMessage<AdditionalMessageOptions>[],
   ) {
-    const initialMessages = [...this.#chatHistory];
+    const initialMessages = [...(chatHistory ?? this.#chatHistory)];
     if (this.#systemPrompt !== null) {
       const systemPrompt = this.#systemPrompt;
       const alreadyHasSystemPrompt = initialMessages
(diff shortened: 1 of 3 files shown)


himself65 (approved):

marcusschiesser:
as `AgentRunner` already implements `ChatEngine` every agent already is a chat engine. Why do we need a new chat engine then? If we need the chat history, how about adding it to the agent implementation?

parhammmm:
> as `AgentRunner` already implements `ChatEngine` every agent already is a chat engine. Why do we need a new chat engine then? If we need the chat history, how about adding it to the agent implementation?

@marcusschiesser it felt safer, one reason was the output streams of ChatEngines and Agents aren't compatible.

I also couldn't traverse exactly what happens the Agents seem to keep a history for internal usage, would help to clarify what the history on the Agent does

marcusschiesser:
@parhammmm @himself65 i guess we should then better unify the output reasponse/streams of ChatEngines and Agents?

parhammmm:
@marcusschiesser I'm not confident I can do it myself safely but I can take a stab at it and sync with @himself65 for review.

Would it be possible in the meantime to release `AgentChatEngine` in the `community` package, if it is I can move it and then take the above

parhammmm:
@marcusschiesser what do you think?

parhammmm:
@marcusschiesser thanks for taking a look. part of the fix for 2 is actually in this PR and can be used to match the other engines here: https://github.com/run-llama/LlamaIndexTS/pull/909/files#diff-e722a872695a5840308e8f2f844e6a3d6f25969880fe3e670ae5797ac38a9724R289

What do you think of doing the two patches in two different PRs? The history bug is the most important and it would help unlock making a gpt style chat, which is what I'm actually using `AgentChatEngine` for.

Thanks again.

marcusschiesser:
@parhammmm great, yes 1 & 2 should be one PR each. If you're blocked by 2, would you like to send a PR for it by changing the `AgentRunner` directly?

parhammmm:
@marcusschiesser sure thing will get to it shortly

parhammmm:
@marcusschiesser @himself65 I'm not deep enough into LITS to be able to give good feed back on 930, but I pushed the history fix to this PR https://github.com/run-llama/LlamaIndexTS/pull/933

And as a work around for the agent's responses I'm doing the following on our side, which translates to adding a utility that handles the merging of the two types (just incase it's helpful)

```ts
import { Response } from "llamaindex";
import { ReadableStream } from "node:stream/web";

type LlamaindexAgentResponse = { response: { delta: string } };
type LlamaindexResponse = Response | LlamaindexAgentResponse;
...
const response: AsyncIterable<Response> | ReadableStream<LlamaindexAgentResponse>;
...
const result =
  typeof value.response === "string"
      ? value.response
      : value.response.delta;
```

parhammmm:
closing this in favour of the https://github.com/run-llama/LlamaIndexTS/pull/933 and https://github.com/run-llama/LlamaIndexTS/pull/930

PR #910, opened 2024-06-05
feat: add json mode and image retrieval to LlamaParse

KindOfAScam:
Testing looks good, pure json results are as expected. 
I'm a bit unsure about how to further use the json results tho.

Is there a function akin to this python [example](https://github.com/run-llama/llama_parse/blob/main/examples/demo_json.ipynb): 

Especially, is there an existing way to use this part:

```
json_objs = parser.get_json_result("./uber_10q_march_2022.pdf")
json_list = json_objs[0]["pages"]

```
In LlamaIndexTS? 

My idea would be to use` jsonToNode` afterwards to turn the results into proper Nodes for further processing. Or am I missing something?

Also restructured most of LlamaParseReader to align more closely with the python version. This modular approach should avoid redundant code.

### DIFF
--- a/packages/core/src/readers/LlamaParseReader.ts
+++ b/packages/core/src/readers/LlamaParseReader.ts
@@ -150,8 +150,8 @@ export class LlamaParseReader implements FileReader {
     }
   }
 
-  async loadData(file: string): Promise<Document[]> {
-    const metadata = { file_path: file };
+  // Create a job for the LlamaParse API
+  private async createJob(file: string): Promise<string> {
 
     // Load data, set the mime type
     const data = await fs.readFile(file);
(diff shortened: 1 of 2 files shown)


himself65 (approved):

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -188,48 +187,170 @@ export class LlamaParseReader implements FileReader {
...
-      ];
+    // Creates a job for the file
+    const jobId = await this.createJob(file);
+    if (this.verbose) {
+      console.log(`Started parsing the file under job id ${jobId}`);
+    }
+
+    // Return results as Document objects
+    const resultJson = await this.getJobResult(jobId, this.resultType);
+    return [
```
himself65, line 269: Should this return a list of documents by page?

KindOfAScam: @himself65 I don't think so, as far as I understand in LlamaParse only json mode relies on pages and even than it seems mostly for the image retrieval functionality. Documentation is a bit spare and so is my knowledge of python, so take it with a grain of salt and maybe check it yourself. But from my understanding. I think it's fine to return the full documents and split them up later instead of listing them by page

PR #914, opened 2024-06-07
fix: LlamaParse json mode returns array + basic example

KindOfAScam:
Just a small fix to align with python structure + very basic example to show the output structure.

The uber report is from python version, so I assume licensing is fine? brk-2022.pdf or any other file already existing did not really return any interesting data to highlight the structure.


examples/readers/src/llamaparse-json.ts
```diff
@@ -0,0 +1,22 @@
+import * as fs from "fs";
```
himself65, line 1: use `node:fs/promises`

himself65 (changes requested):

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -155,9 +155,12 @@ export class LlamaParseReader extends FileReader {
   private async createJob(data: Buffer): Promise<string> {
     // Load data, set the mime type
     const mimeType = await this.getMimeType(data);
+    const matchingExtension = Object.keys(SupportedFiles).filter(
```
marcusschiesser, line 158: should we start the job if the extension is not matching?

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -270,25 +273,25 @@ export class LlamaParseReader extends FileReader {
...
       console.log(`Started parsing the file under job id ${jobId}`);
     }
 
-    // Return results as JSON object
+    // Return results as an array of JSON objects
     const resultJson = await this.getJobResult(jobId, "json");
     resultJson.job_id = jobId;
     resultJson.file_path = file;
-    return resultJson;
+    return [resultJson];
```
marcusschiesser, line 294: what's the advantage of wrapping the JSON in an array?

examples/readers/src/llamaparse-json.ts
```diff
@@ -0,0 +1,30 @@
+import fs from "fs/promises";
+import { LlamaParseReader } from "llamaindex";
+
+async function main() {
+  // Load PDF using LlamaParse json mode
+  const reader = new LlamaParseReader({ resultType: "json" });
+  const jsonObjs = await reader.loadJson("../data/uber_10q_march_2022.pdf");
```
marcusschiesser, line 7: can we use an existing PDF?

packages/core/src/readers/LlamaParseReader.ts
KindOfAScam: This is just supposed to enhance readability by logging the file extensions instead of the mime type.
The getMimeType method should throw an error beforehand anyway if the file is not supported, so a non matching extensions, should have no influence on the job itself

packages/core/src/readers/LlamaParseReader.ts
KindOfAScam: Mostly consistency with the python version (I wasted a lot of time trying to figure out why I couldn't get the python examples to work, only to realize my output was not wrapped in an array lol)

examples/readers/src/llamaparse-json.ts
KindOfAScam: the file is straight from python llama_index, so I hoped licensing would be fine.
We can use an existing pdf, it's just that none of them have an interesting underlying structure with images, tables, etc.. to highlight LlamaParse json mode capabilities.

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -270,25 +273,25 @@ export class LlamaParseReader extends FileReader {
...
+  async loadJson(file: string): Promise<Record<string, any>[]> {
     const data = await fs.readFile(file);
     // Creates a job for the file
     const jobId = await this.createJob(data);
     if (this.verbose) {
       console.log(`Started parsing the file under job id ${jobId}`);
     }
 
-    // Return results as JSON object
+    // Return results as an array of JSON objects
```
marcusschiesser, line 290:
```suggestion
    // Return results as an array of JSON objects (same format as Python version of the reader)
```

packages/core/src/readers/LlamaParseReader.ts
marcusschiesser: I see, alternatively you could return the extension as well from `getMimeType`?

KindOfAScam: That works and actually fixes a potential mismatch error between most likely and assigned mime type. Thanks!

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -350,18 +350,19 @@ export class LlamaParseReader extends FileReader {
...
-  private async getMimeType(data: Buffer): Promise<string> {
-    const mimes = filetypemime(data);
-    const validMime = mimes.find((mime) =>
-      Object.values(SupportedFiles).includes(mime),
-    );
-    if (!validMime) {
+  private async getMimeType(
+    data: Buffer,
+  ): Promise<{ mimeType: string; extension: string }> {
+    const mimes = filetypemime(data); // Get an array of possible MIME types
```
marcusschiesser, line 356: @KindOfAScam don't you skip now the other valid mime types?

KindOfAScam: Yes but on purpose.
 
Before it was iterating through the list of potential mime types and picking the first one that matches a supported extension. This could result in an unsupported mime type being assigned just because a less likely mime type is supported.

Now it's choosing the most likely mime type and checks if it fits a  supported extension. That's also equivalent to python version.

marcusschiesser (approved):

marcusschiesser:
@himself65 please also approve and merge

himself65 (approved):

PR #915, opened 2024-06-09
feat: add new options to LlamaParseReader

KindOfAScam:
also turned the optional params into actual optional params and added a helper function to avoid sending not needed and undefined values


packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -162,14 +170,24 @@ export class LlamaParseReader extends FileReader {
...
+      fast_mode: this.fastMode?.toString(),
+      do_not_unroll_columns: this.doNotUnrollColumns?.toString(),
+      page_seperator: this.pageSeperator,
+      gpt4o_mode: this.gpt4oMode?.toString(),
+      gpt4o_api_key: this.gpt4oApiKey,
+    };
+
+    // Appends body with any defined LlamaParseBodyParams
+    Object.entries(LlamaParseBodyParams).forEach(([key, value]) => {
+      this.appendBody(body, key, value);
```
marcusschiesser, line 189: we not just inline?
```suggestion
       if (value) {
       body.append(key, value);
    }
```

marcusschiesser (approved):

PR #918, opened 2024-06-10
feat: truncate embedding tokens

marcusschiesser:
- truncate text to embed before sending it to OpenAI


packages/core/src/embeddings/OpenAIEmbedding.ts
```diff
@@ -109,6 +123,16 @@ export class OpenAIEmbedding extends BaseEmbedding {
    * @param options
    */
   private async getOpenAIEmbedding(input: string[]): Promise<number[][]> {
```
himself65, line 125: this is not a good solution.

This will break the return type of `getTextEmbedding('large chunk')`

himself65: https://github.com/run-llama/LlamaIndexTS/pull/843#issue-2299093676

packages/core/src/GlobalsHelper.ts
```diff
@@ -1,4 +1,6 @@
-import { encodingForModel } from "js-tiktoken";
+// Note: don't use js-tiktoken it's 60x slower than the WASM implementation
```
marcusschiesser, line 1: @himself65 tiktoken supports edge and cloudflare, see https://www.npmjs.com/package/tiktoken

packages/core/src/embeddings/OpenAIEmbedding.ts
marcusschiesser: doesn't `getTextEmbedding('large chunk')` - it just ensures that `getTextEmbedding('large chunk')` is not throwing an exception

himself65:
let me ignore waku for now, it might be their issue

himself65 (approved):

PR #919, opened 2024-06-10
fix: polyfill for cloudflare worker

himself65:
workaround for Cloudflare worker build


packages/core/src/storage/StorageContext.ts
```diff
@@ -44,7 +44,7 @@ export async function storageContextFromDefaults({
     }
     if (storeImages && !(ModalityType.IMAGE in vectorStores)) {
       vectorStores[ModalityType.IMAGE] = new SimpleVectorStore({
-        embedModel: new ClipEmbedding(),
+        embedModel: new (await getImageEmbedModel())(),
```
marcusschiesser, line 47: @himself65 

i think you can keep `new ClipEmbedding` here and inside of `CllipEmbedding` don't do the dynamic transformers.js import for cloudflare instead

Advantages:
1. Users can still do `new ClipEmbedding` inside of a cloudflare worker at the moment which would fail
2. the selection of the embedding model should not depend on the environment (that's what `getImageEmbedModel` does at the moment)

PR #924, opened 2024-06-12
feat: add split by page mode to LlamaParseReader

KindOfAScam:
Translates and uses this new python method:

```
def _get_sub_docs(docs: List[Document]) -> List[Document]:
    """Split docs into pages, by separator."""
    sub_docs = []
    for doc in docs:
        doc_chunks = doc.text.split("\n---\n")
        for doc_chunk in doc_chunks:
            sub_doc = Document(
                text=doc_chunk,
                metadata=deepcopy(doc.metadata),
            )
            sub_docs.append(sub_doc)

    return sub_docs
```

Not really sure about the relationship of `pageSeperator` (gets sent to the API) and the new split by page mode atm

### DIFF
--- a/.changeset/chilled-tomatoes-visit.md
+++ b/.changeset/chilled-tomatoes-visit.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+feat: add splitByPage mode to LlamaParseReader
--- a/apps/docs/docs/modules/data_loaders/llama_parse/index.mdx
+++ b/apps/docs/docs/modules/data_loaders/llama_parse/index.mdx
@@ -48,6 +48,7 @@ They can be divided into two groups.
 - `gpt4oApiKey?` Deprecated. Use vendorMultimodal params. Optional. Set the GPT-4o API key. Lowers the cost of parsing by using your own API key. Your OpenAI account will be charged. Can also be set in the environment variable `LLAMA_CLOUD_GPT4O_API_KEY`.
 - `boundingBox?` Optional. Specify an area of the document to parse. Expects the bounding box margins as a string in clockwise order, e.g. `boundingBox = "0.1,0,0,0"` to not parse the top 10% of the document.
 - `targetPages?` Optional. Specify which pages to parse by specifying them as a comma-separated list. First page is `0`.
+- `splitByPage` Wether to split the results, creating one document per page. Uses the set `pageSeparator` or `\n---\n` as fallback. Default is true.
 - `useVendorMultimodalModel` set to true to use a multimodal model. Default is `false`.
 - `vendorMultimodalModel?` Optional. Specify which multimodal model to use. Default is GPT4o. See [here](https://docs.cloud.llamaindex.ai/llamaparse/features/multimodal) for a list of available models and cost.
 - `vendorMultimodalApiKey?` Optional. Set the multimodal model API key. Can also be set in the environment variable `LLAMA_CLOUD_VENDOR_MULTIMODAL_API_KEY`.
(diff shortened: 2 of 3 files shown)


packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -283,11 +287,12 @@ export class LlamaParseReader extends FileReader {
 
     // Return results as Document objects
     const resultJson = await this.getJobResult(jobId, this.resultType);
-    return [
+    const docs = [
       new Document({
         text: resultJson[this.resultType],
```
marcusschiesser, line 292: better split the text first and then create documents

packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -130,14 +130,18 @@ export class LlamaParseReader extends FileReader {
...
   // The page separator to use to split the text. Default is None, which means the parser will use the default separator '\\n---\\n'.
   pageSeperator?: string;
   // Whether to use gpt-4o to extract text from documents.
   gpt4oMode: boolean = false;
   // The API key for the GPT-4o API. Optional, lowers the cost of parsing. Can be set as an env variable: LLAMA_CLOUD_GPT4O_API_KEY.
   gpt4oApiKey?: string;
+  // Whether to split by page. Default separator: `\n---\n` Can be overwritten using splitByPageSeperator
+  splitByPage: boolean = true;
+  // The page separator used when splitByPage is true. Does not exist in Python version. Not actually sure what pageSeperator is for. pageSeperator gets sent to the LlamaParse API, while splitByPageSeperator is used in the getSubDocs function.
+  splitByPageSeperator: string = "\n---\n";
```
marcusschiesser, line 144: we should just use whatever page seperator that LlamaParse is using - isn't that specified by the `pageSeperator` parameter? @hexapode can you help out here?

KindOfAScam: It would be nice to get some clarification, `pageSeperator` is a param that gets send to the API Backend and is described as: 

> The page separator to use to split the text. Default is None, which means the parser will use the default separator '\\n---\\n'.

while `splitByPage` is an internal function described as:

> Whether to split by page (NOTE: using a predefined separator `\n---\n`)

 `pageSeperator` is not responsible for the seperator in `splitByPage`, so I have no clue what it actually does on the backend side. I'll draft this for now

marcusschiesser: According to @hexapode, the JSON of the job result (https://docs.cloud.llamaindex.ai/llamaparse/API/get-job-json-result-api-parsing-job-job-id-result-json-get) already returns a collection of pages, so seems to me that we should handle that first

KindOfAScam: That's already handled to the same extent as in python.
`loadJson` returns an array of json Objects containing `pages`, which represent a single file parsed. Each object under `pages` represent a single page and has a `page` field, next to `text`, `md`, etc.

`splitByPage` is only used in `loadData` method, so only for markdown or text and still not sure what about `pageSeperator`

BinaryBrain: Related: https://github.com/run-llama/LlamaIndexTS/pull/1023

KindOfAScam: Oh.. thank you, apparently, I don't know how to spell `separator` lol.

KindOfAScam:
I've rebased this branch, applied your suggestion and aligned usage of `pageSeparator` with Python's version.

Should be ready to merge now if you could have a look again @marcusschiesser thanks!

himself65 (approved):

marcusschiesser:
@KindOfAScam thanks looking good, can you please add a changeset too?

KindOfAScam:
@marcusschiesser added one

marcusschiesser (approved):
thanks @KindOfAScam

PR #928, opened 2024-06-12
feat: add api key methods, usage retrieval and ignoreErrors flag to LlamaParseReader

KindOfAScam:
Use cases:

> automation of api key renewal

> usage retrieval can be used to limit spending, as llamaparse has no spending limit that can be set.

`max_pdf_pages` is currently always responding with `1000`, so backend has no implementation for unlimited max_pdf_pages, but that's just cosmetical really

### DIFF
--- a/packages/core/src/readers/LlamaParseReader.ts
+++ b/packages/core/src/readers/LlamaParseReader.ts
@@ -1,6 +1,7 @@
 import { fs, getEnv } from "@llamaindex/env";
 import { filetypemime } from "magic-bytes.js";
 import { Document } from "../Node.js";
+import type { UUID } from "../types.js";
 import { FileReader, type Language, type ResultType } from "./type.js";
 
 const SupportedFiles: { [key: string]: string } = {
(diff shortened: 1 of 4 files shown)


packages/core/src/readers/LlamaParseReader.ts
```diff
@@ -325,49 +368,250 @@ export class LlamaParseReader extends FileReader {
...
+  }
+
+  /**
+   * A function to generate an API key.
+   *
+   * @param {string} name - The name for the API key. Must be less than 150 characters.
+   * @param {string} [session] - Optional session information.
+   * @return {Promise<LlamaParseApiResponse>} A promise that resolves to the generated API key response.
+   */
+  async generateApiKey(
```
marcusschiesser, line 434: I wouldn't add these wrappers to the LlamaParse API here. 
Better to use https://github.com/fern-api/fern to automatically generate them.

`@llamaindex/cloud` is our package that contains automatically generated API wrappers to the LlamaCloud.

@EmanuelCampos can we add LlamaParse there too?

KindOfAScam: Oh that's cool, in that case I'll leave it up to you, but would be nice to add them there. I'll split out the parts that are not a new wrapper into a new PR and close this one. And sorry if my PR's are a bit meh sometimes.. I'm still learning a lot xD. Thanks for always reviewing!

PR #929, opened 2024-06-13
feat: added support for embeddings via HuggingFace Inference API

tasmay:
Added support for generating embeddings via the HuggingFace Inference API. The same exists in the Python version but missing in LlamaIndexTS.


packages/core/src/embeddings/HuggingFaceEmbedding.ts
```diff
@@ -46,3 +47,64 @@ export class HuggingFaceEmbedding extends BaseEmbedding {
     return Array.from(output.data);
   }
 }
+
+// TODO workaround issue with @huggingface/inference@2.7.0
```
marcusschiesser, line 51: maybe move this to `HuggingFaceInferenceAPIEmbedding`? as it's very different to the inference via transformers.js?

marcusschiesser, line 51: is there still a TODO?

tasmay: @huggingface/inference no longer exports Options, if they fix it in later versions, we can switch to using it.

himself65: link to an issue?

tasmay: I am not sure if there is an open issue regarding this. I kept the implementation similar to [llm/huggingface.ts](https://github.com/run-llama/LlamaIndexTS/blob/01658fdb31efdf705519b6072f689182c3fdc87f/packages/llamaindex/src/llm/huggingface.ts#L20) @huggingface/inference has not exported Options in their latest version. If I do:

```
import { type Options as HfInferenceOptions } from "@huggingface/inference"
```
gives an error: Module '"@huggingface/inference"' has no exported member 'Options'. Hence, this workaround.

himself65: Oh I see, you can try this to infer type from it
 
```ts
type Option = ConstructorParameters<...>[1]
```

marcusschiesser (approved):
Thanks @tasmay

marcusschiesser:
@tasmay looks good, please resolve conflicts, then we can merge it

tasmay:
Done. Thanks @marcusschiesser and @himself65

himself65 (approved):

PR #930, opened 2024-06-14
refactor: unify response and agent response

marcusschiesser:


packages/core/src/agent/base.ts
```diff
@@ -101,16 +97,6 @@ export function createTaskOutputStream<
   });
 }
 
-export type AgentStreamChatResponse<Options extends object> = {
-  response: ChatResponseChunk<Options>;
-  sources: ToolOutput[];
-};
-
-export type AgentChatResponse<Options extends object> = {
```
marcusschiesser, line 109: `AgentChatResponse` is better than `Response` but most people are using normal chat engines I assume, so we better improve the existing `Response` class (see above)

packages/core/src/Response.ts
```diff
@@ -1,16 +1,27 @@
 import type { NodeWithScore } from "./Node.js";
+import type { ChatMessage, ChatResponse } from "./llm/types.js";
 
-/**
- * Response is the output of a LLM
- */
-export class Response {
+export class Response implements ChatResponse {
```
marcusschiesser, line 4: Idea: we include `message` and `raw` from `ChatResponse` instead of adding `chatResponse: ChatResponse` as an attribute to `Response`

Note: we `ChatResponse` supports extension with `Options` we have to add this to this PR (after we agree on the contract)

packages/core/src/Response.ts
```diff
@@ -1,16 +1,27 @@
...
+  // @deprecated use 'message' instead
```
marcusschiesser, line 5: we tell the user to not use `response` instead migrate to `message`

packages/core/src/Response.ts
marcusschiesser: another advantage is that `Reponse` behaves like `ChatReponse`

marcusschiesser: TBD: how to deal with `chunk`s? We have `ChatReponseChunk`s as type returned by LLMs but chat engines are just streaming `Response` objects. 

Suggestion: we add an optional `delta` attribute to `Response`

himself65, line 4: could we don't naming `Response` it's already a Web API

marcusschiesser: how about `EngineResponse` (see new code)?

packages/core/e2e/node/claude.e2e.ts
```diff
@@ -71,12 +71,11 @@ await test("anthropic agent", async (t) => {
         },
       ],
     });
-    const { response, sources } = await agent.chat({
+    const response = await agent.chat({
       message: "What is the weather in San Francisco?",
     });
     consola.debug("response:", response.message.content);
 
-    strictEqual(sources.length, 1);
```
marcusschiesser, line 79: @himself65 i removed `sources` (the tooloutput) 1. the name is confusing 2. it can be retrieved with events

himself65 (approved):

PR #948, opened 2024-06-20
feat: add the model ids for Anthropic Claude 3.5 Sonnet model on Anthropic and Bedrock

petergoldstein:


himself65 (approved):
nice catch, thanks for update

PR #955, opened 2024-06-21
feat: tool calling for Bedrock's Claude and General LLM Agent

### LINKED ISSUE #939: feat: Add Anthropic Agent support via bedrock
Starting work on this, and adding this here for it's visibility, 

It will be an addition to `@llamaindex/community`

parhammmm:
A note: let me know what naming is preferred for the agent. I called it `LLMAgent` but `Agent` sounds good too I can change it if preferred. 

I reshuffled a bunch of things but that was only to make Agent development easier and faster since I'll be adding Gemini tool calling support next.

Technically `AnthropicAgent` and `OpenAIAgent` are not explicitly needed but I left them there since someone might be using it and it seems LITS current doesn't have streaming tool support for Calude Models via Anthropic which AnthropicAgent takes into account
```ts
  import { LLMAgent } from "llamaindex";

  const agent = new LLMAgent({
    llm, // can be Anthropic, Bedrock, OpenAI or any LITS LLM that has tool calling support added
    tools: [tool]
  });

  const response = await agent.chat({
    stream: false,
    message: "call the tool",
  });

```

closes #939 

@marcusschiesser @himself65


marcusschiesser:
@himself65 as your agent is involved, it's better your reviewing this

himself65 (approved):
LGTM

PR #960, opened 2024-06-22
Added support for Meta LLMs in AWS Bedrock

ksfreitas:


packages/community/src/llm/bedrock/base.ts
```diff
@@ -194,9 +197,59 @@ class AnthropicProvider extends Provider {
...
+    messages: T[],
+  ): InvokeModelCommandInput | InvokeModelWithResponseStreamCommandInput {
+    let promptFunction: (messages: ChatMessage[]) => string;
+
+    if (metadata.model.startsWith("meta.llama3")) {
+      promptFunction = mapChatMessagesToMetaLlama3Messages;
+    } else if (metadata.model.startsWith("meta.llama2")) {
+      promptFunction = mapChatMessagesToMetaLlama2Messages;
+    } else {
+      throw new Error(`Metal model ${metadata.model} is not supported`);
```
petergoldstein, line 232: I think "Metal" here is supposed to be "Meta"

ksfreitas: Ops... Typo fixed

marcusschiesser (commented):
@ksfreitas thanks a lot. looks great! does it make sense to move the anthropic stuff to an anthropic.ts file and your meta stuff to a mets.ts file (inside of the bedrock dir)?

packages/community/src/llm/bedrock/base.ts
```diff
@@ -194,9 +197,59 @@ class AnthropicProvider extends Provider {
...
+    const result = this.getResultFromResponse(response);
+    return result.generation;
+  }
+
+  getTextFromStreamResponse(response: Record<string, any>): string {
+    const event: StreamEvent | undefined = response.chunk?.bytes
+      ? JSON.parse(toUtf8(response.chunk?.bytes))
+      : undefined;
+
+    if (event?.type === "content_block_delta") return event.delta.text;
```
parhammmm, line 217: this is Anthropic specific and can be removed

parhammmm, line 217: This is Anthropic specific the chunk from Meta looks like:

```js
{
  generation: ' hi',
  prompt_token_count: null,
  generation_token_count: 19,
  stop_reason: null
}
```

so here it should be:

`event.generation`, `StreamEvent` name could be updated to `AnthropicStreamEvent` and `MetaStreamEvent` can be created too

ksfreitas:
@parhammmm @marcusschiesser changes done, any more suggestions to improve (or fix) my code?

himself65 (approved):

parhammmm (changes requested):
All looking good, did some testing and there is one change that's needed for Meta, which is caused by the Tool calling features being added (after this PR was started - so not the fault of this PR)

in `base.ts` `nonStreamChat` should be:

```ts
  protected async nonStreamChat(
    params: BedrockChatParamsNonStreaming,
  ): Promise<BedrockChatNonStreamResponse> {
    const input = this.provider.getRequestBody(
      this.metadata,
      params.messages,
      params.tools,
      params.additionalChatOptions,
    );
    const command = new InvokeModelCommand(input);
    const response = await this.client.send(command);
    const tools = this.supportToolCall // Tool call support check
      ? this.provider.getToolsFromResponse(response)
      : [];
    const options: ToolCallLLMMessageOptions = tools.length
      ? { toolCall: tools }
      : {};
    return {
      raw: response,
      message: {
        role: "assistant",
        content: this.provider.getTextFromResponse(response),
        options,
      },
    };
  }
```

A warning log might also be helpful too

ksfreitas:
> All looking good, did some testing and there is one change that's needed for Meta, which is caused by the Tool calling features being added (after this PR was started - so not the fault of this PR)
> 
> in `base.ts` `nonStreamChat` should be:
> 
> ```ts
>   protected async nonStreamChat(
>     params: BedrockChatParamsNonStreaming,
>   ): Promise<BedrockChatNonStreamResponse> {
>     const input = this.provider.getRequestBody(
>       this.metadata,
>       params.messages,
>       params.tools,
>       params.additionalChatOptions,
>     );
>     const command = new InvokeModelCommand(input);
>     const response = await this.client.send(command);
>     const tools = this.supportToolCall // Tool call support check
>       ? this.provider.getToolsFromResponse(response)
>       : [];
>     const options: ToolCallLLMMessageOptions = tools.length
>       ? { toolCall: tools }
>       : {};
>     return {
>       raw: response,
>       message: {
>         role: "assistant",
>         content: this.provider.getTextFromResponse(response),
>         options,
>       },
>     };
>   }
> ```
> 
> A warning log might also be helpful too

Done in my last commit!

parhammmm (approved):
@ksfreitas LGTM, retested and all working well! 

Thanks for putting this together

PR #969, opened 2024-06-25
docs: change references from core package to llamaindex

garviand:


marcusschiesser:
Thanks @garviand! great catch

PR #994, opened 2024-06-30
fix: disable External Filters for Gemini

PeronGH:
This pull request disables external filters for Google Gemini models. This could fix errors like this:

```
Error: [GoogleGenerativeAI Error]: Candidate was blocked due to SAFETY
```

The modification is based on the [official documentation](https://ai.google.dev/gemini-api/docs/safety-settings).

### DIFF
--- a/.changeset/fluffy-knives-glow.md
+++ b/.changeset/fluffy-knives-glow.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+fix: disable External Filters for Gemini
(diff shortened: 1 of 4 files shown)


himself65 (changes requested):
use `HARM_BLOCK_THRESHOLD_UNSPECIFIED` instead

PeronGH:
> use `HARM_BLOCK_THRESHOLD_UNSPECIFIED` instead

The [documentation](https://ai.google.dev/gemini-api/docs/safety-settings#safety-filtering-per-request) states that `HARM_BLOCK_THRESHOLD_UNSPECIFIED` will "block using default threshold", which still leaves the external filters on. `BLOCK_NONE` will "always show regardless of probability of unsafe content".

himself65 (approved):
Oh, nice catch, thanks for the fix

Issue #996, opened 2024-07-01
Image input not working correctly with Anthropic models in LlamaIndex v0.4.6


erik-balfe:
## Description

When using LlamaIndex v0.4.6 with Anthropic models (specifically Claude-3), the image input is not being processed correctly. The model returns text indicating that no image was uploaded, or returns an error about an empty message when only an image is provided.

## Description

When using LlamaIndex v0.4.6 with Anthropic models (specifically Claude-3), the image input is not being processed correctly. The model returns text indicating that no image was uploaded, or returns an error about an empty message when only an image is provided.

## Steps to Reproduce

1. Install LlamaIndex v0.4.6
2. Use the following code to make a request to an Anthropic model:

``` typescript
export async function testLlamaIndexRequest() {
  const llm = new Anthropic({
    model: 'claude-3-haiku',
    apiKey: ANTHROPICAPIKEY,
    maxTokens: 3000,
  });

  const response = await llm.chat({
    messages: 
      {
        role: "system",
        content: `You are a helpful assistant.`,
      },
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "What do you see in this image?"
          },
          {
            type: 'image_url',
            image_url: {
[...]

PR #999, opened 2024-07-01
fix: anthropic image input

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/996

### DIFF
--- a/packages/llamaindex/src/llm/anthropic.ts
+++ b/packages/llamaindex/src/llm/anthropic.ts
@@ -219,7 +219,7 @@ export class Anthropic extends ToolCallLLM<AnthropicAdditionalChatOptions> {
         return {
           content:
             typeof message.content === "string"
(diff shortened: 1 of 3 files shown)


packages/llamaindex/src/llm/anthropic.ts
```diff
@@ -214,7 +217,32 @@ export class Anthropic extends ToolCallLLM<AnthropicAdditionalChatOptions> {
...
+                          text: content.text,
+                        }
+                      : {
+                          type: "image",
+                          source: {
+                            data: content.image_url.url.substring(
+                              content.image_url.url.indexOf(",") + 1,
+                            ),
+                            media_type:
+                              `image/${content.image_url.url.substring("data:image/".length, content.image_url.url.indexOf(";base64"))}` as
```
himself65, line 237: let it throw

PR #1026, opened 2024-07-08
refactor: move llm & callback manager to core module

himself65:


yisding:
Is there a way to do this without breaking every existing app/example? (Maybe by reexporting) I think we created a lot of pain for folks on the Python side when we made the similar change there.

himself65:
> Is there a way to do this without breaking every existing app/example? (Maybe by reexporting) I think we created a lot of pain for folks on the Python side when we made the similar change there.

There are some ways to do this.

1. keep a shadow file there and re-export from the new file. (I personally don't do this because it's too hard to maintain)
2. migration script, like use codemod(https://docs.codemod.com/guides/building-codemods/codemod2) you can search AST tree and replace the module import to the new path

himself65:
But I think in the most case, LITS won't break user side, because we suggest import from top entry `import {} from 'llamaindex'` So everything will keep the same

yisding:
OK as long as the most common import that 90% of folks use doesn't break, that's the most important thing. And then for advanced users, a codemod should suffice.

PR #1037, opened 2024-07-12
fix: DataCloneError when using FunctionTool

brianpeiris:
In #1026, `CallbackManager.dispatchEvent` was changed from [using a spread operator](https://github.com/run-llama/LlamaIndexTS/blob/llamaindex%400.4.14/packages/llamaindex/src/callbacks/CallbackManager.ts#L223-L225) to using `structuredClone`. This breaks with `llm-tool-result` events, since `FunctionTool` has an `_fn` property with the original function. `structuredClone` throws a `DataCloneError` since it cannot clone functions.

I'm not sure if there was more motivation behind using `structruedClone`, but I've added a test to reproduce the error with a simplified event, and changed `dispatchEvent` back to using a spread operator to perform a shallow clone of the event detail.

### DIFF
--- a/.changeset/short-plants-sip.md
+++ b/.changeset/short-plants-sip.md
@@ -1,0 +1,6 @@
+---
+"@llamaindex/core": patch
+"llamaindex": patch
+---
+
+fix: DataCloneError when using FunctionTool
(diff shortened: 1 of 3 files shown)


himself65 (approved):

himself65:
> I'm not sure if there was more motivation behind using `structruedClone`, but I've added a test to reproduce the error with a simplified event, and changed `dispatchEvent` back to using a spread operator to perform a shallow clone of the event detail.

I was thinking event itself shouldn't have non-clonable data.

PR #1063, opened 2024-07-22
feat: add a reader for JSON data

KindOfAScam:
I've been working with a lot of JSON data recently, and since a JSON reader already exists in python, I figured it would be a good addition.

## Summary
This PR introduces a JSON reader capable of parsing JSON data and creating documents from top-level arrays or performing a recursive depth-first traversal of the JSON structure.

## Features
**Default Parsing:**
- Parses the entire JSON string.
- Removes purely structural characters.
- Creates a document for each top-level array element.

**Recursive Depth-First Traversal:** 
- Provides an alternative parsing method.
- Performs a depth-first traversal of the JSON structure.
- Yields paths during traversal for more granular control and analysis.

## Implementation Details

- Added custom error classes for better debugging. Could be expanded on.
- Handles both JSON and JSON Lines format
- Create documents with `traversal_data` as metadata

## Testing
- I did run test with a bunch of different files and inputs to test for edge cases
- Added a comprehensive test suite

## TO-DO:

- Implement JSON streaming to handle large JSON data

#

Looking forward to any feedback or suggestions! Also happy about anyone testing for more edge cases.

### DIFF
--- a/packages/llamaindex/src/readers/JSONReader.ts
+++ b/packages/llamaindex/src/readers/JSONReader.ts
@@ -89,26 +89,42 @@ export class JSONReader<T extends JSONValue> extends FileReader {
 
   private *parseJsonString(jsonStr: string): Generator<T> {
     if (this.options.isJsonLines) {
-      // Process each line as a separate JSON object for JSON Lines format
-      for (const line of jsonStr.split("\n")) {
-        if (line.trim() !== "") {
-          try {
-            yield JSON.parse(line.trim());
-          } catch (e) {
-            throw new JSONParseError(
-              `Error parsing JSON Line: ${e} in "${line.trim()}"`,
-            );
-          }
+      yield* this.parseJsonLines(jsonStr);
+    } else {
+      yield* this.parseJson(jsonStr);
+    }
+  }
+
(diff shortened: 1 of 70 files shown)


packages/llamaindex/src/readers/JSONReader.ts
himself65, line 1: Should we do such work on parsing JSON? since you can do `globalThis.JSON.parse(text)`

just wanna ask for a real case for json reader

KindOfAScam: @himself65 Valid question.

- JSON.parse parse creates a more human readable output, with indent an all, while `JSONReader` creates something of a minimum viable output, where opening brackets are included, and no comma indicates closing brackets. This should result in less tokens for embeddings.
- can handle JSON Lines
- depth-first-traversal: I found it really useful when working with deeply nested structure. With `collapseLength` properly set, you also get a second way of portraying JSON output in a compact way.
- it creates documents.

So in all, it should prepare JSON better for embedding with more options than just `JSON.parse`

himself65 (commented):
could you please add example in `examples/readers/...`

KindOfAScam:
@himself65 Yeah, added one

himself65 (approved):

PR #1069, opened 2024-07-24
feat: add support for buffer on loadJSON API on LlamaParseReader

diervo:
Allows `loadJson` to receive not only a filePath but a buffer with the content of the file.


diervo:
I should have asked first via issue if the maintainers would consider this change, but since its a non-breaking change and the change is small, I figured might as well try with a PR :)

himself65 (approved):

marcusschiesser (approved):

marcusschiesser:
Thanks @diervo !

PR #1087, opened 2024-07-30
fix: handling errors in splitBySentenceTokenizer

marcusschiesser:


himself65 (approved):

himself65:
I made some changes; ignore the error for now. It's safe to just return the as-is downstream sentence splitter will use another splitter in this case

PR #1097, opened 2024-08-01
fix: enforce `no-base-to-string`

himself65:


packages/llamaindex/src/engines/query/SubQuestionQueryEngine.ts
```diff
@@ -119,15 +119,15 @@ export class SubQuestionQueryEngine
...
         query: question,
       });
 
-      if (!responseText) {
+      if (responseValue == null) {
         return null;
       }
 
-      const nodeText = `Sub question: ${question}\nResponse: ${responseText}`;
+      const nodeText = `Sub question: ${question}\nResponse: ${typeof responseValue === "string" ? responseValue : JSON.stringify(responseValue)}`;
```
himself65, line 130: maybe need abstract a function to prettify json value to string

Issue #1098, opened 2024-08-02
SentenceSplitter doesn't take care of excludedEmbedMetadataKeys


pserrer1:
The current SentenceSplitter has some issues with `excludedEmbedMetadataKeys`. The `TextNode`s created after the split containing the whole content including all metadata entries as text.

I'm not fully sure how this is supposed to work, but my naive assumption would be that `MetadataAwarTextSplitter` should use `node.getContent(MetadataMode.NONE)` instead of `node.getContent(MetadataMode.ALL)`

himself65:
Could you please add a test for the case? Thanks

An-nay-marks:
Hi, this is still an issue for llamaindex version 0.12.22: I use the `Document` class with `excluded_embed_metadata_keys` and forward it into an ingestion pipeline with SentenceSplitter:
`transformations = [
            SentenceSplitter(chunk_size=150, chunk_overlap=10), 
            dense_embedding_model
        ]
            ingestion_pipeline = IngestionPipeline(
            transformations=transformations,
            vector_store=self.vector_store,
        )
`

Input to the ingestion pipeline are Documents like
`Document(text=..., metadata=...,  id_=..., excluded_embed_metadata_keys=[...])`


 However, all metadata get's taken into account in the SentenceSplitter when checking for max chunk size. I get the error:
```
../../../../anaconda3/envs/testenv/lib/python3.11/site-packages/llama_index/core/ingestion/pipeline.py:545: in run
    nodes = run_transformations(
../../../../anaconda3/envs/testenv/lib/python3.11/site-packages/llama_index/core/ingestion/pipeline.py:97: in run_transformations
    nodes = transform(nodes, **kwargs)
../../../../anaconda3/envs/testenv/lib/python3.11/site-packages/llama_index/core/instrumentation/dispatcher.py:322: in wrapper
    result = func(*args, **kwargs)
../../../../anaconda3/envs/testenv/lib/python3.11/site-packages/llama_index/core/node_parser/interface.py:193: in __call__
    return self.get_nodes_from_documents(nodes, **kwargs)  # type: ignore
../../../../anaconda3/envs/testenv/lib/python3.11/site-packages/llama_index/core/node_parser/interface.py:165: in get_nodes_from_documents
    nodes = self._parse_nodes(documents, show_progress=show_progress, **kwargs)
[...]

PR #1099, opened 2024-08-02
fix: metadata should not be included after sentence splitter

pserrer1:
See https://github.com/run-llama/LlamaIndexTS/issues/1098

### DIFF
--- a/packages/core/src/node-parser/base.ts
+++ b/packages/core/src/node-parser/base.ts
@@ -140,7 +140,7 @@ export abstract class MetadataAwareTextSplitter extends TextSplitter {
     return nodes.reduce<TextNode[]>((allNodes, node) => {
       const metadataStr = this.getMetadataString(node);
       const splits = this.splitTextMetadataAware(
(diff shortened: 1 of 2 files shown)


packages/core/src/node-parser/base.ts
```diff
@@ -140,7 +140,7 @@ export abstract class MetadataAwareTextSplitter extends TextSplitter {
     return nodes.reduce<TextNode[]>((allNodes, node) => {
       const metadataStr = this.getMetadataString(node);
       const splits = this.splitTextMetadataAware(
-        node.getContent(MetadataMode.ALL),
+        node.getContent(MetadataMode.NONE),
```
himself65, line 143: You are right. Nice catch

https://github.com/run-llama/llama_index/blob/769e76a28aa497ec9885975a09eb8eec24713597/llama-index-legacy/llama_index/legacy/node_parser/interface.py#L175

himself65 (approved):

PR #1119, opened 2024-08-13
feat: add json streaming to JSONReader

KindOfAScam:
added streaming, refactored and cleaned up the JSON Reader class.

I did test with various JSON files ranging from a few hundred MBs up to a GB, but definitely could use some more testing to find edge cases. 

(cc: @ozgur142)


himself65 (approved):

packages/llamaindex/src/readers/JSONReader.ts
```diff
@@ -1,71 +1,199 @@
+import { parseChunked } from "@discoveryjs/json-ext";
 import type { JSONValue } from "@llamaindex/core/global";
 import { Document, FileReader } from "@llamaindex/core/schema";
+import { ReadableStream } from "@llamaindex/env";
```
himself65, line 4:
```suggestion
```

PR #1123, opened 2024-08-15
feat: support organization id in llamacloud index

thucpn:


packages/llamaindex/src/cloud/LlamaCloudIndex.ts
```diff
@@ -132,18 +132,46 @@ export class LlamaCloudIndex {
...
-    name: string,
-    projectName: string,
+  public async getPipelineId(
+    name?: string,
+    projectName?: string,
   ): Promise<string> {
     const pipelines = await PipelinesService.searchPipelinesApiV1PipelinesGet({
-      projectName,
-      pipelineName: name,
+      projectName: projectName ?? this.params.projectName,
```
marcusschiesser, line 140: as the projectName is not unique, doesn't we need here org ID too? (please check python code)

marcusschiesser: and please double check other uses of just projectName in this file

marcusschiesser (approved):

PR #1133, opened 2024-08-23
fix: use metadata mode LLM for generating context

thucpn:
Should use mode LLM to generating context in ContextChatEngine LITS.
See Python usage: https://github.com/run-llama/llama_index/blob/96a196a9d0bed9ea8a3b9f9caac18639faaf0249/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py#L167-L169

### DIFF
--- a/.changeset/tough-cups-doubt.md
+++ b/.changeset/tough-cups-doubt.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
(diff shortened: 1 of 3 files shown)


marcusschiesser (approved):

marcusschiesser (approved):
Thanks @thucpn

Issue #1135, opened 2024-08-27
Create MongoDB Vector index programatically


marcusschiesser:
It's possible to create an index programmatically, see https://www.mongodb.com/docs/atlas/atlas-search/create-index/#create-an-fts-index-programmatically

I tried this a couple of months ago, here:
https://github.com/run-llama/LlamaIndexTS/blob/822a2665badf7d7ec1661b9d9715d1eabcbe9f4e/examples/mongodb/2_load_and_index.ts#L52

Which did not work

Motivation: people think MongoDB integration is not working, see https://github.com/run-llama/create-llama/issues/240

PR #1139, opened 2024-08-30
feat: update mongodb vector store example

thucpn:
https://github.com/run-llama/LlamaIndexTS/issues/1135

### DIFF
--- a/packages/llamaindex/src/storage/vectorStore/MongoDBAtlasVectorStore.ts
+++ b/packages/llamaindex/src/storage/vectorStore/MongoDBAtlasVectorStore.ts
@@ -244,4 +244,26 @@ export class MongoDBAtlasVectorSearch
     console.debug("Result of query (ids):", ids);
     return result;
   }
+
+  async createSearchIndex() {
+    // define your Atlas Search index. See detail https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/
+    const index = {
+      name: this.indexName,
+      definition: {
+        /* search index definition fields */
+        mappings: {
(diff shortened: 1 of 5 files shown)


packages/llamaindex/src/storage/vectorStore/MongoDBAtlasVectorStore.ts
```diff
@@ -244,4 +244,26 @@ export class MongoDBAtlasVectorSearch
     console.debug("Result of query (ids):", ids);
     return result;
   }
+
+  async createSearchIndex() {
```
marcusschiesser, line 248: how about
1. create this.collection lazily (async funciton)
2. add `autoCreateIndex` flag to constructor (defaults to `true`)
3. if `autoCreateIndex==true`, create the index (if necessary)

then user doesn't have to call `vectorStore.createSearchIndex();`

(better DX)

- also don't forget to about README (removing manually creating the index)

packages/llamaindex/src/storage/vectorStore/MongoDBAtlasVectorStore.ts
```diff
@@ -127,6 +146,32 @@ export class MongoDBAtlasVectorSearch
     this.insertOptions = init.insertOptions;
   }
 
+  async ensureCollection() {
+    if (!this.collection) {
+      const collection = await this.mongodbClient
+        .db(this.dbName)
+        .createCollection(this.collectionName);
```
marcusschiesser, line 153: what happens if the collection already exists?

marcusschiesser (approved):

PR #1144, opened 2024-08-30
feat: support convert from python side

himself65:


logan-markewich:
@himself65 just checking, is this a breaking change for people that already have saved data?

himself65:
> @himself65 just checking, is this a breaking change for people that already have saved data?

I think so, let me add backforward compability for now

PR #1150, opened 2024-09-05
refactor: engine stuff

himself65:
unified `ChatEngine`, `QueryEngine`, `Synthesizer` and related module and types

goal is to make module same as llama-index python and give user reasonable APIs


himself65:
we have moved, clsoing this one

PR #1151, opened 2024-09-05
feat: add upstash vector store

ogzhanolguncu:
This PR adds Upstash Vector as a storage.


packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+        "Must specify UPSTASH_VECTOR_REST_TOKEN via env variable.",
+      );
+    }
+    if (!endpoint) {
+      throw new Error("Must specify UPSTASH_VECTOR_REST_URL via env variable.");
+    }
+    this.db = new Index({ token, url: endpoint });
+  }
+
+  private async getDb(): Promise<Index> {
```
fahreddinozcan, line 50: do we really need this method? we can achieve the same thing with`new UpstashVectorStore`

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+   * @returns Due to limitations in the Upstash client, does not return the upserted ID list, only a Promise resolve/reject.
+   */
+  async add(embeddingResults: BaseNode<Metadata>[]): Promise<string[]> {
+    if (embeddingResults.length == 0) {
+      return Promise.resolve([]);
+    }
+
+    const db = this.db;
+    const nodes = embeddingResults.map(this.nodeToRecord);
+    const result = await db.upsert(nodes, { namespace: this.namespace });
```
fahreddinozcan, line 79: We can maybe automatically split a data with >1000 vectors, automatically ourselves? We do this with langchain

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+    return Promise.resolve([]);
+  }
+
+  /**
+   * Adds plain text record(s) to the table. Upstash take cares of embedding conversion.
+   * @param embeddingResults The Nodes to be inserted, optionally including metadata tuples.
+   * @returns Due to limitations in the Upstash client, does not return the upserted ID list, only a Promise resolve/reject.
+   */
+  async addPlainText(
+    embeddingResults: TextNode<Metadata>[],
```
fahreddinozcan, line 92: When we're adding plaintext, is this still `embeddingResult`? We can maybe rename this param

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+   * @param options Required by VectorStore interface.  Currently ignored.
+   * @returns Zero or more Document instances with data from the vector store.
+   */
+  async query(
+    query: VectorStoreQuery,
+    _options?: any,
+  ): Promise<VectorStoreQueryResult> {
+    const filter = this.toUpstashFilter(query.filters);
+
+    const defaultOptions: any = {
```
fahreddinozcan, line 142: I guess we can get the type from sdk, instead of any

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
ogzhanolguncu: It is in the interface of the class

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
ogzhanolguncu: Sure, sounds good.

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
ogzhanolguncu: I think its safe to keep it like this since its internal.

ryanleecode:
this should be a `minor` changeset b/c its adding new feature

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+} from "./types.js";
+
+import type { BaseNode, Metadata, TextNode } from "@llamaindex/core/schema";
+import { getEnv } from "@llamaindex/env";
+import { Index } from "@upstash/vector";
+import { metadataDictToNode, nodeToMetadata } from "./utils.js";
+
+type UpstashParams = {
+  namespace?: string;
+  token: string;
```
ryanleecode, line 17: adding docs that it defaults to `process.env.UPSTASH_VECTOR_REST_TOKEN` and `process.env.UPSTASH_VECTOR_REST_URL` would be good

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return Promise.resolve([]);
+  }
+
+  /**
+   * Adds plain text record(s) to the table. Upstash take cares of embedding conversion.
+   * @param embeddingResults The Nodes to be inserted, optionally including metadata tuples.
+   * @returns Due to limitations in the Upstash client, does not return the upserted ID list, only a Promise resolve/reject.
+   */
+  async addPlainText(
```
ryanleecode, line 91: this is already asnyc/await function, we don't need to use `Promise.resolve` / `Promise.reject`. Just return to throw.

packages/llamaindex/src/storage/vectorStore/UpstashVectorStore.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Deletes a single record from the database by id.
+   * NOTE: Uses the collection property controlled by setCollection/getCollection.
+   * @param refDocId Unique identifier for the record to delete.
+   * @returns Promise that resolves if the delete query did not throw an error.
+   */
+  async delete(refDocId: string): Promise<void> {
+    await this.db.namespace(this.namespace).delete(refDocId);
+    return Promise.resolve();
```
himself65, line 115: for async function, you dont need wrap `Promise.resolve`

CahidArda:
Hi @himself65,

we would like to close this PR and continue with another PR because Oguzhan is no longer with our team.

I can fork his fork and continue on another PR

Issue #739, opened 2024-04-19
strong typing for LlamaIndexTS


himself65:
I've noticed most of the PRs recently are generated by LLM, and this project has had some issues.

1. No type check, no information. It takes work to maintain and understand what's going on.

For example, this API doesn't tell what are `objects`, `indexCls`, and `indexKwargs`. Even though there's some trouble, it won't get fixed as time passes.

https://github.com/run-llama/LlamaIndexTS/blob/208282d62fdcdb4e91c0b0800727f00a8456bba7/packages/core/src/objects/base.ts#L155-L161

2. Not a good design; it's just a literary copy-pasted version of `llama index` Python code. llama index is a widely validated and used package, but some architectures are unsuitable for JS/TS. I listed here:
	1. async/sync: there's no `sync` call on the JS side, so every function call relies on LLM (uses network request) and is async-forced.
	2. iterator: JS has built-in async/streaming support, so there's no need to wrap into a manual iterator
	3. class & interface: JS doesn't have an actual class syntax; it's a syntactic sugar. And there's some issue when combining the TS interface with the class (abstract class/ static class?)
	4. For dynamic function calls, there are lots of `fromDefaults` | `fromObjects` ... which you can rename them all into `from` and use typescript function overloading to handle it, which will simplify the APIs
	5. generic type: python doesn't have a substantial type check as the typescript. So, most of the code is more challenging to maintain when pasted from the Python side. Because on the TS side, fewer developers are maintaining this project.
	6. What is the goal of the TS project? LITS aims to support all core features of the Python side. However, the developers in JS/TS are more likely to use LLM in web apps and some end products instead of data analysis... Anyway, even the core design will be the same, but the final APIs should go toward the JS&Web standard and be compatible with JS runtimes(node.js/deno/bun/edge/cloudflare...) as much as possible. At this point, we are not targeting the Web(which means you are not supposed to literary import LITS in the browser), but we can gradually port some packages for cross-runtimes step by step.

Some PRs need to be reworked to fit the strong type:

- https://github.com/run-llama/LlamaIndexTS/pull/594
- https://github.com/run-llama/LlamaIndexTS/pull/559
- https://github.com/run-llama/LlamaIndexTS/pull/543
- https://github.com/run-llama/LlamaIndexTS/pull/484
- https://github.com/run-llama/LlamaIndexTS/pull/412
- https://github.com/run-llama/LlamaIndexTS/pull/408

himself65:
[...]

PR #1154, opened 2024-09-05
refactor: prompt system

himself65:
Related issue: https://github.com/run-llama/LlamaIndexTS/issues/739

1. type safe, will even check string template include {str}
2. python template string compatible
3. same API with PromptTemplate and BaseTemplate...


packages/core/src/schema/type/base-output-parser.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * An OutputParser is used to extract structured data from the raw output of the LLM.
+ */
+export interface BaseOutputParser {
+	parse(output: string): any;
```
himself65, line 5: could we remove any?

packages/core/src/prompts/prompt.ts
himself65, line 1: haven't finished yet

packages/core/package.json
```diff
@@ -132,7 +146,8 @@
   "devDependencies": {
     "ajv": "^8.17.1",
     "bunchee": "5.3.2",
-    "natural": "^8.0.1"
+    "natural": "^8.0.1",
+    "python-format-js": "^1.4.3"
```
AndreMaz, line 150: @himself65 I think that the `python-format-js` should be in `dependencies` and not in `devDependencies`. 

I'm saying this because it's being used at  `packages/core/src/prompts/base.ts`.

Can you please confirm?

packages/core/src/prompts/base.ts
```diff
@@ -0,0 +1,225 @@
+import format from "python-format-js";
```
AndreMaz, line 1: @himself65 this is where `python-format-js` is being imported

PR #1155, opened 2024-09-06
feat: build wasm with extism

thucpn:
https://github.com/run-llama/create-llama/issues/252


packages/extism-tools/src/WikipediaTool.ts
```diff
@@ -0,0 +1,61 @@
...
+      query: {
+        type: "string",
+        description: "The query to search for",
+      },
+    },
+    required: ["query"],
+  },
+};
+
+export class WikipediaTool implements BaseTool<WikipediaParameter> {
```
marcusschiesser, line 28: what do we need to do to make this more genric, like an `ExtismTool`?

marcusschiesser: e.g. the how can we get the WASM function signature at run time and convert it to a JSON schema?

packages/extism-tools/bin/compile.js
```diff
@@ -0,0 +1,28 @@
+import { execSync } from "child_process";
```
marcusschiesser, line 1:
```suggestion
// needed as extism-js doesn't support compiling multiple files
import { execSync } from "child_process";
```

.github/workflows/lint_on_push_or_pull.yml
```diff
@@ -19,6 +19,10 @@ jobs:
         with:
           node-version-file: ".nvmrc"
           cache: "pnpm"
+      - name: Install Extism
```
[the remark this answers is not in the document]
marcusschiesser: also please double check if installing extism is really needed for all the other workflow steps below

packages/extism-tools/README.md
```diff
@@ -0,0 +1,30 @@
+## Extism Tools
+
+### Prerequisites for Development
+
+- [Extism PDK](https://github.com/extism/js-pdk?tab=readme-ov-file#linux-macos)
+
+### Build WASM files
+
+```bash
+cd packages/wasm-tools/extism/wiki
```
marcusschiesser, line 10: there is a build script now - document this instead

packages/extism-tools/README.md
```diff
@@ -0,0 +1,30 @@
...
+extism-js wiki.js -i wiki.d.ts -o wiki.wasm
+```
+
+### Run WASM files in Node.js using Extism SDK (https://github.com/extism/js-sdk)
+
+```bash
+node examples/wasm/wiki.js
```
marcusschiesser, line 17: how about using tsx?

packages/extism-tools/examples/tool/wiki.js
```diff
@@ -0,0 +1,12 @@
+import { OpenAIAgent } from "llamaindex";
+import { ExtismToolFactory } from "../../dist/ExtismToolFactory.js";
```
marcusschiesser, line 2: goal should be to use `exitism-tools` as package, so add a package.json to the examples folder that is using this package

packages/extism-tools/src/ExtismToolFactory.ts
```diff
@@ -0,0 +1,85 @@
...
+  WIKI = "wiki",
+}
+
+export const ToolMap: Record<`${ExtismTool}`, CreateToolClassParams> = {
+  [ExtismTool.WIKI]: {
+    wasmFilename: "wiki.wasm",
+    allowedHosts: ["*.wikipedia.org"],
+    maxHttpResponseBytes: DEFAULT_MAX_HTTP_RESPONSE_BYTES,
+    transformResponse: (response: any) => response.extract,
+  },
```
marcusschiesser, line 31: 1. can we move these default paramters to the WASM? (Similar to `getMetadata`)
2. The user can overwrite these parameters - should not be part of this factory. The factory should be tool code agnostic

packages/extism-tools/src/ExtismToolFactory.ts
```diff
@@ -0,0 +1,85 @@
...
+          const data = await pluginInstance.call("call", JSON.stringify(input));
+          if (!data) return "No result";
+          const result = config.transformResponse(data.json());
+          await pluginInstance.close();
+          return result;
+        }
+      };
+    } catch (e) {
+      console.error(e);
+      throw new Error("Failed to create WikipediaTool instance.");
```
marcusschiesser, line 80:
```suggestion
      throw new Error("Failed to create Tool instance.");
```

packages/extism-tools/src/ExtismToolFactory.ts
```diff
@@ -0,0 +1,85 @@
...
+    memory: { maxHttpResponseBytes },
+  });
+  return plugin;
+};
+
+export class ExtismToolFactory {
+  static async createToolClass(
+    toolName: `${ExtismTool}`,
+  ): Promise<new (params: ToolClassParams) => BaseTool<ToolParams>> {
+    const config = ToolMap[toolName];
```
marcusschiesser, line 51: to be tool-agnostic, just use a convention to find the WASM file
```suggestion
    const wasmFileName =`${WASM_DIR}/${toolName}.wasm`;
```

packages/extism-tools/src/ExtismToolFactory.ts
marcusschiesser: then how about we use per default:
1. we allow all hosts 
2. we use xMB memory (do we need 100MB?)
3. we use `response.extract` 
all this can be changed in `createToolClass` call if needed

marcusschiesser: or better, we have a `${toolName}.config` file that contains these parameters and it is loaded with `createToolClass`

PR #1177, opened 2024-09-11
Had to make these changes to get RouterQueryEngine working

kieransimkin:
I had to make these changes in order to get RouterQueryEngine working at all on my Windows desktop. Interestingly, the quotes in the JSON only seem to be necessary in Windows - it works in linux totally fine. 

The change from `query.query` to just `query` was necessary on both platforms though, so I'm not sure if that ever worked.


packages/llamaindex/src/selectors/llmSelectors.ts
```diff
@@ -159,7 +159,7 @@ export class LLMSingleSelector extends BaseSelector {
     const prompt = this.prompt.format({
       numChoices: `${choicesText.length}`,
       context: choicesText,
-      query: extractText(query.query),
+      query: extractText(query),
```
himself65, line 162: nice catch

himself65 (approved):

packages/llamaindex/src/outputParsers/selectors.ts
```diff
@@ -12,8 +12,8 @@ const formatStr = `The output should be ONLY JSON formatted as a JSON instance.
 Here is an example:
 [
     {
-        choice: 1,
-        reason: "<insert reason for choice>"
+        "choice": 1,
+        "reason": "<insert reason for choice>"
```
himself65, line 16: this is might unnecessary

himself65:
landed in https://github.com/run-llama/LlamaIndexTS/commit/4810364788f5a38307206e4a5021bcf274f22696

PR #1180, opened 2024-09-11
feat: add PostgreSQL storage

gorango:
Update kvStore, docStore, and indexStore with required integrations (#1171)

Tested locally but unable to add tests due to dependence on a running pg instance.


gorango (commented):
I was not aware that the `pg` package leverages env vars as a fallback for initializing the client (I updated the config thinking it was an oversight in the current implementation).

I will revert the changes to the config here and update the kvStore

gorango:
For the sake of naming consistency, I recommend that we either:

- rename PGVectorStore to PostgresVectorStore, or...
- rename PostgresKVStore, ...DocStore, and ...IndexStore to PGKVStore, ...DocStore, and ...IndexStore

I would be in favor of having Postgres explicit in the filenames (instead of PG) but will leave it up to maintainers to decide.

(FWIW the Python package uses postgres in the filenames)

himself65 (commented):
The code looks good to me. But maybe I need to merge https://github.com/run-llama/LlamaIndexTS/pull/1175, which adds unit test and some small fix

gorango:
Reverted all changes to the vectorStore (they were mostly semantic anyway) - should help to avoid merge conflicts.

himself65 (approved):

PR #1188, opened 2024-09-12
Add workflows

marcusschiesser:
usage see examples in examples/workflows


examples/workflow/collect-events.ts
```diff
@@ -0,0 +1,70 @@
...
+  Workflow,
+  WorkflowEvent,
+} from "@llamaindex/core/workflow";
+import { OpenAI } from "llamaindex";
+
+// Create LLM instance
+const llm = new OpenAI();
+
+// Create custom event types
+export class JokeEvent extends WorkflowEvent<{ joke: string }> {}
```
marcusschiesser, line 14: I was first not a big fan of creating the events via subclassing and preferred a factory.

You can compare both options by checking this commit: https://github.com/run-llama/LlamaIndexTS/pull/1188/commits/0237d9563049140ee9b5b23de5c15f7924736e7e

Advantages of sub-classing:
1. No issue in dealing with unnamed events
2. Just one line per event definition (in TS)
3. Simpler events code in framework

Disadvantages of sub-classing:
1. Have to use the `data` property

So now I am preferring the sub-classing variant.

packages/core/src/workflow/workflow.ts
```diff
@@ -0,0 +1,221 @@
...
+      timeout?: number;
+      validate?: boolean;
+    } = {},
+  ) {
+    this.#verbose = params.verbose ?? false;
+    this.#timeout = params.timeout ?? null;
+    this.#validate = params.validate ?? false;
+  }
+
+  addStep<T extends WorkflowEvent>(
```
logan-markewich, line 38: For some reason, my brain really wants `addStep` to have the step as the first argument 😅

logan-markewich (approved):

himself65 (commented):
adding some unit test? this not require llm

PR #1189, opened 2024-09-12
fix: cannot import groq-sdk in nextjs

thucpn:


himself65 (commented):
This will cause nextjs build error.

Issue #1179, opened 2024-09-11
Error: Unknown node type: DOCUMENT


AndreMaz:
@himself65 bumped to `0.5.24` to see If the https://github.com/run-llama/LlamaIndexTS/issues/1176 is fixed. It is but now I'm getting the following error:

> Error: Unknown node type: DOCUMENT
at splitNodesByType (webpack-internal:///(rsc)/../../node_modules/@llamaindex/core/dist/schema/index.js:708:19)
at createMessageContent (webpack-internal:///(rsc)/../../node_modules/llamaindex/dist/synthesizers/utils.js:11:94)
at DefaultContextGenerator.generate (webpack-internal:///(rsc)/../../node_modules/llamaindex/dist/engines/chat/DefaultContextGenerator.js:50:107)
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
at async ContextChatEngine.prepareRequestMessages (webpack-internal:///(rsc)/../../node_modules/llamaindex/dist/engines/chat/ContextChatEngine.js:470:25)
at async ContextChatEngine.chat (webpack-internal:///(rsc)/../../node_modules/llamaindex/dist/engines/chat/ContextChatEngine.js:437:33)
at async Object.start (webpack-internal:///(rsc)/./src/app/api/chat/route.ts:60:32)

Note: This error was not present in the v0.5.20. Something that I initially described in https://github.com/run-llama/LlamaIndexTS/issues/1172

I can repro the issue exactly in the same way as I've mentioned in the ticket above.
From the ticket:

> I'm using llamaindex at app/api/chat/route.ts , in other words, I'm using [Route Handlers](https://nextjs.org/docs/app/building-your-application/routing/route-handlers)
> When I start my app and open /home next.js compiles the app and I can call llamaindex at app/api/chat/route.ts without any error. Then, when I navigate to another page (e.g, /products) next.js does another compilation. After this when I call the handler at app/api/chat/route.ts I get the error above.

himself65:
what's your code snippt? I'm guessing there's dual module in the output

himself65:
I cannot reproduce the issue but I think this might related to the bundler

AndreMaz:
Here's my `next.config.mjs` 
```js
/**
 * @type {import("next").NextConfig}
 */

import { fileURLToPath } from "url";
import _jiti from "jiti";
import withLlamaIndex from "llamaindex/next";

[...]

PR #1214, opened 2024-09-16
fix: replace instanceof check with `.type` check

himself65:
Fixes: https://github.com/run-llama/LlamaIndexTS/issues/1179
Similar issue: https://github.com/yjs/yjs/issues/438

### DIFF
--- a/.changeset/moody-feet-sit.md
+++ b/.changeset/moody-feet-sit.md
@@ -1,0 +1,6 @@
+---
+"@llamaindex/core": patch
+"@llamaindex/env": patch
+---
+
+fix: replace instanceof check with `.type` check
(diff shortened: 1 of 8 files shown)


parhammmm:
if anyone is using next.js and comes across this, adding the following to your `next.config.mjs` this will fix it:

`experimental: { serverComponentsExternalPackages: ["@llamaindex/env",...] },`

himself65:
thanks for updating solution

nick-solly:
> if anyone is using next.js and comes across:
> 
> `llamaindex was already imported. This breaks constructor checks and will lead to issues!`
> 
> Adding the following to your `next.config.mjs` this will fix it:
> 
> `experimental: { serverComponentsExternalPackages: ["@llamaindex/env",...] },`

Thanks @parhammmm - this was driving me insane!

parhammmm:
@nick-solly 

No prob! glad it helped 👍🏼

PR #1219, opened 2024-09-17
feat: add Amazon Bedrock Retriever

ajohn-wick:


parhammmm:
The retrievers shouldn't be in the llm folder, they should be one level higher in `src/retrievers`

ajohn-wick:
> The retrievers shouldn't be in the llm folder, they should be one level higher in `src/retrievers`

Thank you very much for your feedback @parhammmm 
I just moved it according to your comment and pushed my changes.

himself65 (changes requested):
we have 25k+ deletion on pnpm lock, please check that

ajohn-wick:
> we have 25k+ deletion on pnpm lock, please check that

Done, it was a mistake from my end @himself65

packages/community/src/retrievers/bedrock.ts
```diff
@@ -0,0 +1,160 @@
+import {
+  BedrockAgentRuntimeClient,
+  RetrieveCommand,
+  type BedrockAgentRuntimeClientConfig,
+  type RetrievalFilter,
+  type SearchType,
+} from "@aws-sdk/client-bedrock-agent-runtime";
+
+import { BaseRetriever, Document } from "llamaindex";
```
himself65, line 9: import Document from llamaindex/core/schema and copy BaseRetriever type here for now.

`llamaindex` is the to end user level pacakge and shouldn't be imported from our other pacakge, so please remove `llamaindex` from deps

remind me that we need move `BaseRetriever` into core module

himself65 (changes requested):

packages/community/src/retrievers/bedrock.ts
ajohn-wick: Thank you a lot for your review, please proceed with moving `BaseRetriever` into core module. Then, I will update code to import it if you are OKay?
Otherwise, I will copy `BaseRetriever` type here.

himself65: too late to me, will do tmr

himself65 (approved):

PR #1223, opened 2024-09-18
chore: add 'late_chunking' for Jina embedding

DresAaron:
Modifications:
- renamed parameter: `task_type` to `task`
- extended parameter: `late_chunking`

### DIFF
--- a/.changeset/yellow-donkeys-complain.md
+++ b/.changeset/yellow-donkeys-complain.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+Extend JinaAPIEmbedding parameters
--- a/packages/llamaindex/src/embeddings/JinaAIEmbedding.ts
+++ b/packages/llamaindex/src/embeddings/JinaAIEmbedding.ts
@@ -20,8 +20,9 @@ export type JinaEmbeddingRequest = {
   input: Array<{ text: string } | { url: string } | { bytes: string }>;
   model?: string;
   encoding_type?: EncodingType;
-  task_type?: TaskType;
+  task?: TaskType;
   dimensions?: number;
+  late_chunking?: boolean;
 };
 
 export type JinaEmbeddingResponse = {
@@ -44,9 +45,10 @@ export class JinaAIEmbedding extends MultiModalEmbedding {
   apiKey: string;
   model: string;
   baseURL: string;
-  taskType: TaskType | undefined;
+  task?: TaskType | undefined;
   encodingType?: EncodingType | undefined;
   dimensions?: number | undefined;
+  late_chunking?: boolean | undefined;
 
   async getTextEmbedding(text: string): Promise<number[]> {
     const result = await this.getJinaEmbedding({ input: [{ text }] });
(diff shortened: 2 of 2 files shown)


marcusschiesser:
@DresAaron cool idea to add late chunking, but we need to make a couple of changes to use it in an ingestion pipeline like this:

```
const pipeline = new IngestionPipeline({
    transformations: [
      new JinaAIEmbedding({ useLateChunking: true}),
      new SentenceSplitter({ useLateChunking: true, chunkSize: 1024 }),
    ],
  });
```
(Note that the embedding is called before the chunking)

The reason is that by using late chunking, we get not just one embedding per text but an array of embeddings (one embedding for each token in the text, see https://github.com/jina-ai/late-chunking).

Here are some changes I can think of to make it work:

1. We need to store the token embeddings produced by `JinaAIEmbedding` in `BaseNode` (as the pipelines `TransformerComponent`s work on `BaseNode` instances)
2. The `SentenceSplitter` must consider token embeddings of the input `BaseNode` and copy them to the output `BaseNode`s according to the splitting (this applies for other splitters, too). 
3. When an embedding for a node is used, we must apply mean pooling to convert the token embeddings to one embedding for the node (this could maybe be done lazily, or we have to trigger it explicitly, e.g. with a new `MeanPoolEmbeddings` transformer)

@DresAaron, would you like to look into this?

DresAaron:
> @DresAaron cool idea to add late chunking, but we need to make a couple of changes to use it in an ingestion pipeline like this:
> 
> ```
> const pipeline = new IngestionPipeline({
>     transformations: [
>       new LateChunking({ embedModel: new JinaAIEmbedding(...), splitter: new SentenceSplitter({ chunkSize: 1024 }) })
>     ],
>   });
> ```
> 
> The reason is that by using late chunking, we get not just one embedding per text but an array of embeddings (one embedding for each token in the text; see https://github.com/jina-ai/late-chunking).
> 
> The `LateChunkingSplitter` could be implemented like this:
> 
> 1. Call `JinaAIEmbedding` for the input node and temporarily store the token embeddings of that input node.
> 2. Call the assigned splitter to split input nodes into a list of output nodes
> 3. Tokenize the text of each output node (the splitted text) and assign token embeddings for the output node (handle the case that splitter doesn't split at token boundaries)
> 4. Calculate the final embedding for each output node by using mean pooling of the output node's token embeddings
> 
> @DresAaron, would you like to look into this?

@marcusschiesser `late_chunking` by Jina Embedding model v3 does not change the return format, you can check the blogpost here: https://jina.ai/news/late-chunking-in-long-context-embedding-models/
Technically, we don't need to change anything for the pipeline.

JoanFM (approved):

marcusschiesser:
@DresAaron Sorry, maybe there's a misunderstanding. I read https://jina.ai/news/late-chunking-in-long-context-embedding-models/
and here the change of the data pipeline is mentioned:
image
According to this, using late chunking, we call the embedding model on the whole document, and it returns embeddings for each token of the document.

DresAaron:
@marcusschiesser since we are using the API in the implementation, just like `jina-embeddings-v2-base-en`, the response of `jina-embeddings-v3` model in the end returns same format. `jina-embeddings-v3` is already available, maybe you can try it out at: https://jina.ai/embeddings/ 😄

marcusschiesser:
@DresAaron, I guess we've been talking about two different things, but now I am still wondering how to get the token embeddings of a longer text using Jina. Please let me know if you find out.

marcusschiesser (approved):

PR #1225, opened 2024-09-18
feat: update `PGVectorStore` constructor parameters

AndreMaz:
The `PGVectorStore` had several issues that this PR tries to address. 
Problems:
- When passing a custom `pgClient` it is not possible to define PGVectorStore `dimensions` 
- Passing a custom `pgClient` to the PGVectorStore  would not throw any error but the application would "hang" forever trying to resolve the promise that waits for data from the DB. The reason for that is that `private async getDb()`  does not call the `db.connect()`

More context here: https://github.com/run-llama/LlamaIndexTS/issues/1179#issuecomment-2355013674

Link to pg's `end` event: https://node-postgres.com/apis/client#end


himself65 (commented):
> Passing a custom pgClient to the PGVectorStore would not throw any error but the application would "hang" forever trying to resolve the promise that waits for data from the DB. The reason for that is that private async getDb() does not call the db.connect()

If user pass a pg client, we expect it already connected before. And it will throw error if you call `connect` twice

packages/llamaindex/src/vector-store/PGVectorStore.ts
```diff
@@ -131,10 +136,19 @@ export class PGVectorStore
       }
     }
 
-    const db = this.db;
+    if (this.db && !this.dbConnected) {
+      try {
+        await this.db.connect();
+      } catch (error) {
+        console.error(error);
+        return Promise.reject(
```
himself65, line 144: im personally don't like try-catch block to check if already connected. do we have other solution?

AndreMaz: > personally don't like try-catch block to check if already connected

I wouldn't it call it that. The 
```ts
if (this.db && !this.dbConnected) {}
```

in combination with the following event listener

```ts
db.on("end", () => (this.dbConnected = false));
```

Acts as a "re-connection" handler.

With this change,  whenever you call `getDb()` or `client()` it will try to return a client that's already connected to the DB.

If you prefer we can remove this logic completely

AndreMaz:
> If user pass a pg client, we expect it already connected before. And it will throw error if you call connect twice

Yep, yep. I had to create a singleton to solve the " already connected" error.

However, personally I'm not a big fan of the idea that I need to pass an already connected client. I would rather just provide a fully configured client and then leave the connection, re-connection and etc. logic to the `PGVectorStore`.

AndreMaz:
Regarding the error https://github.com/run-llama/LlamaIndexTS/actions/runs/10921048207/job/30334321408?pr=1225

> Error packages/llamaindex/e2e/node/vector-store/pg-vector-store.test.ts(31,41): error TS2559: Type 'PoolClient' has no properties in common with type 'PGVectorStoreConfig'.

and thinking a little bit re the possibility of passing a `pg.Pool` and error handling.

We can potentially modify the `PGVectorStoreConfig` to look like this:
```ts
export type PGVectorStoreConfig = {
  schemaName?: string | undefined;
  tableName?: string | undefined;
  database?: string | undefined;
  connectionString?: string | undefined;
  dimensions?: number | undefined;
  embedModel?: BaseEmbedding | undefined;

  /**
   * Client configuration options for the pg client.
   */
  clientConfigs?: pg.ClientConfig | pg.PoolConfig | undefined;  // <=== CLIENT or POOL configs
  usePool?: boolean | undefined; 
};
```

then `constructor` to something like
```ts
  constructor(configs?: PGVectorStoreConfig) {
    super(configs?.embedModel);
    this.schemaName = configs?.schemaName ?? PGVECTOR_SCHEMA;
    this.tableName = configs?.tableName ?? PGVECTOR_TABLE;
    this.database = configs?.database;
    this.connectionString = configs?.connectionString;
    this.dimensions = configs?.dimensions ?? DEFAULT_DIMENSIONS;
    this.clientConfigs = configs?.clientConfigs ?? {};
    this.usePool = configs?.usePool ?? false;  // <=== use client by default
  }
```

Modify the getDb() like
```ts
// ....
const pg = await import("pg");
const { Client, Pool } = pg.default ? pg.default : pg;

const { registerType } = await import("pgvector/pg");
// Create DB connection
// Read connection params from env - see comment block above
const db = this.usePool
  ? new Client({
      ...this.clientConfigs,
      database: this.database,
      connectionString: this.connectionString,
    })
  : new Pool({
      ...this.clientConfigs,
      database: this.database,
      connectionString: this.connectionString,
    });

// ....
```

and then add even listeners. Something like
```ts
  private registerConnectionListeners(db: pg.Pool | pg.Client) {
    if (this.usePool) {
      const ref = db as pg.Pool;

      ref.on("error", (err) => {
        this.dbConnected = false;
      });

      ref.on("connect", () => {
        this.dbConnected = true;
      });

      return;
    }

    const ref = db as pg.Client;
    ref.on("end", () => {
      this.dbConnected = false;
    });
  }
  ```

packages/llamaindex/src/vector-store/PGVectorStore.ts
AndreMaz: At the end of the day I just want to have the ability to customize my PG client (e.g, pass certs) and to configure the PGVectorStore's `dimensions` and etc.

So please give a guide on what's the way that you want to go

himself65:
> However, personally I'm not a big fan of the idea that I need to pass an already connected client. I would rather just provide a fully configured client and then leave the connection, re-connection and etc. logic to the `PGVectorStore`.

My original idea is to have more compatable with more db provider like `nestjs + prisma`, so in this case it obslutely already coonected before you call llamaindex pg vector store. So I'm thinking to make vector store to be a wrapper of something else, not just a standalone class for simple use.

himself65:
@AndreMaz I have updated the code, please give some feedback!

AndreMaz:
Hi @himself65 yep, looks good to me. Thank you

himself65:
why building is pending...🤔

himself65:
will do it tmr

packages/llamaindex/e2e/node/vector-store/pg-vector-store.e2e.ts
himself65, line 1: some notes to me here:

1. node test runner won't helps you cleanup resources, so if there's any hanging promise, it will stuck the CI or you machine. But it's werid to me, I think it's a node.js bug or DX issue, at least there's no way to find it out why hangign happening... (maybe [this](https://www.npmjs.com/package/why-is-node-running) would helps)
2. run `await pgClient.query("CREATE EXTENSION IF NOT EXISTS vector");` before call `registerTypes`.
3. It's very tricky that this e2e has side effect on system and each other, the top one test case will create database, so the following test cases will skip some logics. I think we should cleanup db beforeEach?

PR #1231, opened 2024-09-19
fix: use Blob instead of File

himself65:


packages/llamaindex/e2e/node/vector-store/pg-vector-store.e2e.ts
```diff
@@ -1,19 +1,27 @@
+/* eslint-disable turbo/no-undeclared-env-vars */
+import { config } from "dotenv";
 import { Document, VectorStoreQueryMode } from "llamaindex";
 import { PGVectorStore } from "llamaindex/vector-store/PGVectorStore";
 import assert from "node:assert";
 import { test } from "node:test";
 import pg from "pg";
 import { registerTypes } from "pgvector/pg";
 
+config({ path: [".env.local", ".env", ".env.ci"] });
```
himself65, line 10: this is confusing, nodejs test runner actually doesn't read environment variable by default

PR #1237, opened 2024-09-20
fix: context not working in contextchatengine

### LINKED ISSUE #1228: ContextChatEngine does not use context in chat
Issue in 6.1 onwards.

When calling .chat() from a ContextChatEngine the responses clearly do not use the context. 

If I run .retrieve instead straight from the Index.asRetriever, context is returned just fine. Context is also returned okay if I call `.contextGenerator.generate(prompt) ` from the chat engine. It is just when calling .chat that the context does not seem to be included, judging by the response.

marcusschiesser:
fixes #1228

### DIFF
--- a/.changeset/cold-gifts-help.md
+++ b/.changeset/cold-gifts-help.md
@@ -1,0 +1,7 @@
+---
+"llamaindex": patch
+"@llamaindex/core": patch
+"@llamaindex/examples": patch
+---
+
+Fix context not being sent using ContextChatEngine
--- a/.changeset/cold-gifts-help.md
+++ b/.changeset/cold-gifts-help.md
@@ -1,7 +1,6 @@
 ---
 "llamaindex": patch
 "@llamaindex/core": patch
-"@llamaindex/examples": patch
 ---
 
 Fix context not being sent using ContextChatEngine
--- a/packages/core/src/memory/chat-memory-buffer.ts
+++ b/packages/core/src/memory/chat-memory-buffer.ts
@@ -1,5 +1,5 @@
 import { Settings } from "../global";
-import type { ChatMessage, LLM, MessageContent } from "../llms";
+import type { ChatMessage, LLM } from "../llms";
 import { type BaseChatStore } from "../storage/chat-store";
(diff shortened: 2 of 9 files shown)


.changeset/cold-gifts-help.md
```diff
@@ -0,0 +1,7 @@
+---
+"llamaindex": patch
+"@llamaindex/core": patch
+"@llamaindex/examples": patch
```
himself65, line 4: remove this line

himself65 (changes requested):
could you please update test cases?

packages/core/src/memory/chat-memory-buffer.ts
```diff
@@ -34,7 +34,7 @@ export class ChatMemoryBuffer<
   }
 
   getMessages(
-    input?: MessageContent | undefined,
+    input?: ChatMessage<AdditionalMessageOptions>[] | undefined,
```
himself65, line 37: this should be `MessageContent` actually, according to python side API

himself65: should we implement this function in context engine?

https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/chat_engine/context.py#L151

himself65:
https://github.com/run-llama/LlamaIndexTS/pull/1237#discussion_r1768115702

packages/core/tests/memory/chat-memory-buffer.test.ts
```diff
@@ -0,0 +1,74 @@
+import { Settings } from "@llamaindex/core/global";
+import type { ChatMessage } from "@llamaindex/core/llms";
+import { ChatMemoryBuffer } from "@llamaindex/core/memory";
+import { beforeEach, describe, expect, test } from "vitest";
+
+describe("ChatMemoryBuffer", () => {
+  beforeEach(() => {
+    // Mock the Settings.llm
```
himself65, line 8: btw, We should we have a mock llm class?

himself65 (approved):
Im gonna merge this first, but I think should follow api of python version in the future

PR #1240, opened 2024-09-22
fix: update `PostgresKVStore` constructor params

gorango:
Allow initializing with pg client or pool instance (just like the PGVectorStore).

### DIFF
--- a/.changeset/fluffy-apes-guess.md
+++ b/.changeset/fluffy-apes-guess.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+fix: update `PostgresKVStore` constructor params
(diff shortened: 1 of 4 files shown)


gorango:
Also, since Postgres can store raw JSON, I update the `docStore.utils` handlers for `docToJson` and `jsonToDoc` to allow skipping the calls to JSON parse & stringify when handling doc data in the `PostgresDocumentStore`.

himself65 (commented):
I like this PR but I will consider the usage of `nativeJson` here

gorango:
I can revert those particular changes as they not directly unrelated to this PR.

And I can open a separate PR for the JSON bit. My rationale is:
- storing the raw JSON allows querying in pg and
- there are performance benefits in avoiding the additional steps involved in parse/stringify especially with large objects.

himself65 (approved):

PR #1243, opened 2024-09-22
fix: add `serializer` in doc store

gorango:
- allow querying documents `__data__` in `pg` by storing plain JSON
- avoid performance costs involved with `parse`/`stringify`


himself65:
I'm thinking whether to have a more general interface like `serializable`

gorango:
Agreed, that's a much cleaner solution. Was just about to push some similar updates but thank you for taking care of it!

himself65 (approved):
thank u

PR #1246, opened 2024-09-23
experimental: stream ui

himself65:
stream UI with RSC

```tsx
export const chatWithAI = () => {
  "use server"
  const { ui } = await run(() => {

  }, {
    on: {
      'tool-call': (props) => 'tool call'
      'tool-result': (props) => 'tool result' + props.data.result
    }
  })

  return ui
}
```

### DIFF
--- a/packages/ui/examples/next-openai/src/app/stream-ui/actions.tsx
+++ b/packages/ui/examples/next-openai/src/app/stream-ui/actions.tsx
@@ -1,0 +1,136 @@
+import { CoreMessage, generateId } from 'ai';
+import {
+  createAI,
+  createStreamableValue,
+  getMutableAIState as $getMutableAIState,
+  streamUI,
+} from 'ai/rsc';
+import { Message, BotMessage } from './message';
+import { z } from 'zod';
+import { run } from '@llamaindex/ui/rsc'
+import { OpenAI, OpenAIAgent } from 'llamaindex';
+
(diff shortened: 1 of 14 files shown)


packages/ui/src/rsc.tsx
```diff
@@ -0,0 +1,61 @@
...
+	}>
+}
+
+export type UIOptions = {
+	initial?: ReactNode
+	on: UIMap
+}
+
+export function run<T>(
+	fn: () => Promise<T>,
```
himself65, line 24: my idea is we have event manager, so we can listen the event and show the loading/result UI to the client side

What do u think

@marcusschiesser

packages/ui/examples/next-openai/src/app/stream-ui/actions.tsx
```diff
@@ -0,0 +1,136 @@
...
+			{
+				name: 'get_current_weather',
+				description: 'Get the current weather',
+				parameters: z.object({
+					location: z.string(),
+				}),
+			},
+    ]
+  })
+  const result =  run(() => {
```
marcusschiesser, line 53: @himself65, how would that from the caller side look like? I didn't look into the problem, but I think from the DX we're closest to OpenAI's assistants: https://sdk.vercel.ai/docs/ai-sdk-ui/openai-assistants maybe we can provide our own `useAssistants` function?

PR #1248, opened 2024-09-24
feat: support `npm:postgres`

himself65:


himself65 (commented):
forward to some our previous contributors see if like/dislike this change

/cc @gorango @AndreMaz

packages/core/src/vector-store/index.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * should compatible with npm:pg and npm:postgres
```
himself65, line 2: I think in the future, there could have sqlite vector store, so type should also comptatable with node:sqlite sqlite3 better-sqlite3

himself65: https://github.com/run-llama/LlamaIndexTS/pull/1184

AndreMaz:
This is just personal but what I would do things in a slightly different way.

Instead of `function fromPostgres()` and `function fromPG()` I would create a dir called `adapters` at `src/vector-store` and then create a `base.ts` adapter with the `connect()`, `disconnect()`, `query()` and etc. methods. 

Then I would create the `pg.ts` that would extend the `base.ts` and implement the logic in there. Same for the `postgres` and other clients.

After this the `PGVectorStore` would have an reference to the adapter class and then call it as needed.

I would also `export` the `base.ts` to allow the user to create their own version of adapter (`node:sqlite`, `sqlite3`,  `better-sqlite3` or any other) and pass it as a param to a PGVectorStore's `constructor()`. Essentially the user would be allowed to pass a custom `adapter` to the vector-store (as long as it correctly implements the methods of `base.ts`)


@himself65 what I said above is just a suggestion that I think will be easier to maintain in the long run.  Other than that your PR looks good :+1:

PR #1257, opened 2024-09-25
fix(core): backward support for legacy typescript

himself65:
Upstream: https://github.com/run-llama/LlamaIndexTS/pull/1256

### DIFF
--- a/packages/cloud/reader/package.json
+++ b/packages/cloud/reader/package.json
@@ -1,0 +1,7 @@
+{
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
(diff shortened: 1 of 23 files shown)


packages/cloud/reader/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "module",
```
himself65, line 2: this seems like a nodejs 18 issue, it will read cwd package.json type as file type(cjs/esm), instread of the node_modules package.json type

himself65: no, this is a expected behavior

PR #1264, opened 2024-09-26
refactor: workflow API

himself65:
See discussion: https://github.com/run-llama/LlamaIndexTS/discussions/1261

## checklist before landing

- [ ] validation API
- [ ] more example
  - [ ] websocket
  - [ ] http
  - [ ] database persistence
- [ ] update document


packages/core/tests/workflow.test.ts
```diff
@@ -113,15 +113,16 @@ describe("Workflow", () => {
...
-    const jokeFlow = new Workflow({ verbose: true, validate: true });
-    jokeFlow.addStep(StartEvent, generateJoke, { outputs: StopEvent });
-    jokeFlow.addStep(JokeEvent, critiqueJoke, { outputs: StopEvent });
-    const run = jokeFlow.run("pirates");
-    await expect(run).rejects.toThrow(
-      "The following events are consumed but never produced: JokeEvent",
-    );
-  });
+  // fixme: design a better API for validation
+  // test("workflow validation", async () => {
```
himself65, line 117: This is seems like a helper function for in/out Event relationship.

But should we check this in runtime, instead of top of the calling run

packages/core/src/workflow/context.ts
```diff
@@ -1,31 +1,77 @@
...
   #verbose: boolean = false;
 
-  constructor(params: { workflow: Workflow; verbose?: boolean }) {
-    this.#workflow = params.workflow;
-    this.#verbose = params.verbose ?? false;
+  #getStepFunction(event: WorkflowEvent): StepFunction<Start> | undefined {
+    return this.#eventMap.get(event.constructor as EventTypes);
   }
 
-  set(key: string, value: any): void {
```
himself65, line 18: I removed the globals here. because I think it's not really used in the code. Im thinking wheter to support this by adding `extend` syntax

packages/core/src/workflow/context.ts
```diff
@@ -1,20 +1,81 @@
...
+    }
     this.#verbose = params.verbose ?? false;
+
+    // push start event to the queue
+    const step = this.#getStepFunction(this.#startEvent);
+    assertExists(step, "No step found for start event");
+    this.#pushEvent(this.#startEvent, step);
   }
 
+  // these two API is not type safe
```
logan-markewich, line 78: since get/set can can be used to store anything, I don't think it can be type safe?

packages/core/src/workflow/context.ts
logan-markewich: I don't see globals removed? But in any case, the get/set api (which stores stuff into globals) is a pretty useful api for sharing data between steps

packages/core/src/workflow/workflow.ts
```diff
@@ -53,169 +40,15 @@ export class Workflow {
...
-
-    const context = this.#contexts.values().next().value;
-    if (!context) {
-      throw new Error("No active context found for streaming events.");
-    }
-
-    yield* context.streamEvents();
-  }
-
-  validate(): void {
```
logan-markewich, line 81: why remove the validate feature?

packages/core/src/workflow/workflow.ts
```diff
@@ -53,169 +40,15 @@ export class Workflow {
...
-    const unconsumedEvents = Array.from(consumedEvents).filter(
-      (event) => !producedEvents.has(event),
-    );
-    if (unconsumedEvents.length > 0) {
-      const names = unconsumedEvents.map((event) => event.name).join(", ");
-      throw new Error(
-        `The following events are consumed but never produced: ${names}`,
-      );
-    }
+  run(event: StartEvent<Start> | Start): Context<Start> {
```
marcusschiesser, line 43: i like that idea that we return a Context calling `run`

marcusschiesser: when I run i also would like to know the type of the stop event, see https://github.com/run-llama/create-llama/blob/ef070c0b4befd62cae3d50295e89ffdbe41ac854/templates/components/multiagent/typescript/nextjs/route.ts#L34

packages/core/src/workflow/events.ts
```diff
@@ -1,12 +1,14 @@
...
   data: T;
 
   constructor(data: T) {
     this.data = data;
+    this.displayName = this.constructor.name;
   }
 
   toString() {
-    return `${this.constructor.name}(${JSON.stringify(this.data)})`;
+    return this.displayName;
```
marcusschiesser, line 11: yes this is better

packages/core/src/workflow/context.ts
```diff
@@ -62,50 +122,140 @@ export class Context {
...
+  #signal: AbortSignal | null = null;
+  get #iteratorSingleton(): AsyncGenerator<WorkflowEvent, void, void> {
+    if (this.#iterator === null) {
+      this.#iterator = this.#createStreamEvents();
     }
-    return undefined;
+    return this.#iterator;
   }
 
-  writeEventToStream(event: WorkflowEvent): void {
```
marcusschiesser, line 94: This seems different:
The original code only sends explicitly sent events (via writeEventToStream) to streamEvents

marcusschiesser: that's why there's the streamingQueue

packages/core/src/workflow/context.ts
```diff
@@ -62,50 +122,140 @@ export class Context {
...
+  async *#createStreamEvents(): AsyncGenerator<WorkflowEvent, void, void> {
     while (true) {
-      const event = this.#streamingQueue.shift();
+      const event = this.#queue.shift();
       if (event) {
         yield event;
+        // event handling
+        const step = this.#getStepFunction(event);
+        assertExists(step, `No step found for event ${event.displayName}`);
+        const nextEvent = await step.call(null, this, event);
```
marcusschiesser, line 160: does it support receiving the same event for multiple steps?

marcusschiesser: (and executing the steps in parallel?)

packages/core/src/workflow/context.ts
himself65: remove this now. you can customize your context data to have such queue.

packages/core/src/workflow/workflow.ts
himself65: added!

packages/core/src/workflow/context.ts
himself65: in the new implemenetation we have multiple run support

packages/core/src/workflow/context.ts
```diff
@@ -1,111 +1,298 @@
...
+          const acceptableInputs: WorkflowEvent[] =
+            this.#pendingInputQueue.filter((event) =>
+              inputs.some((input) => event instanceof input),
+            );
+          const events: WorkflowEvent[] = [event, ...acceptableInputs];
+          assertExists(
+            events.length === inputs.length,
+            `Expect ${inputs.length} inputs, but got ${events.length}`,
+          );
+          const nextEvent: WorkflowEvent = await step.call(
```
marcusschiesser, line 174: blocking other steps (see discussion in slack)

packages/core/src/workflow/context.ts
```diff
@@ -1,111 +1,510 @@
...
+  #pendingInputQueue: WorkflowEvent[] = [];
+
+  /**
+   * Stream events from the start event
+   *
+   * Note that this function will stop once there's no more future events,
+   *  if you want stop immediately once reach a StopEvent, you should handle it in the other side.
+   * @private
+   */
+  #createStreamEvents(): AsyncIterableIterator<WorkflowEvent> {
```
marcusschiesser, line 239: 1. can we reduce the complexity of this function?
2. how can we implement to limit the number of promises per step - in python, that's: `@step(num_workers=4)`

examples/workflow/04_complex_workflow.ts
```diff
@@ -34,89 +34,61 @@ const truncate = (str: string) => {
...
-  const spec = context.get("specification");
+  await context.put("specification", ev.data.input);
+  const spec = await context.get("specification");
   // write a message to send an update to the user
-  context.writeEventToStream(
-    new MessageEvent({
-      msg: `Writing app using this specification: ${truncate(spec)}`,
-    }),
-  );
+  console.log(`Writing app using this specification: ${truncate(spec)}`);
```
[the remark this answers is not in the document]
himself65: I can expose sendEvent to context api, but I'm not sure if it's oky for seriazlize

examples/workflow/04_complex_workflow.ts
```diff
@@ -34,89 +34,61 @@ const truncate = (str: string) => {
...
+    console.log(`Reviewer says: ${review}`);
     return new StopEvent({ result: code });
   }
 
   return new ReviewEvent({ review, code });
 };
 
-const codeAgent = new Workflow({ validate: true });
+const codeAgent = new Workflow().with(kvContext);
 codeAgent.addStep(StartEvent, architect, { outputs: CodeEvent });
```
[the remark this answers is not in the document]
himself65: Yeah, I like this way

himself65 (commented):
I'm gonna open a new PR from this. This is hard to merge

himself65:
Some ideas since last month

1. Cloudflare also has a JS workflow API https://developers.cloudflare.com/workflows/build/workers-api/, it designed not only for llm
2. Workflow should unrelated to LLM actually, so somehow it could decoupled to `@llamaindex/workflow` or something
3. Some web hoster has some native API to wait for the event, I think we should allow a way to intergate with that, like [vercel waitUntil](https://vercel.com/docs/functions/functions-api-reference#waituntil) [cloudflare waituntil](https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil)

logan-markewich:
@himself65 that definitely makes sense

for 3, you can kind of imagine those as different workflow "runtimes" -- by default, its just using the JS event loop. But that could be replaced by another "runtime" like verbal or cloudflare (we had similar ideas for the python version, but for now, the runtime there is just a single async event loop)

himself65:
already landed in another, closing

PR #1270, opened 2024-09-30
text: add cycle test

marcusschiesser:


himself65 (approved):
thanks for improving coding strongability by adding test cases

PR #1272, opened 2024-09-30
feat: add filters for pinecone

thucpn:
Based on Python filters:
https://github.com/run-llama/llama_index/blob/a620a2661faabb49ba2f257bff7ae2ac04d0c12b/llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py#L132-L162

### DIFF
--- a/.changeset/breezy-donkeys-buy.md
+++ b/.changeset/breezy-donkeys-buy.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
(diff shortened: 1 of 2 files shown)


marcusschiesser (approved):

marcusschiesser:
@himself65 merging as needed for using pinecone with LITS

PR #1281, opened 2024-10-01
fix agent not saving chat history for regular msg (without tool-calli…

irevived1:
…ng) when streaming


packages/core/src/agent/utils.ts
```diff
@@ -67,6 +67,20 @@ export async function stepToolsStreaming<Model extends LLM>({
   // otherwise, it's a regular message
   const hasToolCall = !!(value.options && "toolCall" in value.options);
 
+  if (!hasToolCall) {
+    let content = value.delta;
+    for await (const chunk of pipStream) {
+      content += chunk.delta;
+    }
+    step.context.store.messages = [
```
himself65, line 75: I was thinking message is only for input, we won't update

lisbeth-santo: Hi, when we set stream: false, we do store the output in the messages array.
Do we plan to remove output msg from stepTools (non-streaming) for consistency?
see: https://github.com/run-llama/LlamaIndexTS/blob/805a3f3710e3cf4500c7388ab0adec3f7ebdf778/packages/core/src/agent/utils.ts#L153

Do you have a recommended function/property to get the output?
Thanks!

himself65: Oh, I remmeber why we cannot store the message when stream: true, because if you consume the stream, user will lost the delta. We can only pass to the user and only when they consumed we can store here. So I think the best practice should be handled on inner module, not here

irevived1: Thanks for the explanation.
Would you be able to do that in the inner module? I can close the current PR.
Or I can update it too if you can point me to the filename/location. TY

irevived1: Looks like context is already assigned at 
https://github.com/run-llama/LlamaIndexTS/blob/449274ca5ae945099fe167e650f67b67d30aaadf/packages/core/src/agent/base.ts#L79
before controller.close() at 
https://github.com/run-llama/LlamaIndexTS/blob/main/packages/core/src/agent/utils.ts#L56

any context assignment after is not saved.
Do you have other suggestions?

himself65:
Hi, just let you now I fixed this issue in #1456, basically I put the logic into the `createTask` API instead of the utils

PR #1282, opened 2024-10-01
fix: ensure `id_` is correctly passed during creation

### LINKED ISSUE #1278: Index record id not setting in Pinecone
I have been trying to get this to work all day and not sure if it is me or a bug.

I am setting id_ in a new Document like so

`new Document({
        id_: `${fileId}-${index}`,
        text: page.text,
        metadata: { page: page.page }
    })`

When it gets pushed to Pinecone it is saving this id to the metadata but creating a random UUID as the record id instead.

Screenshot 2024-09-30 at 23 40 26

LawrenceGB:
Fixes: #1278

This PR addresses issue #1278 where the id_ field was not being correctly passed as the id when creating a TextNode. This caused the upsert operation to the vector database to rely on the idGenerator instead of using the provided document ID, if available.

Changes:

Updated the logic for creating TextNode to ensure the id_ is set as the document ID when available.
This ensures that when upserting to the vector database, the correct document ID is used, avoiding the generation of unnecessary random IDs.

This fix improves consistency in ID handling, ensuring that document IDs are properly retained in the vector database.

### DIFF
--- a/packages/core/src/schema/node.ts
+++ b/packages/core/src/schema/node.ts
@@ -479,7 +479,7 @@ export function buildNodeFromSplits(
     ) {
       const imageDoc = doc as ImageNode;
       const imageNode = new ImageNode({
-        id_: idGenerator(i, imageDoc),
+        id_: imageDoc.id_ ?? idGenerator(i, imageDoc),
         text: textChunk,
         image: imageDoc.image,
         embedding: imageDoc.embedding,
--- a/packages/core/src/schema/node.ts
+++ b/packages/core/src/schema/node.ts
@@ -496,7 +496,7 @@ export function buildNodeFromSplits(
     ) {
       const textDoc = doc as TextNode;
       const node = new TextNode({
-        id_: idGenerator(i, textDoc),
+        id_: textDoc.id_ ?? idGenerator(i, textDoc),
         text: textChunk,
         embedding: textDoc.embedding,
         excludedEmbedMetadataKeys: [...textDoc.excludedEmbedMetadataKeys],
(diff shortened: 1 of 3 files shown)


packages/core/src/schema/node.ts
```diff
@@ -496,7 +496,7 @@ export function buildNodeFromSplits(
     ) {
       const textDoc = doc as TextNode;
       const node = new TextNode({
-        id_: idGenerator(i, textDoc),
+        id_: textDoc.id_ ?? idGenerator(i, textDoc),
```
himself65, line 499: LGTM, could you also update other node creation parts?

LawrenceGB: The issue is in buildNodeFromSplits as far as I can see, I have added imageNode id also

himself65 (commented):
LGTM overall, but why test is failing... interesting

marcusschiesser:
@LawrenceGB and @himself65, this PR was causing us a little headache:

A document is parsed into several nodes. This PR causes all these nodes to have the same ID as the parent. So we get duplicate IDs, and these IDs are no longer UUIDs (which is required by some vector DBs). 

I, therefore, reverted this PR in https://github.com/run-llama/LlamaIndexTS/pull/1311 

To get the ID of the source document from a split node, please check `node.relationships[NodeRelationship.SOURCE]`

LawrenceGB:
@marcusschiesser Could you please help clarify how I can get the actual ID that is being set as I can't seem to get it working.

I am creating a unique document like so, it's not using that ID but generates the random one when upserting to Pinecone and also the doc returned relationships are empty.

```
new Document({
        id_: `${fileId}-${index}`,
        text: page.text,
        metadata: { page: page.index }
    })
```

Thanks!

PR #1289, opened 2024-10-03
feat: add metadata filters for vector stores

thucpn:
Implement metadata filters for vector stores (based on Python code)
- Qdrant filters: https://github.com/run-llama/llama_index/blob/2068d14732894962bc1c21dbf5cfe07e7f39f2f7/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py#L1122

### DIFF
--- a/.changeset/yellow-birds-protect.md
+++ b/.changeset/yellow-birds-protect.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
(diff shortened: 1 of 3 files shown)


marcusschiesser (commented):
@thucpn looks good, please test and then we can release it

marcusschiesser (approved):

PR #1294, opened 2024-10-03
fix: lazy load openai

himself65:
feedback from Okta. Now we set Settings.llm to OpenAI by default and lazy loading the OpenAI

### DIFF
--- a/.changeset/afraid-pots-raise.md
+++ b/.changeset/afraid-pots-raise.md
@@ -1,0 +1,8 @@
+---
+"llamaindex": patch
+"@llamaindex/groq": patch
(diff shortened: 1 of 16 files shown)


parhammmm:
What was the reasoning behind this?

himself65:
> What was the reasoning behind this? @himself65

I put Setting.llm = new OpenAI on top, It will read the open api key at very begining so I would delay the initlization

parhammmm:
@himself65 you mean so that on initial load the env variable is ignored (maybe it's not available on initial load?) until the llm is invoked? that makes sense. It was causing some issues on our side but figured out a work around

Issue #1299, opened 2024-10-04
Inserting points to QdrantVectorStore fail because of wrong Point ID format


thucpn:
In QdrantVectorStore, the format of PointID should be 64-bit unsigned integers or UUID. Document: https://qdrant.tech/documentation/concepts/points/#point-ids

In LITS, we are using node id (eg: `./data/andrei.txt_1`) as the PointID when inserting to Qdrant (Python also do the same):
https://github.com/run-llama/LlamaIndexTS/blob/b7382b0d24847d5a82888a04c2acbe467a3f0ac5/packages/llamaindex/src/vector-store/QdrantVectorStore.ts#L166-L169

As a result, Qdrant Client will throw this error:
```
Format error in JSON body: value ./data/andrei.txt_1 is not a valid point ID, valid values are either an unsigned integer or a UUID
```

We can reproduce above issue with this simple example:
```
import {
  Document,
  QdrantVectorStore,
  storageContextFromDefaults,
  VectorStoreIndex,
} from "llamaindex";

async function generate() {
  const documentID = "./data/andrei.txt_1"; // replace by UUID will work. Eg: crypto.randomUUID()
  const qdrantVs = new QdrantVectorStore({}); // input your settings
  const docs = [new Document({ id_: documentID, text: "Content" })];
  const ctx = await storageContextFromDefaults({ vectorStore: qdrantVs });
  await VectorStoreIndex.fromDocuments(docs, { storageContext: ctx });
  console.log("Done generating");
}

void generate();
```

PR #1301, opened 2024-10-04
fix: generate uuid when inserting to Qdrant

thucpn:
Fix issue: https://github.com/run-llama/LlamaIndexTS/issues/1299

### DIFF
--- a/.changeset/tame-carrots-drop.md
+++ b/.changeset/tame-carrots-drop.md
@@ -1,0 +1,6 @@
+---
+"@llamaindex/core": patch
+"llamaindex": patch
+---
+
+fix: generate uuid when inserting to Qdrant


packages/env/src/index.ts
```diff
@@ -33,6 +33,23 @@ export function createSHA256(): SHA256 {
   };
 }
 
+export function randomUUIDFromString(input: string) {
```
marcusschiesser, line 36: 1. the idea of env is to provide env specific implementations, here we're only providing the node-specific, so the non-node is missing
2. if we use `createSHA256()`, then we can move this to core (in `utils/uuid.ts` file) as createSHA256 is provided for all envs. 

isn't the result always the same? then it shouldn't be `random...`:
```suggestion
export function UUIDFromString(input: string) {
```

packages/env/src/index.ts
```diff
@@ -33,6 +33,23 @@ export function createSHA256(): SHA256 {
...
+  // Create a hash using SHA-1
+  const hash = createHash("sha1").update(input).digest("hex");
```
marcusschiesser, line 38:
```suggestion
  const hash = createSHA256().update(input).digest("hex");
```
check if this still generates a UUID and add test cases

packages/env/src/index.ts
```diff
@@ -21,14 +21,16 @@ import { createWriteStream, fs } from "./fs/node.js";
 import "./global-check.js";
 import type { SHA256 } from "./node-polyfill.js";
 
-export function createSHA256(): SHA256 {
+export function createSHA256(
```
marcusschiesser, line 24: need to also change node-polyfill.ts

.changeset/clean-items-tickle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@llamaindex/core": patch
+"@llamaindex/env": patch
+"llamaindex": patch
+"@llamaindex/core-tests": patch
```
marcusschiesser, line 5: not needed
```suggestion
```

marcusschiesser (approved):

marcusschiesser:
@peec yes there was a bug with the node IDs and this PR was barking at the wrong tree - is now fixed in Release 0.6.18

PR #1323, opened 2024-10-11
feat: ChromaDb metadata filters

thucpn:
Based on Python implement:
https://github.com/run-llama/llama_index/blob/a0bc834321b718c52ab449957abcb3afa7a4b548/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py#L59

### DIFF
--- a/.changeset/soft-papayas-care.md
+++ b/.changeset/soft-papayas-care.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+feat: ChromaDb metadata filters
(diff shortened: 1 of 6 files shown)


.changeset/soft-papayas-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+feat: ChromaDb metadata filters
```
marcusschiesser, line 5:
```suggestion
feat: Add metadata filters to ChromaDb and update to 1.9.2
```

marcusschiesser (approved):

PR #1327, opened 2024-10-13
fix: specify conditions for react-server

himself65:


himself65:
sadly this is not working, will do separate the packages then try again later

PR #1344, opened 2024-10-18
refactor: move llm into single packages

himself65:
Repoted by @seldo that some people will see `no exported member` from llamaindex. This actually mean that in some environment we don't support some modules, so I'm gonna separate llm, then embedding, finally vector-store and doc store to double check the thrid-party modules whether can run in all js runtime.


logan-markewich:
hey @himself65, wanted to quickly clarify some stuff in case someone else stumbles upon this PR, so others can understand the issue here and what this change does.

1. I think @seldo reported reported that `huggingface is not a constructor` was the error a user was getting at the hackathon. Is that related to this change? If so, how?
2. Do we know in which environments do we not support which modules?
3. What does "separate" mean in this context? Turn them into new modules? Something else? 
4. By "double check" what did you mean?

himself65:
1. For example, the old behavior is `huggingface is not a constructor`, but ideally should be `Error: vercel edge runtime is not supported`. Because it doesn't support no matter how.
2. By default we export all, until it crash in some runtime, then we hide it
3. Will turn into different module to use node.js feature: conditional exports

For example,

```ts
import 'my-module'
``` 

```json
{
  "name": "my-module",
  "exports": {
    ".": "./index.js"
  }
}
```

Will be `index.js` no matter how

```json
{
  "name": "my-module",
  "exports": {
     ".": {
      "edge-light": "./error.js",
      ".": "index.js"
    }
  }
}
```

```ts
export class MyClass {
  constructor() {
    throw new Error('not supported in edge runtime')
  }
}
```

will be error.js in next.js with edge runtime. Finally, all thrid party modules (llm, embedding, vector-store) will be moved to under a npm package.

by "double check" means, export by default until someone report it's not working 😆

himself65:
`huggingface is not a constructor` is a case when we export by default but not export it in react server, the next.js webpack bundler will get a empty value with `['react-server']` import conditions

PR #1369, opened 2024-10-23
fix: unable to resolve unpdf in nextjs

thucpn:
fixes https://github.com/run-llama/create-llama/issues/384


himself65 (commented):
it should support load cjs

https://github.com/unjs/unpdf/blob/ca834860d9720830c4bfa4927f076cc2af432dfa/package.json#L29-L31

PR #1370, opened 2024-10-23
fix: replicate deps warning in nextjs

thucpn:
I got this warning related to `replicate` package when using it in NextJs create-llama template
 
 ```
 ⚠ ../../../LlamaIndexTS/node_modules/.pnpm/replicate@1.0.1/node_modules/replicate/lib/util.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

Import trace for requested module:
../../../LlamaIndexTS/node_modules/.pnpm/replicate@1.0.1/node_modules/replicate/lib/util.js
../../../LlamaIndexTS/node_modules/.pnpm/replicate@1.0.1/node_modules/replicate/index.js
../../../LlamaIndexTS/packages/providers/replicate/dist/index.js
../../../LlamaIndexTS/packages/llamaindex/dist/llm/replicate_ai.js
../../../LlamaIndexTS/packages/llamaindex/dist/llm/index.js
../../../LlamaIndexTS/packages/llamaindex/dist/index.edge.js
../../../LlamaIndexTS/packages/llamaindex/dist/index.react-server.js
```

### DIFF
--- a/packages/llamaindex/src/next.ts
+++ b/packages/llamaindex/src/next.ts
@@ -43,7 +43,7 @@ export default function withLlamaIndex(config: any) {
     const externals: Record<string, string> = {
       "onnxruntime-node": "commonjs onnxruntime-node",
       sharp: "commonjs sharp",
-      chromadb: "commonjs chromadb",
+      chromadb: "chromadb",
       unpdf: "unpdf",
     };
 
--- a/packages/llamaindex/src/next.ts
+++ b/packages/llamaindex/src/next.ts
@@ -38,7 +38,8 @@ export default function withLlamaIndex(config: any) {
       "onnxruntime-node": "commonjs onnxruntime-node",
       sharp: "commonjs sharp",
       chromadb: "commonjs chromadb",
-      unpdf: "commonjs unpdf",
+      unpdf: "unpdf",
+      replicate: "commonjs replicate",
     });
     return webpackConfig;
   };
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by himself65 — COMMENTED** [review]
it not the real solution just pretend everything as `commonjs`. (also that one you mentioned)

Most of the issue you see is from our upstream, like this one https://github.com/replicate/replicate-javascript/issues/273

If you see such error next time, you can try https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages, https://nextjs.org/docs/app/api-reference/next-config-js/serverExternalPackages, instead of webpack since next.js will use turbopack in the future

**Comment by marcusschiesser** [comment]
@himself65 https://nextjs.org/docs/app/api-reference/next-config-js/serverExternalPackages can't be used for pnpm, see 
https://github.com/vercel/next.js/issues/43433

File: packages/llamaindex/src/next.ts [hunk 1]
```diff
@@ -24,22 +24,34 @@ export default function withLlamaIndex(config: any) {
...
       webpackConfig = userWebpack(webpackConfig);
     }
     webpackConfig.resolve.alias = {
       ...webpackConfig.resolve.alias,
       "@google-cloud/vertexai": false,
     };
+
+    // Disable modules that are not supported in vercel edge runtime
+    if (options?.nextRuntime === "edge") {
+      webpackConfig.resolve.alias["replicate"] = false;
```
marcusschiesser, line 38: @himself65 i guess an alternative would be to not export replicate from `llamaindex`?

himself65:
just double checked and unpdf is always empty, idk why

packages/llamaindex/src/next.ts
himself65: yeah

PR #1394, opened 2024-10-25
Implement context-aware agent functionality.

erik-balfe:
# Context-Aware Agent Implementation

This PR introduces context-awareness functionality for agents, inspired by the discussion in #1224.

## Implementation:

- Added `withContextAwareness` mixin in `contextAwareMixin.ts`
- Created `OpenAIContextAwareAgent` and `AnthropicContextAwareAgent`
- Updated relevant documentation and examples

## Motivation:

The goal is to allow agents to access indexed data without explicitly calling retriever tools, reducing latency in real-time scenarios especially with not so fast LMs. This approach is analogous to the ContextChatEngine, which automatically incorporates retriever data into the chat context.

## Approach:

The implementation uses a mixin pattern to add context retrieval and injection capabilities to existing agent classes. This method allows for flexibility and easy extension to different agent types.

Feedback on the implementation approach and any potential improvements is welcome.

##
Just recreated branch and PR (old one: #1235) because of conflicts in prev version. All modules were moved to core after that PR was been created. So its easier to create a new one.

### DIFF
--- a/packages/llamaindex/src/agent/anthropic.ts
+++ b/packages/llamaindex/src/agent/anthropic.ts
@@ -1,21 +1,7 @@
-import {
-  AnthropicAgent,
-  type AnthropicAgentParams,
-} from "@llamaindex/anthropic";
-import {
-  withContextAwareness,
-  type ContextAwareConfig,
-} from "./contextAwareMixin.js";
+import { AnthropicAgent } from "@llamaindex/anthropic";
+import { withContextAwareness } from "./contextAwareMixin.js";
 
-export type AnthropicContextAwareAgentParams = AnthropicAgentParams &
-  ContextAwareConfig;
-
-export class AnthropicContextAwareAgent extends withContextAwareness(
-  AnthropicAgent,
-) {
-  constructor(params: AnthropicContextAwareAgentParams) {
-    super(params);
-  }
-}
+export const AnthropicContextAwareAgent = withContextAwareness(AnthropicAgent);
+export type { ContextAwareConfig } from "./contextAwareMixin.js";
 
 export * from "@llamaindex/anthropic";
--- a/packages/llamaindex/src/agent/anthropic.ts
+++ b/packages/llamaindex/src/agent/anthropic.ts
@@ -1,1 +1,21 @@
+import {
+  AnthropicAgent,
+  type AnthropicAgentParams,
+} from "@llamaindex/anthropic";
+import {
+  withContextAwareness,
+  type ContextAwareConfig,
+} from "./contextAwareMixin.js";
+
+export type AnthropicContextAwareAgentParams = AnthropicAgentParams &
+  ContextAwareConfig;
+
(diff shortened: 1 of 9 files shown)


packages/llamaindex/src/agent/contextAwareMixin.ts
```diff
@@ -0,0 +1,80 @@
...
+  retrieveContext(query: MessageContent): Promise<string>;
+  injectContext(context: string): Promise<void>;
+}
+
+/**
+ * ContextAwareAgentRunner enhances the base AgentRunner with the ability to retrieve and inject relevant context
+ * for each query. This allows the agent to access and utilize appropriate information from a given index or retriever,
+ * providing more informed and context-specific responses to user queries.
+ */
+export function withContextAwareness<T extends Constructor<AgentRunner<LLM>>>(
```
himself65, line 29: why not just `T extends typeof AgentRunner<LLM>`

packages/llamaindex/src/agent/contextAwareMixin.ts
```diff
@@ -0,0 +1,80 @@
...
+  BaseClass: T,
+) {
+  return class extends BaseClass implements ContextAwareAgentRunner {
+    contextRetriever: BaseRetriever;
+    retrievedContext: string | null = null;
+
+    constructor(...args: any[]) {
```
himself65, line 36: I don't recommend any here, it might lose type hint in user side

himself65: should be `T extends typeof AgentRunner<LLM>, Args extends ConstructorParameter<T>`

and `constructor(...args: Args)`

packages/llamaindex/src/agent/anthropic.ts
```diff
@@ -1 +1,21 @@
...
+} from "@llamaindex/anthropic";
+import {
+  withContextAwareness,
+  type ContextAwareConfig,
+} from "./contextAwareMixin.js";
+
+export type AnthropicContextAwareAgentParams = AnthropicAgentParams &
+  ContextAwareConfig;
+
+export class AnthropicContextAwareAgent extends withContextAwareness(
```
himself65, line 13: same

himself65 (commented):
overall LGTM, just need some type improve

erik-balfe:
@himself65 About all the comments above.

There’s a TypeScript error: 
`A mixin class must have a constructor with a single rest parameter of type 'any[]'`.
(See the issue here (https://github.com/microsoft/TypeScript/issues/37142)).

I believe it’s currently impossible to implement this correctly with all the necessary types while extending base LLMAgent or AgentRunner classes.

The simple approach that should work doesn’t function as intended in TypeScript, unfortunately.

In latest version I have rewritten the code using a less general approach that works with OpenaiAgent and AnthropicAgent, along with some other workarounds, but without using 'any'.

himself65:
nice work

PR #1404, opened 2024-10-27
feat: add `@llamaindex/readers` package

himself65:
- separate readers module, enhance more js runtime DX
- allow user install `@llamaindex/readers` for smaller bundler size
- update BaseReader type for futuring use
- update corrsponding document
- put some unit test to top level for better file structure


himself65 (commented):
Revert https://github.com/run-llama/LlamaIndexTS/pull/1404/commits/f8aaff6319e9d1fba2b45fbabba2f0c01da7ffa3 when https://github.com/huozhi/bunchee/issues/589 get fixed

PR #1436, opened 2024-11-05
feat: add example using chat-ui components to docs

marcusschiesser:
Example using SimpleChatEngine and chat-ui components


himself65 (changes requested):
could you implement it under apps/next? It's also using next.js

himself65 (approved):

PR #1481, opened 2024-11-14
feat: chat UI with rsc example

thucpn:


apps/next/src/components/demo/chat/rsc/ai-action.tsx
```diff
@@ -0,0 +1,56 @@
+import { Markdown } from "@llamaindex/chat-ui/widgets";
+import { generateId, Message } from "ai";
+import { createAI, createStreamableUI, getMutableAIState } from "ai/rsc";
+import { ChatMessage, OpenAIAgent } from "llamaindex";
```
himself65, line 4: Do not use llamaindex for now, it has lots of dependencies

himself65 (changes requested):

apps/next/src/components/demo/chat/rsc/ai-action.tsx
himself65, line 1: im not sure, but we don't need really have a backend impletation for document, because it costs the real money from openai. We might need a FakeLLM and some fixed input options in user side in the future

packages/cloud/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "license": "MIT",
   "scripts": {
-    "generate": "./node_modules/.bin/openapi-ts",
+    "generate": "openapi-ts",
```
marcusschiesser, line 7: why this change?

himself65 (approved):

PR #1482, opened 2024-11-14
docs(next): fix stackbliz preview

himself65:


himself65:
I think this won't work. I will use JS sdk

PR #1485, opened 2024-11-14
feat: add Azure AI Vector Search

manekinekko:
wip


packages/llamaindex/src/vector-store/azure/AzureAISearchVectorStore.ts
```diff
@@ -0,0 +1,1264 @@
...
+  AzureQueryResultSearchDefault,
+  AzureQueryResultSearchHybrid,
+  AzureQueryResultSearchSemanticHybrid,
+  AzureQueryResultSearchSparse,
+} from "./AzureQueryResultSearch.js";
+
+/**
+ * Enumeration representing the supported index management operations
+ */
+export const enum IndexManagement {
```
manekinekko, line 54: When accessing `IndexManagement.CREATE_IF_NOT_EXISTS` in userland, we get the following error:

```
Type error: Cannot access ambient const enums when 'isolatedModules' is enabled.
```

@himself65 any idea how to fix this?

packages/llamaindex/src/vector-store/types.ts
```diff
@@ -18,6 +18,9 @@ export enum VectorStoreQueryMode {
   LINEAR_REGRESSION = "linear_regression",
   // maximum marginal relevance
   MMR = "mmr",
+
+  // for Azure AI Search
+  SEMANTIC_HYBRID = "semantic_hybrid",
```
manekinekko, line 23: When accessing `VectorStoreQueryMode.SEMANTIC_HYBRID` from userland, we get the following error:

```
Cannot read properties of undefined (reading 'SEMANTIC_HYBRID')
```

@himself65 is there a better way to add a new query mode?

manekinekko: Alright, then maybe that's just my local config. Thank you.

packages/llamaindex/src/vector-store/azure/AzureAISearchVectorStore.ts
himself65: You canno export `const enum` it's the rule by typescript

himself65 (approved):

manekinekko:
Thank you @logan-markewich  @himself65 for helping reviewing and merging this PR.

PR #1493, opened 2024-11-15
refactor: move mockLLM to core

thucpn:


packages/core/src/utils/llms.ts
```diff
@@ -143,3 +145,30 @@ export async function imageToDataUrl(
   }
   return await blobToDataUrl(input);
 }
+
+class MockLLM {
```
himself65, line 149: it's better to implement all interface from LLM

himself65 (changes requested):

himself65 (approved):

Issue #1508, opened 2024-11-19
PGVectorStore doesn't identify Vercel Postgres inside Server Actions


madarco:
**Describe the bug**
PGVectorStore constructor relies on `config.client.constructor.name.includes("Vercel")` to identify if the passed client is from `@vercel/postgres`. 
However in "edge" environment (ie: Server Actions), Vercel builds all the dependencies in a single file and wrap the methods in a .bind call. 
This makes this check on the constructor name fail. An Alternative could be to check if the class have specific methods (ie: connect, query, but not unsafe)

**To Reproduce**
Code to reproduce the behavior:

(needs to deploy this on Vercel to test)
```ts
function serverAction() {
  "use server"
  import { sql, db } from "@vercel/postgres";
  return new PGVectorStore({
      client: db,
    });
}
```

**Expected behavior**
PGVectorStore shouldn't give the error that the "unsafe" method is not defined (ie: is trying to use the fromPostgres instead of fromVercelPool)

**Screenshots**
none

**Desktop (please complete the following information):**

- OS: [e.g. macOS, Linux]
- JS Runtime / Framework / Bundler (select all applicable)
- [X] Next.js
- [X ] Others: only happens when deployed in Vercel cloud

PR #1509, opened 2024-11-19
fix: supports vercel bundling

madarco:
As described in https://github.com/run-llama/LlamaIndexTS/issues/1508

Inside a Server Action deployed on Vercel, the check done in the constructor of PGVectorStore fails to identify the vercel/posgres pool used.

In this change, I propose to check the presence or absence of some methods.
(Note that the methods also will be bond)

```ts
if (
    config.client.constructor.name.includes("Vercel") ||
    (!!(config.client as VercelPool).connect &&
      !!(config.client as VercelPool).query &&
      !(config.client as Sql).unsafe)
   )
```

### DIFF
--- a/.changeset/cuddly-geckos-clap.md
+++ b/.changeset/cuddly-geckos-clap.md
@@ -1,0 +1,5 @@
+---
+"llamaindex": patch
+---
+
+fix: supports Vercel bundling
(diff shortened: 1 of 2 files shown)


himself65 (approved):
Make sense, LGTM

PR #1517, opened 2024-11-21
fix: filter out empty chunks in SentenceSplitter

parhammmm:


himself65 (changes requested):
could you please add test for this?

himself65 (changes requested):
I didn't find the case then splits get empty, could you please add some tests

parhammmm:
@himself65 this was the case where it happened https://gist.github.com/parhammmm/c69e37420c92cb85db788454f2a9fe27

It's a bit of an edge case and I found a way to work around it. Will need to find some time to figure out Llamaindex's test suite and then will add it

PR #1518, opened 2024-11-21
feat: add Bedrock provider integration with embedding functionality

jaummm1:
The Bedrock embedding models have been added.


himself65 (commented):
LGTM but I think we have bedrock in community package, let me see if these are conflicts

PR #1528, opened 2024-11-28
feat: add AzureCosmosDBMongoVectorStore

crisjy:


crisjy:
**Add Vector Store for CosmosDB**

1. Support multiple operations for CosmosDB vCore such as query, delete, createIndex 
2. Support multiple index types for CosmosDB vCore.

PR #1544, opened 2024-12-03
fix: allow Node 18 again to make Stackblitz work

marcusschiesser:
Upstream issue: https://github.com/stackblitz/webcontainer-core/issues/560

### DIFF
--- a/packages/env/src/internal/multi-model/node.ts
+++ b/packages/env/src/internal/multi-model/node.ts
@@ -7,6 +7,13 @@ export {
 } from "./shared.js";
 
 export async function loadTransformers(onLoad: OnLoad) {
+  const nodeVersions = process.versions.node.split(".");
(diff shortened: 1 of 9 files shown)


packages/env/src/internal/multi-model/node.ts
```diff
@@ -7,6 +7,13 @@ export {
 } from "./shared.js";
 
 export async function loadTransformers(onLoad: OnLoad) {
+  const nodeVersions = process.versions.node.split(".");
```
himself65, line 10: this is not working, npm install will throw error before you do this

marcusschiesser: @himself65 hmm, on ci we're using pnpm and it works. what do you suggest to check for the node version at runtime instead?

himself65: npm wont work

Issue #1552, opened 2024-12-06
Compact Response Synthesizer does not return source nodes


GunnarHolwerda:
**Describe the bug**
The `'compact'` response synthesizer since `v.0.6.3` does not return the original source nodes that were provided to the prompt. This was _partially_ fixed in `v0.8.26` when the `'refine'` synthesizer began passing the source nodes back through.

But since the `'compact'` response synthesizer passes through the compacted text node through to the super class (`Refine`), the resulting `sourceNodes` on the engine response are a single TextNode that lost all context of the original nodes that were provided and used to build that text node.

**To Reproduce**
Code to reproduce the behavior:

```ts
import { getResponseSynthesizer, RetrieverQueryEngine } from 'llamaindex';
import { VectorIndexRetriever } from 'llamaindex/vector-store';
const queryEngine = new RetrieverQueryEngine(new VectorIndexRetriever(), getResponseSynthesizer('compact'));

const response = queryEngine.query({ query: 'hello world', stream: false });

// These will not be the original nodes that were returned from the VectorIndexRetriever. Instead it will be a TextNode or TextNodes.
console.log(response.sourceNodes);
```

**Expected behavior**
I would expect `sourceNodes` on the `EngineResponse` to hold the original `NodeWithScore` objects that were returned from the retriever.

**Screenshots**

**Desktop (please complete the following information):**
I'm on Mac, running this in NextJs, but it isn't platform specific.
Version `0.8.26`

**Additional context**

Here is the implementation of `CompactAndRefine` as it exists in `v0.8.26` with comments denoting the problem.
```typescript
class CompactAndRefine extends Refine {
  async getResponse(
    query: MessageContent,
    nodes: NodeWithScore[],
    stream: boolean,
  ): Promise<EngineResponse | AsyncIterable<EngineResponse>> {
    const textQATemplate: TextQAPrompt = this.textQATemplate.partialFormat({
      query: extractText(query),
    });
    const refineTemplate: RefinePrompt = this.refineTemplate.partialFormat({
      query: extractText(query),
    });
    const textChunks = nodes.map(({ node }) =>
      node.getContent(MetadataMode.LLM),
    );

    const maxPrompt = getBiggestPrompt([textQATemplate, refineTemplate]);
    const newTexts = this.promptHelper.repack(maxPrompt, textChunks);
    const newNodes = newTexts.map((text) => new TextNode({ text }));
    if (stream) {
      return super.getResponse(
        query,
[...]

PR #1554, opened 2024-12-07
fix: return actual source nodes with compact and refine response synt…

GunnarHolwerda:
### Description of Changes

Updates the `Refine` responseSynthesizer to return the `sourceNodes` in the response as the original nodes that were provided to the synthesizer. This persists the behavior of it prior to 0.5.20 and the other synthesizers where `EngineResponse.sourceNodes` is set to the value of the original source nodes that were provided. With the prior implementation it was returning the packed text chunks which would lose context of the document information.

Fixes: #1552

### DIFF
--- a/packages/core/src/response-synthesizers/factory.ts
+++ b/packages/core/src/response-synthesizers/factory.ts
@@ -80,6 +80,16 @@ class Refine extends BaseSynthesizer {
   async getResponse(
     query: MessageContent,
     nodes: NodeWithScore[],
+    stream: true,
+  ): Promise<AsyncIterable<EngineResponse>>;
+  async getResponse(
+    query: MessageContent,
+    nodes: NodeWithScore[],
+    stream: false,
+  ): Promise<EngineResponse>;
+  async getResponse(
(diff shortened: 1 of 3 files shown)


packages/core/src/response-synthesizers/factory.ts
```diff
@@ -77,6 +77,16 @@ class Refine extends BaseSynthesizer {
...
+  async getResponse(
+    query: MessageContent,
+    nodes: NodeWithScore[],
+    stream: true,
+  ): Promise<AsyncIterable<EngineResponse>>;
+  async getResponse(
+    query: MessageContent,
+    nodes: NodeWithScore[],
+    stream: false,
+  ): Promise<EngineResponse>;
```
GunnarHolwerda, line 89: Added these signatures so that I didn't have to do more type jumbling after grabbing the original response and transforming it. If this is not accurate please let me know, but based on my knowledge if `stream = true` it will always be an AsyncIterable.

himself65 (approved):

Issue #1564, opened 2024-12-12
Add Tool Call History Support for Anthropic Agent


erik-balfe:
## Feature Overview

Added proper message formatting for tool calls and results in chat history when using Anthropic agent. This makes Anthropic agent handle tool history consistently with OpenAI agent, which already had this functionality.

## Problem Solved

Previously, when providing chat history with tool calls to Anthropic agent, it would throw API errors about incorrect message format. The same chat history worked fine with OpenAI agent. This has been fixed by implementing correct message transformation for Anthropic's specific format requirements.

### Error Before Fix

When trying to use chat history with tool calls in Anthropic agent, the following error was thrown:

```
error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.1.content.1.tool_use.input: Input should be a valid dictionary"}}
      at new AnthropicError (path/to/project/node_modules/.pnpm/@anthropic-ai+sdk@0.32.1_encoding@0.1.13/node_modules/@anthropic-ai/sdk/error.mjs:7:9)
      at new APIError (path/to/project/node_modules/.pnpm/@anthropic-ai+sdk@0.32.1_encoding@0.1.13/node_modules/@anthropic-ai/sdk/error.mjs:86:9)
```

This happened because Anthropic's API expects tool calls in a specific format, different from the standard LlamaIndex format that worked with OpenAI.

### Chat History with Tool Calls (now works with both agents)

```typescript
const chatHistory: ChatMessage[] = [
  {
    role: "user",
    content: "What's the weather in London?",
  },
  {
    role: "assistant",
    content: "Let me check the weather.",
    options: {
      toolCall: [
        {
          id: "call_123",
          name: "weather",
          input: JSON.stringify({ location: "London" }),
        },
      ],
    },
  },
  {
    role: "tool",
    content: "The weather in London is sunny, +20°C",
    options: {
      toolResult: {
        id: "call_123",
      },
    },
  },
];
```

## Implementation Details

- Added proper message format transformation in Anthropic provider
- Maintains consistent chat history format across providers while handling provider-specific requirements internally
- Tool calls and results are now correctly recognized by Anthropic agent

## Testing

Added comprehensive tests to verify:

- Correct transformation of tool call messages for Anthropic format
- Proper handling of tool results
- Consistent behavior with OpenAI agent's existing functionality

## Usage Example

Here's a complete example demonstrating the fixed functionality:

```typescript
[...]

PR #1565, opened 2024-12-12
add tool call history support in chat messages

erik-balfe:
Add proper message format transformation for tool calls and results in chat history when using Anthropic agent. This allows Anthropic agent to handle tool history consistently with OpenAI agent, which already had this functionality.

Previously, chat history with tool calls would cause API errors with Anthropic but worked fine with OpenAI. Now both agents handle the same message format correctly.
Closes  https://github.com/run-llama/LlamaIndexTS/issues/1564

### DIFF
--- a/.changeset/pink-days-appear.md
+++ b/.changeset/pink-days-appear.md
@@ -1,0 +1,8 @@
+---
+"@llamaindex/anthropic": minor
+"@llamaindex/llamaindex-test": minor
+"llamaindex": minor
+"@llamaindex/core": minor
+---
+
+added support for tool calls with results in message history for athropic agent
--- a/.changeset/pink-days-appear.md
+++ b/.changeset/pink-days-appear.md
@@ -1,8 +1,7 @@
 ---
-"@llamaindex/anthropic": minor
-"@llamaindex/llamaindex-test": minor
-"llamaindex": minor
-"@llamaindex/core": minor
+"@llamaindex/anthropic": patch
+"llamaindex": patch
+"@llamaindex/core": patch
 ---
 
 added support for tool calls with results in message history for athropic agent
--- a/packages/core/src/llms/type.ts
+++ b/packages/core/src/llms/type.ts
@@ -54,7 +54,7 @@ export interface LLM<
   ): Promise<CompletionResponse>;
 }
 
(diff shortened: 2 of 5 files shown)


erik-balfe:
@marcusschiesser could you help me please? Is it something required from me or some permission issue?
In the checks here.
`Vercel – llamaindex-ts-doc — Authorization required to deploy`

packages/core/src/llms/type.ts
```diff
@@ -54,7 +54,7 @@ export interface LLM<
   ): Promise<CompletionResponse>;
 }
 
-export type MessageType = "user" | "assistant" | "system" | "memory";
+export type MessageType = "user" | "assistant" | "system" | "memory" | "tool";
```
marcusschiesser, line 57: @erik-balfe we currently don't use "tool", a tool message is one that has a `toolResult`, see the conversion for openai: https://github.com/run-llama/LlamaIndexTS/blob/9e648a0b4f670e131dc6f978872009f8939b2825/packages/providers/openai/src/llm.ts#L267

@himself65 i remember you added this

.changeset/pink-days-appear.md
```diff
@@ -0,0 +1,8 @@
+---
+"@llamaindex/anthropic": minor
+"@llamaindex/llamaindex-test": minor
+"llamaindex": minor
+"@llamaindex/core": minor
```
marcusschiesser, line 5: contract is not changing right? 
```suggestion
"@llamaindex/anthropic": patch
"llamaindex": patch
"@llamaindex/core": patch
```

marcusschiesser:
> @marcusschiesser could you help me please? Is it something required from me or some permission issue? In the checks here. `Vercel – llamaindex-ts-doc — Authorization required to deploy`

that's only for building the docs, which are not changed by this PR, so no worries

packages/core/src/llms/type.ts
erik-balfe: Ok, just removed it now

himself65: we put tool info into "user" and "assistant"

erik-balfe:
So are the any other problems with the PR? @himself65
I have fixed all these comments

himself65 (approved):

PR #1567, opened 2024-12-15
feat(vectorstore): add libsql vector store

notrab:
- [ ] Update ENV naming
- [ ] Adds some inline docs/`@example` for the table/index

### DIFF
--- a/packages/llamaindex/src/vector-store/LibSQLVectorStore.ts
+++ b/packages/llamaindex/src/vector-store/LibSQLVectorStore.ts
@@ -1,0 +1,283 @@
+import {
+  BaseVectorStore,
+  type MetadataFilters,
+  type VectorStoreBaseParams,
+  type VectorStoreQuery,
+  type VectorStoreQueryResult,
+} from "./types.js";
+
+import { createClient, type Client, type InStatement } from "@libsql/client";
+import {
+  MetadataMode,
+  type BaseNode,
+  type Metadata,
+  type TextNode,
+} from "@llamaindex/core/schema";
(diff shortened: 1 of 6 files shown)


marcusschiesser:
@notrab thanks for your contribution, can you move your change to a own package? see https://github.com/run-llama/LlamaIndexTS/pull/1605

marcusschiesser:
@notrab thanks a lot for your PR! We moved all the vector stores to their own npm package, can you do that please too? Postgres PR shows how that is done: https://github.com/run-llama/LlamaIndexTS/pull/1597

packages/llamaindex/src/vector-store/LibSQLVectorStore.ts
```diff
@@ -0,0 +1,283 @@
+import {
+  BaseVectorStore,
+  type MetadataFilters,
+  type VectorStoreBaseParams,
+  type VectorStoreQuery,
+  type VectorStoreQueryResult,
+} from "./types.js";
+
+import { createClient, type Client, type InStatement } from "@libsql/client";
```
himself65, line 9: further more, I want support also sqlite, sqlite3, node:sqlite, bun:sqlite...

notrab: This PR is focused on the local and remote libSQL instances. I'm happy to help with the sqlite variants, but in another PR. Does that work for you?

marcusschiesser: It's ok to start like this, but we need to put the store into its own npm package

PR #1588, opened 2025-01-03
fix: clean up docstore when generating embedding fail

thucpn:


packages/llamaindex/src/indices/vectorStore/index.ts
```diff
@@ -112,9 +113,20 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
...
-      });
+      try {
+        await index.buildIndexFromNodes(options.nodes, {
+          logProgress: options.logProgress,
+        });
+      } catch (error) {
+        // clean up doc store when generating embeddings fails
+        const { unusedDocs } = await classify(index.docStore, options.nodes);
+        for (const docId of unusedDocs) {
+          await index.docStore.deleteDocument(docId, false);
```
marcusschiesser, line 124: @thucpn i think this doesn't cleanup a general call of `insertNodes` (`buildIndexFromNodes` is a synonym). we should catch failed calls of `getNodeEmbeddingResults` and then do a clean up (if necessary, see below)

packages/llamaindex/src/indices/vectorStore/index.ts
```diff
@@ -335,15 +336,28 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
...
-    );
+    try {
+      const embeddedNodes = await this.getNodeEmbeddingResults(nodes, options);
+      await addNodesToVectorStores(
+        embeddedNodes,
+        this.vectorStores,
+        this.insertNodesToStore.bind(this),
+      );
+    } catch (error) {
+      await this.cleanDocStore(nodes);
```
marcusschiesser, line 347: don't understand that - if `getNodeEmbeddingResults` fails, we don't add any nodes to the store - why do we need to call `cleanDocStore` then?

marcusschiesser (approved):
looks good

Issue #606, opened 2024-03-05
set a system prompt/message for a queryengine


rawwerks:
i cannot for the life of me figure out how to set a system prompt while using index.asQueryEngine()

could someone point me in the right direction? this has to be possible, otherwise how could anyone use llamaindexts for RAG applications?

here's a related discord thread: https://discord.com/channels/1059199217496772688/1133167189860565033/1214393825473077248

PR #1606, opened 2025-01-25
feat: add system prompt support to ReplicateLLM and OpenAI

igorMSoares:
A way to define a system prompt when instantiating an LLM so the system prompt can be sent as a `system` role message.
The proposed solution currently implements this functionallity for the classes `ReplicateLLM` (llama3 family models), `OpenAI` and the ones that extend it, like `Groq` and `TogetherLLM`.

This might solve #606 , at least if the llm used in the QueryEngine is from such providers.

Would this approach be ok? Any other suggestions on how a system prompt could be provided to the llm?


marcusschiesser:
Thanks @igorMSoares - you can add system messages to LLM instances. I updated #606 for how to modify prompts for queryengines. What's your use case for passing one system prompt to an LLM?

PR #1618, opened 2025-02-04
remove service context

marcusschiesser:
- [x] removed service context
- [x] inline `xxxFromSettings` functions in `Settings.ts`
- [x] fix tests


packages/llamaindex/src/Settings.ts
```diff
@@ -163,42 +162,9 @@ class GlobalSettings implements Config {
...
-export const promptHelperFromSettingsOrContext = (
-  serviceContext?: ServiceContext,
-) => {
-  if (serviceContext?.promptHelper) {
-    return serviceContext.promptHelper;
-  }
-
-  return Settings.promptHelper;
-};
+export const llmFromSettings = () => Settings.llm;
```
marcusschiesser, line 165: can you inline these - instead of calling `llmFromSettings()` in the code call `Settings.llm`

marcusschiesser:
looks good, not merging yet as it will trigger 0.9 release

PR #1620, opened 2025-02-05
fix: missing condition to stringify tool input

thucpn:


.changeset/shiny-maps-thank.md
```diff
@@ -0,0 +1,7 @@
+---
+"llamaindex": patch
```
marcusschiesser, line 2: do we need to release llamaindex?

packages/llamaindex/tests/llm/index.test.ts
```diff
@@ -292,6 +292,45 @@ describe("Message Formatting", () => {
       );
     });
 
+    test("Anthropic throws error for invalid tool input", () => {
```
marcusschiesser, line 295: can we move these tests to anthropic package?

PR #1629, opened 2025-02-06
fix: delete temp competely from params when calling openai o3

parhammmm:
This is a follow up from the previous fix.

The previous fix stopped working for an unknown reason but this one for sure fixes the issue @marcusschiesser


marcusschiesser (approved):

marcusschiesser:
@parhammmm thanks - can you resolve the conflict?

parhammmm:
> @parhammmm thanks - can you resolve the conflict?

done

PR #1654, opened 2025-02-13
refactor(agent): changed regex for tool use extraction to look for optional "Input:" prefix to comply with Cohere command-r models

nea:
Ahoy

Referring to https://github.com/run-llama/LlamaIndexTS/discussions/1652 to make agent tool use extraction more flexible and also include models by [Cohere](https://cohere.com/) from their `command-r` line.

Thank you for your consideration.


marcusschiesser:
Thanks @nea - did you test that also with other models?

PR #1677, opened 2025-02-23
Fix refine synthesizer empty source nodes behavior

GunnarHolwerda:
Previously, if you passed an empty array of source nodes to the Refine response synthesizer `getResponse` method. it would end up throwing an internal error because `response` would be left as undefined. When that was passed to `streamConverter` it would then throw an error in the `for await` loop.

I experienced this intermittent in my development environment or when testing with no data, and it's possible someone could experience this, so I wanted to have the behavior defined rather than throw an error. 

The new behavior is that it defaults the response string to `""` if sourceNodes is provided as an empty array.


marcusschiesser (approved):
thanks @GunnarHolwerda

PR #1693, opened 2025-02-28
feat: additional tool argument

tvanier:
A proposal to add an optional `toolArg` to `AgentTaskContext`, to be passed as an extra argument when calling a tool.
I am discovering LlamaIndex, there might be a better way to implement this feature, happy to refactor. And add tests once the team agrees ;-)

Example with a tool requiring some user token:
```ts
// tool takes LLM input and additional argument from task context
const queryKnowledgeBase = async ({ question }, { userToken }) => {
  const response = await fetch(`https://knowledge-base.com?token=${userToken}&query=${question}`);
  // ...
};

// define tool as usual
FunctionTool.from(queryKnowledgeBase, {
  name: 'queryKnowledgeBase',
  description: 'Query knowledge base',
  parameters: {
    type: 'object',
    properties: {
      question: {
        type: 'string',
        description: 'The user question',
      },
    },
    required: ['question'],
  },
});

// usage example
const response = await agent.chat({
  message: 'How to ...?',
  toolArg: { userToken: 'abcd1234' }
});
```


marcusschiesser:
@tvanier great idea! If i read your code correctly, it's sending `toolArg` to every tool call. How do you want to handle different tools?
How about you create a new tool based on `queryKnowledgeBase` that has `userToken: 'abcd1234'` bound, before calling `agent.chat`?
Does this solve your problem?

tvanier:
Thanks @marcusschiesser! Yes the same `toolArg` would be passed to all tools called within the same task context. I thought each tool could read zero or more fields in the provided `AdditionalToolArgument`.
I also like the idea of bounding arguments, the code below works, wondering if it is ok to re-create a tool+agent on every user request (ex: chat)?
Another possible approach: tools would receive some kind of "task ID" in input and could use it as a key to retrieve some contextual data (such as `userToken`)?

```ts
// bound argument followed by LLM input
const queryKnowledgeBase = async ({ userToken }, { question }) => { /* ... */ };

const createKbTool = (userToken: string) => {
  const boundFunction = queryKnowledgeBaseTool.bind(null, { userToken });
  return FunctionTool.from(boundFunction, {
    // same name, description and parameters
  };
};

const createKbAgent = (userToken: string) => {
  const kbTool = createKbTool(userToken);
  return new LLMAgent({
    tools: [kbTool],
    // llm, systemPrompt etc
  };
};

// usage example
const kbAgent = createKbAgent('abcd1234'); // on every request?
const response = await kbAgent.chat({
  message: 'How to ...?',
});
```

marcusschiesser:
> I thought each tool could read zero or more fields in the provided AdditionalToolArgument.
sounds like this will get confusing if you have more and more tools. same for the `"task ID"` lookup.

I don't see an issue recreating an LLMAgent per request, its a fast operation. 
To simplify your code, you could add a `bind` helper to `FunctionTool` instances like this:

```typescript
const queryKnowledgeBase = async ({ question }, { userToken }) => {
  const response = await fetch(`https://knowledge-base.com?token=${userToken}&query=${question}`);
  // ...
};

// define tool as usual
const kbTool = FunctionTool.from(queryKnowledgeBase, {
  name: 'queryKnowledgeBase',
  description: 'Query knowledge base',
  parameters: {
    type: 'object',
    properties: {
      question: {
        type: 'string',
        description: 'The user question',
      },
    },
    required: ['question'],
  },
});
```

Then in your request handler you just do:

```typscript
const bbAgent = new LLMAgent({
    tools: [kbTool.bind({ userToken })],
    // llm, systemPrompt etc
});
```

If you like that approach, I am happy to review the PR :D

tvanier:
Thanks @marcusschiesser for your feedback and suggestion, I just added a `bind` method to `FunctionTool`, which works for my project. Much simpler and cleaner :-)
1. Wondering if the tests should be in `tools.test.ts` or `function-tool.test.ts`?
2. The new `bind` method should probably be mentioned somewhere in the docs?

Side note: I just saw a workflow context can now be passed to a tool ([only for zod-typed tools?](https://github.com/run-llama/LlamaIndexTS/blob/main/packages/core/src/tools/function-tool.ts#L73)). Could this be redundant/confusing with this PR?

unit/tools/function-tool.test.ts
marcusschiesser, line 1: tools.test.ts is enough

marcusschiesser:
@tvanier yes that looks great! Let me add a doc file for `FunctionTool`, then you can update it with your bind function and we can release this PR.

tvanier:
> @tvanier yes that looks great! Let me add a doc file for `FunctionTool`, then you can update it with your bind function and we can release this PR.

Hi @marcusschiesser , I went ahead and added `apps/next/src/content/docs/llamaindex/modules/tool/index.mdx`, feel free to amend and/or move it. Same for the change set, could be `patch` bumps if preferable.

marcusschiesser (approved):

.changeset/ready-lines-rescue.md
```diff
@@ -0,0 +1,6 @@
+---
+"@llamaindex/core": minor
+"@llamaindex/doc": minor
```
marcusschiesser, line 3:
```suggestion
"@llamaindex/core": patch
"@llamaindex/doc": patch
```

marcusschiesser:
@tvanier thanks for leading the initiative

tvanier:
> @tvanier thanks for leading the initiative

Thanks @marcusschiesser for the great collaboration, much appreciated!

PR #1705, opened 2025-03-05
fix: build fail in edge runtime

thucpn:
In Next.js, this page should build successfully in Edge Runtime:

```
import "llamaindex";

export const runtime = "edge";
export default function Home() {
	return <div>Next.js Edge Runtime</div>;
}
```

### DIFF
--- a/packages/env/src/internal/tokenizers/node.ts
+++ b/packages/env/src/internal/tokenizers/node.ts
@@ -5,7 +5,7 @@ import { Tokenizers } from "./types.js";
 function tryLoadGptTokenizer() {
   try {
     // eslint-disable-next-line @typescript-eslint/no-require-imports
-    return require("gpt-tokenizer");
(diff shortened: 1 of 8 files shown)


packages/env/src/internal/tokenizers/node.ts
```diff
@@ -1,20 +1,47 @@
+import { getEncoding } from "js-tiktoken";
 import type { Tokenizer } from "./types.js";
 import { Tokenizers } from "./types.js";
 
-import cl100kBase from "gpt-tokenizer";
+function tryLoadGptTokenizer() {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    return require("gpt-tokenizer");
```
marcusschiesser, line 8: Why doesn't dynamic import work?

marcusschiesser (approved):

PR #1706, opened 2025-03-05
fix agent workflow tool call for Ollama

leehuwuj:
Ollama uses a partial toolCall, which means that the toolCall option only appears in a chunk.

### DIFF
--- a/packages/workflow/src/agent/function-agent.ts
+++ b/packages/workflow/src/agent/function-agent.ts
@@ -96,6 +96,7 @@ export class FunctionAgent implements BaseWorkflowAgent {
     });
     let response = "";
     let lastChunk: ChatResponseChunk | undefined;
+    const toolCalls: Map<string, AgentToolCall> = new Map();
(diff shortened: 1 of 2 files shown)


packages/workflow/src/agent/function-agent.ts
```diff
@@ -106,20 +107,23 @@ export class FunctionAgent implements BaseWorkflowAgent {
...
     const message: ChatMessage = {
       role: "assistant" as const,
       content: response,
     };
 
-    const toolCalls = lastChunk
-      ? this.getToolCallFromResponseChunk(lastChunk)
-      : [];
-    if (toolCalls.length > 0) {
+    if (toolCalls.size > 0) {
```
marcusschiesser, line 124: lastChunk still needed?

leehuwuj: yeah, i think we should keep last chunk for OpenAI raw data

marcusschiesser (approved):

PR #1712, opened 2025-03-07
Add Gemini 2.0 Flas Lite, Fix tools error with LLM Agent

atpsito:


marcusschiesser (approved):

marcusschiesser:
thanks @atpsito for the PR!

packages/providers/google/src/base.ts
```diff
@@ -278,24 +279,26 @@ export class Gemini extends ToolCallLLM<GeminiAdditionalChatOptions> {
...
-                  mapBaseToolToGeminiFunctionDeclaration,
-                ),
-              },
-            ],
-            safetySettings: DEFAULT_SAFETY_SETTINGS,
-          }
-        : {
-            history: context.history,
-            safetySettings: DEFAULT_SAFETY_SETTINGS,
+    const tools = params.tools?.length
```
tvanier, line 282: Thanks @atpsito and @marcusschiesser  for the fix, I was also facing the problem.
I would suggest to make the same change in `nonStreamChat`?
https://github.com/run-llama/LlamaIndexTS/blob/main/packages/providers/google/src/base.ts#L240
Maybe `nonStreamChat` and `streamChat` could share a `createStartChatParams` utility function?

marcusschiesser: yes that would be great, want to send a PR?

tvanier: Just created https://github.com/run-llama/LlamaIndexTS/pull/1716

PR #1715, opened 2025-03-07
feat: add asQueryTool to index

marcusschiesser:


packages/llamaindex/src/indices/BaseIndex.ts
```diff
@@ -61,6 +67,37 @@ export abstract class BaseIndex<T> {
...
+  /**
+   * Returns a query tool by calling asQueryEngine.
+   * Either options or retriever can be passed, but not both.
+   * If options are provided, they are passed to generate a retriever.
+   */
+  asQueryTool(
+    params: (
+      | {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          options: any;
```
himself65, line 79: could we fix this any type?

marcusschiesser: yes, it's ugly, but let's do it in another PR

PR #1716, opened 2025-03-07
fix: google start chat tools params

tvanier:
Follow-up of https://github.com/run-llama/LlamaIndexTS/pull/1712:

The Google chat API will not accept an empty list of tools, this PR fixes it for both streaming and non-streaming chat calls.


marcusschiesser (approved):
thanks @tvanier !

marcusschiesser:
@tvanier seems like a test is failing, can you have a look?

marcusschiesser:
yes thanks, now it looks good!

PR #1719, opened 2025-03-08
Feat/perplexity

AVtheking:
This PR add support for perplexity  api


packages/providers/perplexity/package-lock.json
marcusschiesser, line 1: please remove (we're using pnpm)

packages/providers/perplexity/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "bunchee",
+    "dev": "bunchee --watch"
+  },
+  "devDependencies": {
+    "bunchee": "6.3.4"
```
marcusschiesser, line 28: use latest from main
```suggestion
    "bunchee": "6.4.0"
```

packages/providers/perplexity/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@llamaindex/core": "workspace:*",
+    "@llamaindex/env": "workspace:*",
+    "@llamaindex/openai": "workspace:*",
+    "openai": "^4.86.0"
```
marcusschiesser, line 37: should not be needed as `@llamaindex/openai` has it
```suggestion
```

packages/providers/perplexity/src/llm.ts
```diff
@@ -0,0 +1,140 @@
...
+type PerplexityModelName = keyof typeof PERPLEXITY_MODELS;
+
+export interface PerplexityConfig {
+  apiKey: string;
+  model: PerplexityModelName;
+  temperature?: number;
+  topP?: number;
+}
+
+export class Perplexity extends BaseLLM {
```
marcusschiesser, line 45: if you're using the llamaindex OpenAI package, why not extending `OpenAI` LLM? (see e.g. fireworks LLM)

AVtheking: Thanks for the review, my only concern with this was it would expose some methods in the perplexity which are not supported by it. I may need to override some methods .

AVtheking: I see we use this pattern in deepseek also and that is not working when I am running the example of deepseek

AVtheking: image
As I have azure endpoint set  in my env it is checking from the azure model and throwing error for the deepseek model.

AVtheking: should I just remove this env variable check as this seems wrong ? @marcusschiesser

marcusschiesser: yes this check seems to be wrong for deepseek and your package, but correct for the openai package. can you disable it?

AVtheking: but this env check shouldn't be there for openai package too imo as user could have both openai key and this azure endpoint in his env right ?

marcusschiesser: yes but if we remove it from openai package, then we need to move this check from the openai to the azure package.
i suggest you ignore this issue in your PR, and if you like you can address this issue in a new PR

marcusschiesser:
@AVtheking can you please also add a factory function for you LLM provider, see https://github.com/run-llama/LlamaIndexTS/pull/1731 - i just started adding them

packages/providers/perplexity/src/llm.ts
AVtheking: sure I can address this in new PR . will revert the change.

AVtheking:
> @AVtheking can you please also add a factory function for you LLM provider, see #1731 - i just started adding them

Pushed

examples/perplexity.ts
marcusschiesser, line 1: this is failing type-check:

```
Error: examples/perplexity.ts(1,28): error TS2307: Cannot find module '@llamaindex/perplexity' or its corresponding type declarations.
```
you'll need to add `@llamaindex/perplexity` to package.json of examples

AVtheking: I was about to ask this, as this package is not yet published yet in npm , how could I add that in package.json ?  one way could be adding it using workspace refrence , or add it assuming it get deployed in npm ?

AVtheking: added it in the package.json.

marcusschiesser: Thanks, changing the package.json you also need to rerun pnpm i to update the lock file. Also I noticed that you need to update the root tsconfig.json

AVtheking: Thanks I missed that , updated the tsconfig.json and reran pnpm i

marcusschiesser:
@AVtheking yeah! tests are passing, i'll merge it

marcusschiesser:
feel free to add a doc to https://ts.llamaindex.ai/docs/llamaindex/modules/llms

PR #1748, opened 2025-03-12
feat(workflow): remove `new` syntax

himself65:


himself65:
https://github.com/run-llama/fluere

PR #1755, opened 2025-03-14
feat: @llamaindex/tools

thucpn:


packages/server/src/index.ts
marcusschiesser, line 1: better own PR for server package?

packages/tools/src/settings.ts
```diff
@@ -0,0 +1,22 @@
+class GlobalToolSettings {
```
marcusschiesser, line 1: how about setting these options per tool?
as a general pattern we could use a factory function to create a tool, e.g. `wiki()` instead of `wiki` - then we can pass parameters to create a tool, e.g.:
```
documentGenerator({ outputDir })
```

AVtheking:
the tools test is failing on node 18.x version

marcusschiesser:
@AVtheking sorry, @thucpn will have a look

Issue #1561, opened 2024-12-11
PineconeVectorStore .delete() not working


LawrenceGB:
**Describe the bug**
.delete() action from PineconeVectorStore not working.

**To Reproduce**
Ran a test here, I called delete from a separate call to test and it does not delete the doc from Pinecone:

```ts
 const nodeId = "5bb16627-f6c0-459c-bb18-71642813ef21";

const pcvs = new PineconeVectorStore({
    indexName: 'index-1',
    namespace: 'test',
});

const doc = new Document({
    id_: nodeId,
    text: 'Hello world.',
})

const storageContext = await storageContextFromDefaults({
    vectorStore: pcvs,
});

await VectorStoreIndex.fromDocuments([doc], {
    storageContext,
});

await pcvs.delete(nodeId);
```

**Expected behavior**
Should have deleted doc from storage

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. macOS, Linux]
- JS Runtime / Framework / Bundler (select all applicable)
- [X ] Node.js
- [ ] Deno
- [ ] Bun
- [ ] Next.js
- [ ] ESBuild
- [ ] Rollup
- [ ] Webpack
- [ ] Turbopack
- [ ] Vite
- [ ] Waku
- [ ] Edge Runtime
- [ ] AWS Lambda
- [ ] Cloudflare Worker
- [ ] Others (please elaborate on this)

**Additional context**

PR #1758, opened 2025-03-16
fix: delete by id in pinecone vector store

AVtheking:
Fixes https://github.com/run-llama/LlamaIndexTS/issues/1561


packages/providers/storage/pinecone/src/PineconeVectorStore.ts
```diff
@@ -149,7 +157,35 @@ export class PineconeVectorStore extends BaseVectorStore {
...
     const idx = await this.index();
-    return idx.deleteOne(refDocId);
+    try {
+      await idx.deleteMany({
+        metadata: {
+          ref_doc_id: refDocId,
+        },
+      });
+    } catch (e) {
+      // fallback to deleting by prefix for serverless indexes
```
marcusschiesser, line 167: 1. why is this different for serverless indexes?
2. can't we programmatically check if its a serverless index (instead of try/catch)

AVtheking: 1. It is mentioned in there official docs that filtering by metadata is not supported for the serverless index.
2. There was no documentation regarding this for ts sdk, but as I was digging through in there ts sdk I found one descibe api which gives us the information about the index. Used that instead of try catch .

AVtheking:
@marcusschiesser this node.js 18.x test is failing after recent merge

.changeset/itchy-seas-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@llamaindex/pinecone": minor
+"@llamaindex/core": minor
```
marcusschiesser, line 3: Core change not needed

packages/providers/storage/pinecone/src/PineconeVectorStore.ts
```diff
@@ -148,8 +156,40 @@ export class PineconeVectorStore extends BaseVectorStore {
...
+      //if the index is a pod, delete the document by the metadata
+      await idx.deleteMany({
+        metadata: {
+          ref_doc_id: refDocId,
+        },
+      });
+    } else if (index?.spec?.serverless) {
+      // get the list of ids with the prefix
+      let list = await idx.listPaginated({
+        prefix: refDocId,
```
marcusschiesser, line 175: Would this also work for non- serverless? My guess is we're doing a different version because it's more efficient? If so a comment would help to understand this

AVtheking: no this listPagniated api is supported for serverless index only, that's why I have put this under serverless check. I will add a comment too here.

AVtheking: I have added comment and link pinecone doc there . will share here also https://docs.pinecone.io/guides/data/delete-data#delete-records-by-metadata

AVtheking:
@thucpn could you please take a look and help me investigate why this tool test failing on node.js 18 version since  you have added those ?

marcusschiesser:
Thanks @AVtheking !

PR #1759, opened 2025-03-17
feat: @llamaindex/server

thucpn:


packages/server/src/workflow/type.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  Workflow,
+  WorkflowEvent,
+  type ChatResponseChunk,
+  type MessageContent,
+} from "llamaindex";
+
+export type AgentInput = {
```
leehuwuj, line 8: How about we refactor to use standard input here?
`userMessage` and `chatHistory`.

examples/agent/wiki.ts
```diff
@@ -1,13 +1,12 @@
...
 import { AgentStream, agent } from "llamaindex";
-import { WikipediaTool } from "../wiki";
 
 async function main() {
   const llm = new OpenAI({ model: "gpt-4-turbo" });
-  const wikiTool = new WikipediaTool();
 
   const workflow = agent({
-    tools: [wikiTool],
+    tools: [wiki()],
```
leehuwuj, line 9: Just personal opinion, Using `wikipedia` is better than wiki.

leehuwuj: It would be easier for us later in UI work to use the same tool name in both Python and TypeScript

packages/server/src/server.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public chatController = async (
+    req: express.Request,
+    res: express.Response,
+  ) => {
+    try {
+      const { messages } = req.body as { messages: Message[] };
+      const userMessageContent = getUserMessageContent(messages);
+      this.callbacks?.beforeChat?.(messages as ChatMessage[]);
```
leehuwuj, line 36: Do this `callbacks.beforeChat` already exist in the workflow? It seem strange to me

packages/server/src/workflow/stream.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (delta) {
+        controller.enqueue({ delta } as EngineResponse);
+      }
+    },
+  });
+
+  return { stream, dataStream };
+}
+
+function handleEvent(
```
leehuwuj, line 68: I guess in this PR scope, we will keep the code only for the workflow and will support the agent workflow later?

packages/server/next/app/layout.tsx
```diff
@@ -0,0 +1,19 @@
+import type { Metadata } from "next";
+import "./globals.css";
+
+export const metadata: Metadata = {
+  title: "Create Next App",
+  description: "Generated by create next app",
```
marcusschiesser, line 6: LlamaIndexServer specific

packages/server/src/workflow/tools.ts
marcusschiesser, line 1: Maybe put these tool helpers to the tools package?

packages/server/src/workflow/stream.ts
marcusschiesser: Can also already support agent workflows

packages/server/src/server.ts
```diff
@@ -0,0 +1,62 @@
...
+  async start() {
+    await this.app.prepare();
+
+    const server = createServer((req, res) => {
+      const parsedUrl = parse(req.url!, true);
+      const pathname = parsedUrl.pathname;
+
+      if (pathname === "/api/chat" && req.method === "POST") {
+        return this.handleChat(req, res);
+      }
```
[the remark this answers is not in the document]
marcusschiesser: Cool. Can we still do one file per handler?

packages/server/src/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  | Workflow<null, AgentInput, ChatResponseChunk>
+  | AgentWorkflow;
+
+/**
+ * A factory function that creates a ServerWorkflow instance, possibly asynchronously.
+ * The requestBody parameter is the body from the request, which can be used to customize the workflow per request.
+ */
+export type WorkflowFactory = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  requestBody?: any,
```
marcusschiesser, line 32: Do we have a type here?

packages/server/src/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  userInput: string;
+  chatHistory: ChatMessage[];
+};
+
+/**
+ * ServerWorkflow can be either a normal Workflow or an AgentWorkflow
+ */
+export type ServerWorkflow =
+  | Workflow<null, AgentInput, ChatResponseChunk>
+  | AgentWorkflow;
```
marcusschiesser, line 24: Just an idea: Can we just use AgentWorkflow for all workflows?

marcusschiesser: I see.
@leehuwuj why isn't a agentworkflow a workflow too?
@thucpn does it make sense then that `ServerWorkflow = Workflow<AgentWorkflowContext, AgentInputData, string> | AgentWorkflow` so that we're at least using the same type of workflow?

packages/server/src/server.ts
```diff
@@ -1,3 +1,4 @@
+import "dotenv/config";
```
marcusschiesser, line 1: why do we want to add dotenv here? this is more a package that we can add to the application code

packages/server/src/types.ts
leehuwuj: At the time, i only use a wrapper for an easier implementation, but i guess there is no issue to extends the AgentWorkflow from Workflow.

packages/server/next/app/env.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+"use client";
+
+/**
+ * A helper function to get environment variables from the window object
```
marcusschiesser, line 5: Wrong doc. These are not environment variables. Then also getEnv is confusing. getConfig ?

marcusschiesser (commented):
.

marcusschiesser (commented):
.

marcusschiesser (approved):
great job @thucpn !

PR #1762, opened 2025-03-18
feat: google multimodal output using their new gen ai library 

parhammmm:


packages/community/src/llm/google/index.ts
```diff
@@ -0,0 +1,284 @@
...
+export type GoogleModelParams = {
+  model: GEMINI_MODEL;
+  temperature?: number;
+  topP?: number;
+  maxTokens?: number;
+};
+
+export type GoogleParams = GoogleGenAIOptions & GoogleModelParams;
+
+export class GoogleStudio extends ToolCallLLM<GoogleAdditionalChatOptions> {
```
marcusschiesser, line 100: @parhammmm thanks for your PR! Looks great, but I am not sure putting it in the community package is the best way.

The community package doesn't make a lot of sense to me anymore now that we have external dependencies wrapped in our own provider packages.
It contains only AWS bedrock code, so it better should be in a new @llamaindex/aws package.

About the new `GoogleStudio` class - do you think we can just move it to @llamaindex/google without impacting the existing google classes?

parhammmm: @marcusschiesser sure I'll update this PR and move the code inside the google provider

parhammmm: @marcusschiesser there's a version clash between `js-genai` and the mix of vertex sdk and generative-ai-js

I think it's best just to have it in the community package for the time being until a full switch

parhammmm: @marcusschiesser turns out it's the model itself that's not behaving as expected and not the libraries

`GoogleStudio` is now in `@llamaindex/google` and can be reviewed

.changeset/tidy-waves-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@llamaindex/community": patch
```
marcusschiesser, line 2:
```suggestion
"@llamaindex/community": patch
"@llamaindex/google": patch
```

PR #1765, opened 2025-03-20
feat: vercel tool response fields options

danielbank:
I want to be able to access source nodes from the tool call with the Vercel integration:

```ts
import { LlamaCloudIndex } from "@llamaindex/cloud";

// Create a LlamaCloud index
const index = await LlamaCloudIndex.fromDocuments({
  documents: [document],
  name: "your-index-name",
  projectName: "your-project",
  apiKey: process.env.LLAMA_CLOUD_API_KEY,
});

// Use it the same way as VectorStoreIndex
const queryTool = llamaindex({
  model: openai("gpt-4"),
  index,
  description: "Search through the documents",
  options: { fields: ["sourceNodes", "messages"]} // get messages and sourceNodes
});

// Use the tool with Vercel's AI SDK
streamText({
  model: openai("gpt-4"),
  prompt: "Your question here",
  tools: { queryTool },
}).toDataStream();
```

### DIFF
--- a/.changeset/purple-gifts-decide.md
+++ b/.changeset/purple-gifts-decide.md
@@ -1,0 +1,6 @@
+---
+"@llamaindex/vercel": minor
+"@llamaindex/doc": minor
+---
+
+Adding an options parameter to vercel tool to tailor responses
--- a/apps/next/src/content/docs/llamaindex/integration/vercel.mdx
+++ b/apps/next/src/content/docs/llamaindex/integration/vercel.mdx
@@ -84,6 +84,7 @@ const queryTool = llamaindex({
   model: openai("gpt-4"),
   index,
   description: "Search through the documents",
+  options: { fields: ["sourceNodes", "messages"]}
 });
 
 // Use the tool with Vercel's AI SDK
--- a/packages/providers/vercel/package.json
+++ b/packages/providers/vercel/package.json
@@ -35,10 +35,12 @@
   },
   "scripts": {
     "build": "bunchee",
-    "dev": "bunchee --watch"
+    "dev": "bunchee --watch",
+    "test": "vitest run"
   },
   "devDependencies": {
-    "bunchee": "6.4.0"
+    "bunchee": "6.4.0",
+    "vitest": "^2.1.5"
   },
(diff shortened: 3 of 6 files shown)


marcusschiesser:
Good idea, how about always passing sourceNodes without specifically setting some options?

danielbank:
> Good idea, how about always passing sourceNodes without specifically setting some options?

I'm also fine with that, and can make that change if that's what is desired.  One argument for allowing it to be configurable is to give the implementer more control over their context growth rate by using the tool (though honestly, I don't know how much more tokens it costs to include the source nodes).

marcusschiesser (approved):

marcusschiesser:
@danielbank then let's just merge it like it is first. Thanks a lot!

marcusschiesser:
@danielbank can you just check the tests please?

marcusschiesser (approved):

marcusschiesser:
i think you need to update the lock file:
```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/packages/providers/vercel/package.json

Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"

    Failure reason:
    specifiers in the lockfile ({"@llamaindex/core":"workspace:*","ai":"^4.0.0","zod":"^3.23.[8](https://github.com/run-llama/LlamaIndexTS/actions/runs/13978240987/job/39137561200?pr=1765#step:5:9)","bunchee":"6.4.0"}) don't match specs in package.json ({"ai":"^4.0.0","bunchee":"6.4.0","vitest":"^2.1.5","@llamaindex/core":"workspace:*","zod":"^3.23.8"})
```

danielbank:
I'm assuming that's the root-level pnpm-lock.yaml file because there isn't a lock file in the vercel provider folder.  I updated the root-level lock file, we will see if that works

marcusschiesser:
@danielbank yes it's the root-level pnpm-lock.yaml file, there's only one for all packages as it's a monorepo.

PR #1772, opened 2025-03-20
fix: context engine additional options not being passed

parhammmm:
This was the root cause of the GoogleStudio failing to generate images.


marcusschiesser (approved):

marcusschiesser:
Thanks!

PR #1773, opened 2025-03-21
Extend AgentWorkflow from Workflow

leehuwuj:


packages/workflow/src/agent/agent-workflow.ts
```diff
@@ -554,37 +559,6 @@ export class AgentWorkflow {
...
-    params?: {
-      chatHistory?: ChatMessage[];
-      context?: AgentWorkflowContext;
-    },
-  ): WorkflowContext<AgentInputData, string, AgentWorkflowContext> {
-    if (this.agents.size === 0) {
-      throw new Error("No agents added to workflow");
-    }
-    this.setupWorkflowSteps();
-    const contextData: AgentWorkflowContext = params?.context ?? {
```
leehuwuj, line 569: By default, Workflow has run signatures for start event and workflow context. Would be better if we can support to set the context.data easier (it's readonly and will be undefined if not provided to the run function)

marcusschiesser: Alex is working on updating workflows, we will align agent workflows with the new ones then

leehuwuj: noted! thanks for the information

marcusschiesser:
not needed anymore

Issue #818, opened 2024-05-07
ElasticSearch Vector Store support


madarco:
Hi, 

and thanks for the great work with this library.

Is there any plan to port the ElasticSearch support from the python version to TS?

Thanks.

PR #1777, opened 2025-03-22
feat: elastic search vector store

AVtheking:
Fixes https://github.com/run-llama/LlamaIndexTS/issues/818


marcusschiesser (approved):
thanks @AVtheking

Issue #32, opened 2023-07-25
Add support for supabase vector store


theprashant-one:

yisding:
Let’s do it!

ky-zo:
Bumping up the topic - are there any plans to implement Supabase here?

marcusschiesser:
We already have PostgresDocumentStore and PGVectorStore, @ky-zo what are you missing?

PR #1790, opened 2025-03-26
Feat/supbase vector store

AVtheking:
- Fixes https://github.com/run-llama/LlamaIndexTS/issues/32

### DIFF
--- a/e2e/examples/vite-import-llamaindex/tsconfig.tsbuildinfo
+++ b/e2e/examples/vite-import-llamaindex/tsconfig.tsbuildinfo
@@ -1,0 +1,1 @@
+{"root":["./src/main.ts","./vite.config.ts"],"version":"5.7.3"}
--- a/e2e/examples/vite-import-llamaindex/tsconfig.tsbuildinfo
+++ b/e2e/examples/vite-import-llamaindex/tsconfig.tsbuildinfo
@@ -1,1 +1,0 @@
-{"root":["./src/main.ts","./vite.config.ts"],"version":"5.7.3"}
--- a/examples/vector-store/supbases/index.ts
+++ b/examples/vector-store/supbases/index.ts
@@ -4,13 +4,13 @@ import {
   GEMINI_MODEL,
   GeminiEmbedding,
 } from "@llamaindex/google";
+import { SupabaseVectorStore } from "@llamaindex/supabase";
 import {
   Document,
   Settings,
   storageContextFromDefaults,
   VectorStoreIndex,
 } from "llamaindex";
-import { SupabaseVectorStore } from "../../../packages/providers/storage/supabase/src";
 async function main() {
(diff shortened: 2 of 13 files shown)


packages/providers/storage/supabase/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@llamaindex/supabase",
```
marcusschiesser, line 2: isn't supabase using postgres? we already have a PGVectorStore in `@llamaindex/postgres` - can we base this package on the postgres package?

AVtheking: Do you mean to extend it one the postgres package instead of creating new ?

marcusschiesser: yes that way you make sure that we also get the other classes of the postgres package. if you like you can also do https://github.com/run-llama/LlamaIndexTS/issues/1791 first

AVtheking: Okay but in the python package we have two different

AVtheking: I initaly thought to use the code from postgres package but there is a lot of difference between both , like in postgres package we are using raw sql query , but here we are using supbase client to perform crud operations

marcusschiesser: i see, if you think it doesn't make sense reusing postgres package, then we can continue with your approach.

AVtheking: yes

examples/vector-store/supbases/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    }),
+    new Document({
+      text: "Vector search enables semantic similarity search",
+      metadata: {
+        source: "research_paper",
+        author: "Jane Smith",
+      },
+    }),
+    new Document({
+      text: "Elasticsearch supports various distance metrics for vector search",
```
marcusschiesser, line 38: whole example is saying https://github.com/run-llama/LlamaIndexTS/pull/1790/files

e2e/examples/vite-import-llamaindex/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
+{"root":["./src/main.ts","./vite.config.ts"],"version":"5.7.3"}
```
marcusschiesser, line 1: needed?

examples/vector-store/supbases/index.ts
AVtheking: sorry didn't get you ?

AVtheking: oh I mispelled that sorry replacing it :sweat_smile:

marcusschiesser (approved):

AVtheking:
why is this deployment on  vercel failing ?

PR #1796, opened 2025-03-28
feat: support llamacloud in @llamaindex/server

thucpn:
- [x] api to get llamacloud pipelines
- [x] support query with llamacloud
- [x] server options: app title, useLlamaCloud
- [x] toQueryTool function for LlamaCloudIndex


packages/server/src/types.ts
```diff
@@ -27,4 +27,6 @@ export type NextAppOptions = Parameters<typeof next>[0];
 export type LlamaIndexServerOptions = NextAppOptions & {
   workflow: WorkflowFactory;
   starterQuestions?: string[];
+  appTitle?: string;
+  useLlamaCloud?: boolean;
```
marcusschiesser, line 31: 1. @leehuwuj the llamacloud parameter should be in sync with the python version
2. seems like the idea is that if this is true then these env variables must be defined:
LLAMA_CLOUD_API_KEY
LLAMA_CLOUD_INDEX_NAME
LLAMA_CLOUD_PROJECT_NAME

how about passing them instead as object:
```
llamaCloud: {
apiKey,
index,
project
}
```
if the object is undefined/null (or `llamaCloud.apiKey` is undefined/null) we don't use llamacloud

then create-llama is generating this object with:
```
llamaCloud: {
apiKey: process.env.LLAMA_CLOUD_API_KEY
index: process.env.LLAMA_CLOUD_INDEX_NAME
project: process.env.LLAMA_CLOUD_PROJECT_NAME
}
```
AFAIK we can't modify the LlamaIndexServer parameters easily

packages/server/src/events.ts
```diff
@@ -36,6 +38,12 @@ export class AgentRunEvent extends WorkflowEvent<{
   data: AgentRunEventData;
 }> {}
 
+export class ReportEvent extends WorkflowEvent<object> {}
+
+export class StopEvent extends StopEventBase<
+  AsyncGenerator<ChatResponseChunk>
+> {}
```
marcusschiesser, line 45: not sure why we have these events here - llamaindexserver should only have events that are rendered in the frontend

packages/server/src/handlers/cloud.ts
```diff
@@ -0,0 +1,29 @@
+import type { IncomingMessage, ServerResponse } from "http";
+import { LLamaCloudFileService } from "llamaindex";
+import { sendJSONResponse } from "../utils/request";
+
+export const getLlamaCloudConfig = async (
+  req: IncomingMessage,
+  res: ServerResponse,
+) => {
+  if (!process.env.LLAMA_CLOUD_API_KEY) {
```
marcusschiesser, line 9: also use getEnv here in this file
```suggestion
  if (!getEnv("LLAMA_CLOUD_API_KEY")) {
```

packages/server/next/app/components/ui/chat/custom/llama-cloud-selector.tsx
```diff
@@ -65,7 +65,7 @@ export function LlamaCloudSelector({
   );
 
   useEffect(() => {
-    if (process.env.NEXT_PUBLIC_USE_LLAMACLOUD === "true" && !config) {
+    if (!config) {
```
marcusschiesser, line 68: how is this hidden if env variables are not defined?

packages/server/src/events.ts
marcusschiesser: this means we should move the deepresearch events here (as they are rendered in this package)

packages/tools/src/tools/document-generator.ts
marcusschiesser, line 1: @thucpn need to update changeset, otherwise no release of this package

marcusschiesser (approved):

marcusschiesser (approved):

PR #1801, opened 2025-03-30
feat:openai responses api

AVtheking:


AVtheking:
@marcusschiesser PTAL

packages/core/src/llms/type.ts
```diff
@@ -158,9 +162,27 @@ export type MessageContentImageDetail = {
...
+export type ResponsesMessageContentTextDetail = {
+  type: "input_text";
+  text: string;
+};
+
+export type ResponsesMessageContentImageDetail = {
+  type: "input_image";
+  image_url: string;
+  detail?: "high" | "low" | "auto";
+};
```
marcusschiesser, line 174: this type seems to me a detail of the openai responses which should not be in core.
comparing them the new information that we need is `detail?: "high" | "low" | "auto";` why not moving it to `MessageContentImageDetail`?

marcusschiesser: we already have 
```
export type MessageContentImageDetail = {
  type: "image_url";
  image_url: { url: string };
};
```
why can't you extend it to 
```
export type MessageContentImageDetail = {
  type: "image_url";
  image_url: { url: string };
  detail?: "high" | "low" | "auto";
};
```
and map it to whatever needed in the openairesponses implementation ?

AVtheking: type in new responses api is ```"input_image"```  which would give typescript error if I use the `MessageContentImageDetail` and also `image_url` directly requires url string not with field url

AVtheking:
```export type ResponsesMessageContentImageDetail = {
  type: "input_image";
  image_url: string;
  detail?: "high" | "low" | "auto";
};``` see the type and image_url field too

Comment (marcusschiesser) [reply packages/core/src/llms/type.ts:174]: what i mean is you map 
```
{
  type: "image_url";
  image_url: { url: string };
  detail?: "high" | "low" | "auto";
}
```
to 
```
{
  type: "input_image";
  image_url: string;
  detail?: "high" | "low" | "auto";
}
```
in your `openairesponses` code. `ResponsesMessageContentImageDetail` is an implementation detail of the openai package

marcusschiesser: until now we didn't do that for openai as we have been using their format. but now they have two formats and we want to stick with one format

AVtheking: oh got it

AVtheking: fixed this .

marcusschiesser (approved):
great job @AVtheking

Issue #1791, opened 2025-03-27
Add reader to @llamaindex/postgres


marcusschiesser:
- Add reader to @llamaindex/postgres package
- add documentation

PR #1813, opened 2025-04-03
feat: reader for postgres

AVtheking:
https://github.com/run-llama/LlamaIndexTS/issues/1791


AVtheking:
@marcusschiesser  PTAL and let me know if the approach for this looks good, I will then proceed to write test and update documentation for this.

packages/readers/src/postgres.ts
```diff
@@ -0,0 +1,147 @@
+import { type BaseReader, Document } from "@llamaindex/core/schema";
```
marcusschiesser, line 1: better add it to @llamaindex/postgres package
(we should not have one big readers package with all sorts of deps)

AVtheking: rest all approach looks good ?

marcusschiesser: easier to judge if there's an example :D

AVtheking: Added example and test also

packages/readers/package.json
marcusschiesser, line 1: not needed

examples/vector-store/pg/pg-reader.ts
```diff
@@ -0,0 +1,52 @@
...
+  port: 5432,
+  database: "test",
+  user: "postgres",
+  password: "postgres",
+};
+Settings.embedModel = new OpenAIEmbedding();
+async function main() {
+  try {
+    // Initialize PGVectorStore and add a document
+    const vectorStore = new PGVectorStore({
```
marcusschiesser, line 17:
```suggestion
    // Initialize PGVectorStore and add a document
    // This is just to generate test data in the DB for the SimplePostgresReader
    // Usually it's purpose is to read documents from a Postgres DB
    const vectorStore = new PGVectorStore({
```

marcusschiesser (approved):

AVtheking:
The tests related to mcp's are failing

PR #1821, opened 2025-04-04
chore: use bunchee for llamaindex

thucpn:
- [x] should solve conflicts in export https://github.com/run-llama/LlamaIndexTS/pull/1816#discussion_r2028143004
- [x] do not export all, only export main entry points
- [x] update docs and examples


himself65 (commented):
I like this, well done

packages/llamaindex/package.json
```diff
@@ -33,8 +33,7 @@
     "magic-bytes.js": "^1.10.0"
   },
   "devDependencies": {
-    "@swc/cli": "^0.5.0",
-    "@swc/core": "^1.9.2",
+    "bunchee": "6.4.0",
```
marcusschiesser, line 36: AFAIK we're using now bunchee everywhere - can we set the version in the root package.json? (then it's easier to upgrade bunchee)

packages/llamaindex/package.json
```diff
@@ -63,33 +62,141 @@
...
+        "types": "./extractors/dist/index.d.cts",
+        "default": "./extractors/dist/index.cjs"
+      },
+      "import": {
+        "types": "./extractors/dist/index.d.ts",
+        "default": "./extractors/dist/index.js"
+      },
+      "default": "./extractors/dist/index.js"
     },
-    "./*": {
```
marcusschiesser, line 84: @himself65 seems like bunchee does not support this wildcard exports (/*)

Adding all current entry points is not maintainable (also slows down the build process a lot)

Our solution is too only export the current key entry points as in this PR, but this is a breaking change.

Do you have any better idea to handle this issue?

himself65: I think we have to specify all the subpath, for now...

marcusschiesser: Thanks, I think then we just export the first level (as we do right now in this PR) and we accept that some code might break 
@thucpn please 
1. double check the docs that we're not suggesting to use an invalid export some where
2. add a minor changeset to be sure

packages/llamaindex/src/index.edge.ts
```diff
@@ -63,6 +63,7 @@ export * from "@llamaindex/core/retriever";
 export * from "@llamaindex/core/schema";
 export * from "@llamaindex/core/storage/chat-store";
 export * from "@llamaindex/core/storage/doc-store";
+export { BaseDocumentStore } from "@llamaindex/core/storage/doc-store"; // explicitly export BaseDocumentStore
```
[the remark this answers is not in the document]
marcusschiesser: @himself65 you know the root cause of this?

marcusschiesser (commented):
LGTM

marcusschiesser (approved):

PR #1825, opened 2025-04-07
Add "thinking" and "thinking_signature" to chat response, and "thinking_signature" to chat stream

r3rer3:
## Overview

This adds the `thinking` messages and their signatures (`thinking_signature`) to the `options` field when `stream` is `false`, and also adds `thinking_signature` to `options` when `stream` is `true`, 

## Changes
 
- I noticed the code at https://github.com/run-llama/LlamaIndexTS/blob/main/examples/anthropic/thinking.ts returns `thinking` in `options`, but no `signature` for the `thinking` messages was being returned
- I noticed that making `stream: false` wasn't returning `thinking` nor `thinking_signature` in `option`, so I added them both
- Since now the user may add the returned messages to the history, I changed `formatMessages` to handle the new keys in `options`, so that Anthropic receives the expected format

## Tests
I added tests to make sure `thinking` is correctly converted and to make sure it comes before `text' and tool calls.

### DIFF
--- a/.changeset/yellow-carpets-return.md
+++ b/.changeset/yellow-carpets-return.md
@@ -1,0 +1,9 @@
+---
+"@llamaindex/anthropic": patch
+---
+
+added `option.thinking` and `option.thinking_signature` to Anthropic's chat response
+
+added `option.thinking_signature` to Anthropic's chat stream response
+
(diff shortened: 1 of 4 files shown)


marcusschiesser (approved):

thalesvinif:
🚀

PR #1833, opened 2025-04-10
feat: enhance config params for LlamaIndexServer

thucpn:
```
new LlamaIndexServer({
  workflow: workflowFactory,
  uiConfig: {
    appTitle: "LlamaIndex App",
    componentsDir: "components",
    starterQuestions: ["What is the LlamaIndex?"],
    llamaCloudIndexSelector: true,
  },
}).start();
```

### DIFF
--- a/.changeset/mighty-ducks-tell.md
+++ b/.changeset/mighty-ducks-tell.md
@@ -1,6 +1,6 @@
 ---
 "@llamaindex/doc": patch
-"@llamaindex/server": patch
+"@llamaindex/server": minor
 ---
 
 feat: enhance config params for LlamaIndexServer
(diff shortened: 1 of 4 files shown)


.changeset/mighty-ducks-tell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@llamaindex/doc": patch
+"@llamaindex/server": patch
```
marcusschiesser, line 3: @thucpn do we need minor as it's breaking and maybe we use ^ dep?
```suggestion
"@llamaindex/server": minor
```

marcusschiesser (approved):

PR #1850, opened 2025-04-15
fix: update the tool call schema for nova

parhammmm:


packages/community/src/llm/bedrock/amazon/utils.ts
```diff
@@ -132,7 +132,9 @@ export const mapBaseToolsToAmazonTools = (tools?: BaseTool[]): Tool[] => {
     return {
       toolSpec: {
         ...options,
-        inputSchema: parameters,
+        inputSchema: {
```
marcusschiesser, line 135: @parhammmm what's the impact of using other models?

parhammmm: Amazon only has the Nova models, so it wont affect none-nova modals (other models are one folder above)

marcusschiesser:
Thanks @parhammmm

PR #1857, opened 2025-04-16
Add o3 and o4-mini models

petergoldstein:


himself65 (approved):
that's qucik, thx

petergoldstein:
@himself65 No problem.  A quick release with this would be great

PR #1859, opened 2025-04-17
Update o4-mini to allow reasoning parameters and exclude temperature

petergoldstein:


petergoldstein:
Missed these when I added o4-mini

marcusschiesser (approved):

PR #1867, opened 2025-04-17
remove fake chat

logan-markewich:
The little chat feature was fake and kind of confusing for users.

Lets just remove it until we implement an actual chat feature


himself65 (approved):

himself65:
I think it's a real check in the first version. who delete it?

PR #1874, opened 2025-04-21
docs: replace with llama-flow docs

logan-markewich:


himself65:
flaky: https://github.com/run-llama/LlamaIndexTS/actions/runs/14577058451/job/40885397050?pr=1874

PR #1895, opened 2025-04-28
chore: migrate agentworkflows to llama-flow

marcusschiesser:


packages/workflow/src/agent/events.ts
```diff
@@ -1,43 +1,48 @@
-import { WorkflowEvent } from "@llama-flow/llamaindex";
+import { workflowEvent } from "@llama-flow/core";
 import type { JSONValue } from "@llamaindex/core/global";
 import type { ChatMessage, ToolResult } from "@llamaindex/core/llms";
 
-export class AgentToolCall extends WorkflowEvent<{
+export type AgentToolCall = {
```
leehuwuj, line 5: Should we use zod here? @marcusschiesser

packages/workflow/src/agent/agent-workflow.ts
```diff
@@ -437,135 +455,100 @@ export class AgentWorkflow {
...
-      });
+      };
 
-      await agent.finalize(ctx, agentOutput, ctx.data.memory);
+      await agent.finalize(this.workflow.getStore(), agentOutput);
 
       if (isHandoff) {
-        const nextAgentName = ctx.data.nextAgentName;
+        const nextAgentName = this.workflow.getStore().nextAgentName;
         console.log(
```
leehuwuj, line 485: should we use a verbose config for showing these log?

packages/workflow/src/agent/agent-workflow.ts
```diff
@@ -437,135 +455,100 @@ export class AgentWorkflow {
...
   ) {
     if (this.agents.size === 0) {
       throw new Error("No agents added to workflow");
     }
-    this.setupWorkflowSteps();
-    const contextData: AgentWorkflowContext = params?.context ?? {
+
+    const data: AgentWorkflowData = params?.data ?? {
       userInput: userInput,
       memory: new ChatMemoryBuffer(),
```
leehuwuj, line 553: This would raise an error if the LLM is not set in the settings. How about using the LLM from the root agent or falling back to Settings.LLM if both of them are not set?

leehuwuj, line 553:
```suggestion
      memory: new ChatMemoryBuffer({
        llm: this.agents.get(this.rootAgentName)?.llm ?? Settings.llm,
      }),
```

packages/workflow/src/agent/agent-workflow.ts
marcusschiesser: the handoff is probably not that verbose, i think it's ok

packages/workflow/src/agent/agent-workflow.ts
marcusschiesser: yes, sounds good!

packages/workflow/test/function-agent.test.ts
```diff
@@ -30,32 +28,27 @@ describe("FunctionAgent", () => {
...
 
-    const dummyResult = {
-      data: {
-        toolName: "add",
-        toolKwargs: { x: 2, y: 2 },
-        toolId: "123",
-        toolOutput: { result: "4", isError: false, id: "123" },
-        returnDirect: false,
-      },
-      displayName: "test",
```
leehuwuj, line 41: It looks to me that we don't have `displayName` in the new workflow event type.

examples/agentworkflow/single-agent.ts
```diff
@@ -14,16 +14,12 @@ async function main() {
     verbose: false,
   });
 
-  // Run the agent and keep the context
-  const context = weatherAgent.run("What's the weather like in San Francisco?");
-  const result = await context;
-  console.log(`${JSON.stringify(result, null, 2)}`);
-
-  // Reuse the context from the previous run
-  const caResult = await weatherAgent.run("Compare it with California?", {
```
leehuwuj, line 23: @marcusschiesser Do you have any idea to support remaining context from the previous run for the new code?

packages/workflow/test/function-agent.test.ts
marcusschiesser: payload of an event is just the data, but you can add labels to an event, e.g.:
```
export const myEvent = workflowEvent<MyEventData, "llamaindex-test">(
  {
    debugLabel: "llamaindex-test",
  },
);
```

himself65:
just swa this, thank u

Issue #1899, opened 2025-04-29
Add support for realtime SDK


marcusschiesser:
start with Google: https://googleapis.github.io/js-genai/main/classes/live.Live.html 

here's an example:

```jsx
import { gemini } from "@llamaindex/google";

const AudioChat = () => {
  const [isRecording, setIsRecording] = React.useState(false);
  const [status, setStatus] = React.useState("Disconnected");
  const sessionRef = React.useRef(null);
  const audioContextRef = React.useRef(null);
  const streamRef = React.useRef(null);
  const audioRef = React.useRef(null);

  // Initialize session and AudioContext on mount
  React.useEffect(() => {
    audioContextRef.current = new AudioContext();
    audioRef.current = new Audio();

    const llm = gemini({
      model: "gemini-2.0-flash",
    });
    // or use openai
    const session = llm.live.connect();
    // get all events from the live api
    const events = llm.live.events;

    // Handle incoming audio stream
    (async () => {
      for await (const event of session.stream) {
        if (events.open.include(event)) {
          // Send system prompt
          session.sendMessage({
            content: "You are a helpful assistant. Respond with audio output.",
            role: "system",
          });
          setStatus("Connected");
          sessionRef.current = session;
        } else if (events.audio.include(event)) {
          const audioData = atob(event.delta);
          const uint8Array = Uint8Array.from(audioData, (c) => c.charCodeAt(0));
          const arrayBuffer = uint8Array.buffer;
          const buffer =
            await audioContextRef.current.decodeAudioData(arrayBuffer);
          const source = audioContextRef.current.createBufferSource();
          source.buffer = buffer;
          source.connect(audioContextRef.current.destination);
          source.start();
        } else if (events.error.include(event)) {
          console.error("Error decoding audio:", event.error);
          setStatus("Error");
        } else if (events.close.include(event)) {
          setStatus("Disconnected");
          setIsRecording(false);
        }
      }
    })();

    return () => {
      if (streamRef.current) {
        streamRef.current.getTracks().forEach((track) => track.stop());
      }
      if (audioContextRef.current) {
        audioContextRef.current.close();
      }
      if (sessionRef.current) {
        sessionRef.current.close();
      }
    };
  }, []);

  // Handle microphone recording
  const toggleRecording = async () => {
    if (!isRecording) {
      try {
        const stream = await navigator.mediaDevices.getUserMedia({
          audio: true,
        });
        streamRef.current = stream;
[...]

PR #1905, opened 2025-05-05
feat: Support for google live api

AVtheking:
Add support for google realtime live api #1899


packages/providers/google/src/studio/utils.ts
```diff
@@ -95,42 +95,53 @@ export const mapChatMessagesToGoogleMessages = <
...
+              functionResponse: {
+                name: functionNames[msg.options.toolResult.id] ?? "",
+                response: msg.options.toolResult,
+              },
             },
-          },
-        ],
-      };
+          ],
+        } as Content,
```
marcusschiesser, line 113: what's missing here to be Content?

packages/providers/google/src/event.ts
```diff
@@ -0,0 +1,37 @@
+export type OpenEvent = { type: "open" };
```
marcusschiesser, line 1: we need these events in core - so they can be the same for openai realtime

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,291 @@
...
+  GeminiLiveConfig,
+  GeminiLiveMessage,
+  GeminiLiveMessageDetail,
+} from "./types";
+import {
+  mapBaseToolToGeminiLiveFunctionDeclaration,
+  mapResponseModalityToGeminiLiveResponseModality,
+} from "./utils";
+
+export class GeminiLive {
```
marcusschiesser, line 22: we need an abstract class for this in core - so we can reuse it for openai realtime

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,291 @@
...
+    } else if (this.isAudioMessage(content as GeminiLiveMessageDetail)) {
+      this.sendAudioMessage(content as GeminiLiveMessageDetail, role);
+    } else if (this.isImageMessage(content as GeminiLiveMessageDetail)) {
+      this.sendImageMessage(content as GeminiLiveMessageDetail, role);
+    } else if (this.isVideoMessage(content as GeminiLiveMessageDetail)) {
+      this.sendVideoMessage(content as GeminiLiveMessageDetail, role);
+    }
+  }
+
+  sendMessage(message: GeminiLiveMessage) {
```
marcusschiesser, line 219: this should be the core `Message` type

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  async disconnect() {
+    if (!this.session) {
+      throw new Error("Session not connected");
+    }
+    this.session.close();
+  }
+
+  async connect(config?: GeminiLiveConfig) {
```
marcusschiesser, line 233: the connect parameter GeminiLiveConfig should be defined in core (so it can be the same for openai realtime)

marcusschiesser: the implementations for google live and openai realtime should only differ in the constructor parameters

packages/providers/google/src/base.ts
```diff
@@ -239,6 +243,12 @@ export class Gemini extends ToolCallLLM<GeminiAdditionalChatOptions> {
     this.session = init?.session ?? GeminiSessionStore.get();
     this.#requestOptions = init?.requestOptions ?? undefined;
     this.safetySettings = init?.safetySettings ?? DEFAULT_SAFETY_SETTINGS;
+
+    const apiKey = init?.apiKey ?? getEnv("GOOGLE_API_KEY");
+    if (!apiKey) {
+      throw new Error("Set Google API Key in GOOGLE_API_KEY env variable");
+    }
+    this.live = new GeminiLive(apiKey);
```
marcusschiesser, line 251: only create this object on request

packages/providers/google/src/base.ts
```diff
@@ -239,6 +243,12 @@ export class Gemini extends ToolCallLLM<GeminiAdditionalChatOptions> {
     this.session = init?.session ?? GeminiSessionStore.get();
     this.#requestOptions = init?.requestOptions ?? undefined;
     this.safetySettings = init?.safetySettings ?? DEFAULT_SAFETY_SETTINGS;
+
+    const apiKey = init?.apiKey ?? getEnv("GOOGLE_API_KEY");
```
marcusschiesser, line 247: only do this check for live

marcusschiesser (commented):
@AVtheking looking good and thanks for getting this started! Most comments are about that we need a common interface for Google Live and OpenAI realtime.

Good practice would be to implement openai realtime too, then we can easily see how to build an abstraction

packages/providers/google/src/live.ts
AVtheking: in the core `Message` type it take image url as input but here we need the base64 of the image, also there is not type for the audio and video... one way could be I move this liveMessage type in the core too

packages/providers/google/src/base.ts
AVtheking: I didn't get it how to do it ?

AVtheking: oh do you mean to have a getter on this and create object on calling getter for live ?

packages/providers/google/src/live.ts
marcusschiesser: yes we need a updated `MessageContentDetail` type, that contains a `MessageContentAudioDetail`

AVtheking: But we need to change the image type also, should we do that ?

AVtheking: How about have the new message type for live events ?

marcusschiesser: why should the type be different for live messages? 

how about adding a second image type?

marcusschiesser: (if really needed - what image type is that?)

AVtheking: image
That's how it accepts the data for the image,which is different from the current version.

AVtheking: How about this approach which I used now ?

packages/core/src/llms/type.ts
```diff
@@ -267,3 +267,60 @@ export type ToolOutput = {
...
+export interface LiveConfig {
+  type: string;
+  model?: string;
+  tools?: BaseTool[];
+  responseModality?: ("text" | "audio" | "image")[];
+  systemInstruction?: string;
+  voiceName?: string | GeminiVoiceName;
+}
+
+export type GeminiVoiceName =
```
marcusschiesser, line 318: please no gemini types in core

marcusschiesser: this should be defined in the LLMs constructor

AVtheking: My motive for this to aide developer while chosing for voices.

AVtheking: Likewise we could have for openai too.... developer who would using it would get auto suggestion to choose voice from

marcusschiesser: yes, that's why it's better to set this in the llm constructor - not when calling `connect` - then you can move it to the LLM provider's package

packages/providers/google/src/live.ts
AVtheking: I would add mediaData instead as type here like 
```
export type MessageContentMediaDetail = {
  type: "text" | "audio" | "image" | "video";
  data: string;
  mimeType: string;
  }
```

How about this?

marcusschiesser: hmm, we already have `text`, better to not make this too generic (the `type` should define the type :) )

packages/core/src/llms/type.ts
AVtheking: okay got it

AVtheking:
@marcusschiesser Thanks for the review, I have updated code, PTAL now!

packages/core/src/llms/realtime.ts
```diff
@@ -0,0 +1,43 @@
+import type { ChatMessage, LiveConnectConfig, LiveEvent } from "./type";
+
+export abstract class RealIimeLLM {
```
marcusschiesser, line 3: Google calls it Live OpenAI realtime. We should be consistent and use one of it.
How about LiveLLM?

packages/core/src/llms/type.ts
```diff
@@ -163,6 +163,12 @@ export type MessageContentImageDetail = {
   detail?: "high" | "low" | "auto";
 };
 
+export type MessageContentMediaDetail = {
```
marcusschiesser, line 166: As discussed better one type for each, audio, video and image

packages/providers/anthropic/src/llm.ts
marcusschiesser, line 1: Why is there a change for anthropic?

packages/providers/google/src/events.ts
```diff
@@ -0,0 +1,28 @@
...
+  CloseEvent,
+  ErrorEvent,
+  OpenEvent,
+  SetupCompleteEvent,
+  TextEvent,
+} from "@llamaindex/core/llms";
+
+import type { LiveEvent } from "@llamaindex/core/llms";
+
+export const liveEvents = {
```
marcusschiesser, line 12: Doesn't seem to be provider specific.
How about moving to core?

packages/providers/google/src/studio/utils.ts
marcusschiesser, line 1: Why is this needed?

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  async disconnect() {
+    if (!this.session) {
+      throw new Error("Session not connected");
+    }
+    this.session.close();
+  }
+
+  async connect(config?: LiveConnectConfig) {
```
marcusschiesser, line 219: What happens if you call connect twice? E.g. with your queue.

examples/gemini/gemini-live.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  const llm = gemini({
+    model: GEMINI_MODEL.GEMINI_2_0_FLASH_LIVE,
+    voiceName: "Zephyr",
+  });
+
+  console.log("📡 Connecting to Gemini Live session...");
+
+  const session = await llm.live.connect({
+    responseModality: ["audio"],
```
marcusschiesser, line 61: 1. We already have a Modalitytype and we want to reuse it
2. Maybe move to constructor?
3. What is the default?

examples/gemini/gemini-live.ts
```diff
@@ -0,0 +1,236 @@
...
+  async function sendPcmAudioFile(session: GeminiLive) {
+    try {
+      console.log("🎤 Reading PCM audio file...");
+
+      const filePath = path.join(__dirname, "hello_are_you_there.pcm");
+      console.log(`Reading file from: ${filePath}`);
+
+      const audioBuffer = await fs.readFile(filePath);
+
+      const base64Audio = audioBuffer.toString("base64");
```
marcusschiesser, line 150: We could also allow Buffer - already do this for file type

examples/gemini/gemini-live.ts
```diff
@@ -0,0 +1,236 @@
...
+          console.log("👋 Session closed");
+
+          if (audioResponse && audioChunks.length > 0) {
+            try {
+              const combinedAudioData = Buffer.concat(audioChunks);
+              console.log(
+                `Total audio data: ${combinedAudioData.length} bytes`,
+              );
+
+              const wavHeader = createWavHeader(
```
marcusschiesser, line 115: How about a saveWavFile helper?

packages/providers/google/src/live.ts
marcusschiesser: We probably should support multiple sessions (each one having their own queue)

packages/providers/google/src/studio/utils.ts
AVtheking: As I upgraded the genai library the type was changed in the new one, it require updating it.

packages/providers/anthropic/src/llm.ts
AVtheking: As have added more type in `ChatMessage` and before we were just returning the message in anthropic, openai and vercel by just checking for toolcall type by beliving it would be the other types which are defined.

AVtheking: like before adding more type it was assurity that in the return type it would be image type, but now we have more that's why added check

packages/providers/google/src/live.ts
AVtheking: How would this going to work and look like ?

examples/gemini/gemini-live.ts
AVtheking: 1->Oh cool I didn't knew about it, It is defined in the node.ts file, should I move it somewhere else?
2-> I think it should be in connect parameter as it would be same for openai too
3-> The default is audio type.

examples/gemini/gemini-live.ts
AVtheking: A helper function for saving was files ??

packages/providers/google/src/events.ts
AVtheking: okay we can move it to core.

examples/gemini/gemini-live.ts
marcusschiesser: yes, not in the framework but in this example (seems to me like you're doing this a couple of times)

AVtheking: Oh okay, I didn't focus much on example 😅

examples/gemini/gemini-live.ts
marcusschiesser: yes, it should not support file type - what i mean is that you can use `Buffer` for audio type (same as for `file` type)

packages/providers/google/src/live.ts
marcusschiesser: you can test it in your example, calling `connect` twice seems to create a second session

examples/gemini/gemini-live-tool-call.ts
```diff
@@ -0,0 +1,121 @@
...
+    responseModality: ["text"],
+    // Register our tools with the session
+    tools: [weatherTool, divideNumbers],
+    // Optional system instruction
+    systemInstruction:
+      "You are a helpful assistant that can use tools. When answering questions about weather or divide numbers, always use the appropriate tool.",
+  });
+
+  // Reference to the event helpers
+  const events = liveEvents;
```
marcusschiesser, line 70: you can now directly use `liveEvents`

marcusschiesser: below

examples/gemini/gemini-live-tool-call.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // Process events from the session
+  (async () => {
+    console.log("🎧 Listening for events...");
+
+    for await (const event of session.streamEvents()) {
+      if (events.open.include(event)) {
```
marcusschiesser, line 77:
```suggestion
      if (liveEvents.open.include(event)) {
```

packages/core/src/llms/realtime.ts
```diff
@@ -0,0 +1,43 @@
...
+  protected eventQueue: LiveEvent[] = [];
+  protected eventResolvers: ((value: LiveEvent) => void)[] = [];
```
marcusschiesser, line 5: e.g. you could use a weakmap with session as key and queue as value to have a queue per session

packages/providers/google/src/live.ts
AVtheking: yeah I missed the check here if the session already exists.But should we have support for multiple sessions here ? or can simply make this static?

AVtheking: How about return the same session if called twice or thrice ?

packages/core/src/llms/realtime.ts
AVtheking: How about splitting the class into LiveLLM and LiveLLMSession? This way we do not need to maintain a weakMap ?

packages/providers/google/src/live.ts
AVtheking: How about splitting the class into LiveLLM and LiveLLMSession? This way we do not need to maintain a weakMap ?

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,358 @@
...
+    if (!this.apiKey) {
+      throw new Error("GOOGLE_API_KEY is not set");
+    }
+
+    this.client = new GoogleGenAI({
+      apiKey: this.apiKey,
+    });
+    this.voiceName = init?.voiceName;
+    /* Only 2.0 flash live is supported for live mode */
+    this.model = GEMINI_MODEL.GEMINI_2_0_FLASH_LIVE;
```
marcusschiesser, line 292: throw error if passed model is different?

packages/providers/google/src/live.ts
```diff
@@ -0,0 +1,358 @@
...
+            voiceName: this.voiceName,
+          },
+        },
+      };
+    }
+
+    const geminiLiveSession = new GeminiLiveSession();
+
+    geminiLiveSession.session = await this.client.live.connect({
+      model: "gemini-2.0-flash-live-001",
```
marcusschiesser, line 332: don't hard code model

packages/providers/google/src/live.ts
marcusschiesser: yes, great idea! looks good!

packages/core/src/llms/type.ts
```diff
@@ -163,6 +164,24 @@ export type MessageContentImageDetail = {
   detail?: "high" | "low" | "auto";
 };
 
+export type MessageContentAudioDetail = {
+  type: "audio";
+  data: string | Buffer;
```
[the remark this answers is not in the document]
AVtheking: Yes base64 encoded. What issue with string? if someone wants to provide base64 directly then he can ?

packages/llamaindex/src/indices/vectorStore/index.ts
```diff
@@ -396,8 +396,10 @@ export class VectorStoreIndex extends BaseIndex<IndexDict> {
 /**
  * VectorIndexRetriever retrieves nodes from a VectorIndex.
  */
+type TextAndImageModalityType = Exclude<ModalityType, ModalityType.AUDIO>;
```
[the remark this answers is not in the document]
AVtheking: I am not sure about it though if we should

packages/providers/google/src/live.ts
AVtheking: This only accept this model for now.

marcusschiesser: yes, but we should pass the one from the constructor and throw an error if it's not GEMINI_2_0_FLASH_LIVE (see comment above)

packages/llamaindex/src/indices/vectorStore/index.ts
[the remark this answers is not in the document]
marcusschiesser: i don't see why we should limit this

packages/core/src/llms/type.ts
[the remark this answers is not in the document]
marcusschiesser: yes, it's ok, we just should document what string means here :D 

also better use Uint8Array and not Buffer (as it's nodejs only) - my mistake

packages/llamaindex/src/indices/vectorStore/index.ts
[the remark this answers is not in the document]
AVtheking: Okay will allow it then with no 2  as 0 is for text and 1 is for image I think

packages/core/src/llms/type.ts
```diff
@@ -267,3 +289,47 @@ export type ToolOutput = {
   output: JSONValue;
   isError: boolean;
 };
+
+export type OpenEvent = { type: "open" };
+
+export type AudioEvent = { type: "audio"; delta: string; mimeType: string };
```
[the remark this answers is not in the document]
AVtheking: it is different here. If we wants to use this type then have to change delta to data, would this be okay ? As you have suggested to use delta in the eg?

packages/core/src/llms/type.ts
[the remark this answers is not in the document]
marcusschiesser: yes all your types, i fixed it for file type in https://github.com/run-llama/LlamaIndexTS/pull/1921 will merge that ASAP. you can also use the helper added

packages/core/src/llms/type.ts
```diff
@@ -267,3 +289,47 @@ export type ToolOutput = {
...
+
+export type TextEvent = { type: "text"; content: string };
```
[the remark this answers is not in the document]
AVtheking: same here content would get replaced with text here

packages/core/src/llms/type.ts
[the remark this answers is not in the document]
AVtheking: Thanks, will add doc for the string there

packages/core/src/llms/type.ts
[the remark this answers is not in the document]
marcusschiesser: Why is that a problem? This way we get the same data format out that we put in

packages/core/src/llms/type.ts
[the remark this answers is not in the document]
marcusschiesser: Yes that's ok

packages/core/src/llms/type.ts
```diff
@@ -267,3 +292,9 @@ export type ToolOutput = {
   output: JSONValue;
   isError: boolean;
 };
+
+export interface LiveConnectConfig {
```
marcusschiesser, line 296: nit: better move to live.ts

examples/gemini/gemini-live.ts
```diff
@@ -0,0 +1,221 @@
+import { fs } from "@llamaindex/env";
+import { gemini, GEMINI_MODEL, GeminiLiveSession } from "@llamaindex/google";
+import { liveEvents } from "llamaindex";
+
+import path from "path";
+
+function createWavHeader(
```
marcusschiesser, line 7: nit: examples are easier to read if helpers are at the bottom

marcusschiesser (approved):
great job @AVtheking this looks great!

PR #1911, opened 2025-05-06
feat: qdrant search params

### LINKED ISSUE #1910: Allow for search params when querying using `Qdrant` vector store.
## The Problem

We are using `Qdrant` as our vector store and recently have updated our collection to use quantization in order to reduce memory utilization and improve response times. Here are the relevant docs: https://qdrant.tech/documentation/guides/quantization/ 

When working with quantized vectors, `Qdrant` allows for search parameters to enable `rescoring` and `oversampling`, which improves accuracy while querying quantized vectors. 

Additionally these search parameters allow one to control other index specific parameters while performing `search` or `query` requests. More info on [query](https://api.qdrant.tech/v-1-13-x/api-reference/search/query-points#request.body.params) and [search](https://api.qdrant.tech/v-1-13-x/api-reference/search/points#request.body.params)

## Proposed Solution

Allow for these params to be set when using the `QdrantVectorStore` object and using the `query` method

anubhavrana:
This PR adds functionality to allow users to pass additional search params when using the `query` method of `QdrantVectorStore`.

## Major Changes

- Add `qdrant_search_params` as options when using the `query` method in `QdrantVectorStore`. 
- Update implementation of vector store and vector index to allow for search params.
- Create an example that uses search params when querying a `Qdrant` vector store.
- Tests incorporating search params.

## Minor Changes

- Typo in `Qdrant` example readme.
- Add a line in `docs` to mention setting search parameters.

closes #1910

### DIFF
--- a/packages/core/src/retriever/index.ts
+++ b/packages/core/src/retriever/index.ts
@@ -8,6 +8,7 @@ import { BaseNode, IndexNode, type NodeWithScore, ObjectType } from "../schema";
 export type RetrieveParams = {
   query: MessageContent;
   preFilters?: unknown;
+  qdrant_search_params?: unknown;
 };
 
 export type RetrieveStartEvent = {
--- a/packages/core/src/retriever/index.ts
+++ b/packages/core/src/retriever/index.ts
@@ -8,7 +8,7 @@ import { BaseNode, IndexNode, type NodeWithScore, ObjectType } from "../schema";
 export type RetrieveParams = {
   query: MessageContent;
   preFilters?: unknown;
-  qdrant_search_params?: unknown;
+  customParams?: unknown;
 };
 
 export type RetrieveStartEvent = {
--- a/packages/core/src/vector-store/index.ts
+++ b/packages/core/src/vector-store/index.ts
@@ -72,20 +72,6 @@ export interface MetadataFilters {
   condition?: `${FilterCondition}`; // and, or
 }
 
-export interface SearchParams {
-  hnsw_ef?: number | null;
-  exact?: boolean;
-  quantization?:
-    | Record<string, unknown>
-    | {
-        ignore?: boolean;
-        rescore?: boolean | null;
-        oversampling?: number | null;
-      }
-    | null;
-  indexed_only?: boolean;
-}
-
 export interface MetadataInfo {
   name: string;
   type: string;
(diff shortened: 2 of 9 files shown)


packages/core/src/vector-store/index.ts
```diff
@@ -72,6 +72,20 @@ export interface MetadataFilters {
   condition?: `${FilterCondition}`; // and, or
 }
 
+export interface SearchParams {
```
marcusschiesser, line 75: move to qdrant provider (as it's qdrant specific)

packages/core/src/vector-store/index.ts
```diff
@@ -92,6 +106,7 @@ export interface VectorStoreQuery {
   alpha?: number;
   filters?: MetadataFilters | undefined;
   mmrThreshold?: number;
+  qdrant_search_params?: SearchParams | undefined;
```
marcusschiesser, line 109: as this is provider specific, this must be either unknown or a generic type, e.g.
```suggestion
  customParams?: T | undefined;
```
we're doing this already for ToolCallLLMs, see:
https://github.com/run-llama/LlamaIndexTS/blob/main/packages/providers/anthropic/src/llm.ts

packages/core/src/retriever/index.ts
```diff
@@ -8,6 +8,7 @@ import { BaseNode, IndexNode, type NodeWithScore, ObjectType } from "../schema";
 export type RetrieveParams = {
   query: MessageContent;
   preFilters?: unknown;
+  qdrant_search_params?: unknown;
```
marcusschiesser, line 11: should be also `customParams`

packages/llamaindex/src/indices/vectorStore/index.ts
```diff
@@ -39,6 +39,7 @@ import { storageContextFromDefaults } from "../../storage/StorageContext.js";
 import type {
   BaseVectorStore,
   MetadataFilters,
+  SearchParams,
```
marcusschiesser, line 42: where is this defined?

packages/core/src/retriever/index.ts
marcusschiesser: or maybe `customSearch`?

marcusschiesser:
@anubhavrana please update branch and check failing tests

anubhavrana:
> @anubhavrana please update branch and check failing tests

Branch updated and I did push a commit to fix the failing tests. When you have a minute if you could re trigger the actions/checks that'd be great!

marcusschiesser:
thanks @anubhavrana

PR #1927, opened 2025-05-09
fix: using new workflow with npm by forcing to use the current versio…

marcusschiesser:
…n of core in llamaindex (and not the one from the old workflow)


marcusschiesser:
re-releasing core worked with npm now, closing

PR #1980, opened 2025-05-26
feat: add support for interrupted and other server content event in live api

AVtheking:


packages/core/src/llms/live.ts
```diff
@@ -17,13 +17,22 @@ export type CloseEvent = { type: "close" };
...
+
 export type LiveEvent =
   | OpenEvent
   | AudioEvent
   | TextEvent
   | ErrorEvent
   | CloseEvent
-  | SetupCompleteEvent;
+  | SetupCompleteEvent
+  | InterruptedEvent
```
marcusschiesser, line 33: If we add those events to core, we need to know how these events would mal for openai.
Is there even a mapping? If not, maybe better wrap provider specific events in one parent event?

AVtheking: so you mean moving it in the provider llms ?

AVtheking: ok will add comments here, also not sure about the open ai events here, it could happen or not, how about wrapping it when integrating the open ai realtime ?

PR #1993, opened 2025-05-30
 fix: ignore empty parts for gemini which confuses agent

parhammmm:
Gemini was returning empty parts consecutively before a tool call, most likely because it was accumulating itself and yielding empty parts

That caused tool calls failing in agent, this fixes the issue


marcusschiesser:
Thanks a lot!

Issue #1846, opened 2025-04-15
How to use the python version of xml reader


byteHulk:
I'm trying to build a rag quiz based on an xml file, how can I call the python version of xml reader, or is there an equivalent?

ref: https://docs.llamaindex.ai/en/v0.10.33/api_reference/readers/file/#llama_index.readers.file.XMLReader

marcusschiesser:
We don't have an XMLReader at the moment, but it's a great idea. We're also open for contributions, need to be added to https://github.com/run-llama/LlamaIndexTS/tree/main/packages/readers

byteHulk:
I'll take care of that.

PR #2031, opened 2025-06-21
feat(readers): add XMLReader for parsing XML files (#1846)

jerinthomascarmel:
1. Added XMLReader class having split level 
2. Added example code for reference.


marcusschiesser:
Thanks @jerinthomascarmel looks good, please just fix the lint error

PR #2032, opened 2025-06-23
feat: Add support for StreamableHTTP MCP Client

leehuwuj:


.changeset/pink-donuts-marry.md
marcusschiesser, line 1: missing docs update

packages/tools/src/tools/mcp.ts
```diff
@@ -60,10 +78,27 @@ class MCPClient {
...
     if ("url" in options) {
-      this.transport = new SSEClientTransport(
-        new URL(options.url),
-        options as SSEClientTransportOptions,
-      );
+      // If useSEE is not provided, default to true
+      // to avoid breaking changes
+      if (options.useSSE === undefined) {
+        options.useSSE = true;
+      }
```
marcusschiesser, line 85: this is weird, why not adding `noSSE` with `false` `undefined` as default?

packages/tools/src/tools/mcp.ts
```diff
@@ -60,10 +78,27 @@ class MCPClient {
...
+      if (options.useSSE) {
+        // Show deprecation warning
+        console.warn(
+          "SSE transport will be soon deprecated. " +
+            "Please use StreamableHTTPClientTransport instead.",
+        );
+        this.transport = new SSEClientTransport(
+          new URL(options.url),
+          options as SSEClientTransportOptions,
+        );
```
marcusschiesser, line 95: if we useSSE this will always use the old `SSEClientTransport` code - why not removing it?

PR #2033, opened 2025-06-23
feat(readers): add ExcelReader for parsing Excel files (run-llama#1959)

jerinthomascarmel:


marcusschiesser:
@jerinthomascarmel please have a look the type-check is failing

Issue #2019, opened 2025-06-13
SentenceSplitter is breaking "Sr. Lastname" to "Sr.Lastname"


BrodaNoel:
**Describe the bug**

While using the code pasted below, when I get the chunks from a big text, these texts: `correspondiente al Dr. Pablo` are translated to `correspondiente al Dr.Pablo` in the resulting chunk (ir removes the space between "Dr." and "Pablo").

Something important is that: If the full text only returns 1 chunk (a small text), this issue doesn't happens. It only happens when the results are more than 1 chunk.

You can also see issues like this: `($200). Blabla` being joined like this: `($200).Blabla`

**To Reproduce**
Code to reproduce the behavior:

```ts
import { SentenceSplitter } from '@llamaindex/core/node-parser';

function getChunks(text) {
  const splitter = new SentenceSplitter({
    chunkSize: 700,
    chunkOverlap: 100,
    paragraphSeparator: '\n\n',
  });
  return splitter.splitText(text);
}

[...]

PR #2041, opened 2025-06-25
Improve tests for medium and long texts

BrodaNoel:
The idea of this PR is to add tests for testing the issue reported on https://github.com/run-llama/LlamaIndexTS/issues/2019

The amount of tokens shown in the code, were gotten from https://platform.openai.com/tokenizer

### DIFF
--- a/packages/core/tests/node-parser/text-splitter.test.ts
+++ b/packages/core/tests/node-parser/text-splitter.test.ts
@@ -75,19 +75,39 @@ describe("sentence splitter", () => {
     expect(splits).toEqual(["This is a sentence. This is another sentence."]);
   });
 
-  test("overall split long text", () => {
+  test("overall split medium-size text", () => {
     const sentenceSplitter = new SentenceSplitter({
-      chunkSize: 10,
+      chunkSize: 12,
       chunkOverlap: 0,
     });
     const splits = sentenceSplitter.splitText(
+      // 24 tokens
       "The first short sentence. The first long long long sentence. The second short sentence. The second long long long sentence.",
     );
     expect(splits).toEqual([
-      "The first short sentence.",
-      "The first long long long sentence.",
-      "The second short sentence.",
-      "The second long long long sentence.",
+      // 12 tokens
+      "The first short sentence. The first long long long sentence.",
+      // 12 tokens
+      "The second short sentence. The second long long long sentence.",
+    ]);
+  });
+
+  test("overall split long-size text", () => {
+    const sentenceSplitter = new SentenceSplitter({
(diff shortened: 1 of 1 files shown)


BrodaNoel:
@marcusschiesser I have found something that I am not sure if I understand well.

Check the long-text test. It has a configuration of 700 tokens and 100 of overlap. While the full text has 671 tokens.

I have copy&paste the chunks previously created by the library, which are published on https://github.com/run-llama/LlamaIndexTS/issues/2019

As you can see, it split everything on 2 chunks: one of 545 tokens and other with 126 tokens.

Now, I wonder if those values are actually OK. If the config is 700 + 100 overlap, shouldn't the first chunk has 700? Shouldn't everything be included just in the first chunk?


On the other hand, I have another question which the answer will help me to add even better and more robusts texts:

If the chunk size is 700, and overlap is 100, and there is a text with 1000 tokens, how should the chunks be? Let me know if I am right with my idea:

- The first chunk should be 600 tokens + 50 right-overlap = 650 total tokens
- The second chunk should be 50 left-overlap + 400 tokens + 50 right-overlap = 500 total tokens

Am I right?
700 with 100 overlap means 600 as "base", plus 50 on left/top and 50 on right/bottom. Or am I wrong?

BrodaNoel:
@marcusschiesser btw, check this unit test failing:

```diff
 FAIL |2|  node-parser/text-splitter.test.ts > sentence splitter > overall split medium-size text
AssertionError: expected [ …(2) ] to deeply equal [ …(2) ]

- Expected
+ Received

  Array [
-   "The first short sentence. The first long long long sentence.",
-   "The second short sentence. The second long long long sentence.",
+   "The first short sentence.The first long long long sentence.",
+   "The second short sentence.The second long long long sentence.",
  ]
```

It is clearly joining (`join('')`) sentences with no spaces.
Not sure where to find that. And for some reason, unit tests are not running on my computer (I have an old 2015 macbook), so I can't find a fix on my end.

BrodaNoel:
ChatGPT is telling me that 700 + 100 chunks works like this:

Chunk 1: characters 0 to 699  
Chunk 2: characters 600 to 1299  
Chunk 3: characters 1200 to 1899  

Is that correct?

BrodaNoel:
@marcusschiesser  Any plan to fix it soon?

I see your PR, and I would suggest to add the test that I fixed. The one with 12 tokens, which shows that at the code is removing the space between the words and the "."

marcusschiesser:
Fixed in https://github.com/run-llama/LlamaIndexTS/pull/2046 - Thanks for finding this @BrodaNoel !

PR #2072, opened 2025-07-08
feat: adding structured output support for anthropic

AstraBert:
Adding support for structured output (via Zod types) in anthropic!

Test it out with this script:

```typescript
import { Anthropic } from "./src";
import { z } from "zod";

const responseSchema = z.object({
    title: z.string(),
    author: z.string(),
    year: z.number(),
})

const llm = new Anthropic({"model": "claude-4-0-sonnet"})

async function main() {
    const response = await llm.chat({
    messages: [
      {
        role: 'system',
        content: `You are a book expert. Your task is, given a user message, extract the title, author and publication year of the book and output them in JSON format.`,
      },
      {
        role: 'user',
        content: `I have been reading La Divina Commedia by Dante Alighieri, published in 1321, which tells the story of a guy who goes through Hell, Purgatory and Heaven just to meet his beloved ex-girlfriend.`,
      },
    ],
    responseFormat: responseSchema,
  })
  return response.message.content
}

const antResponse = await main()
console.log(antResponse)
```

### DIFF
--- a/packages/providers/anthropic/package.json
+++ b/packages/providers/anthropic/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@llamaindex/anthropic",
   "description": "Anthropic Adapter for LlamaIndex",
-  "version": "0.3.15",
+  "version": "0.3.16",
   "type": "module",
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
@@ -31,9 +31,9 @@
     "test": "vitest run"
   },
   "devDependencies": {
-    "vitest": "^2.1.5",
     "@llamaindex/core": "workspace:*",
-    "@llamaindex/env": "workspace:*"
+    "@llamaindex/env": "workspace:*",
+    "vitest": "^2.1.5"
   },
   "peerDependencies": {
     "@llamaindex/core": "workspace:*",
@@ -41,6 +41,8 @@
   },
   "dependencies": {
     "@anthropic-ai/sdk": "0.37.0",
-    "remeda": "^2.17.3"
+    "remeda": "^2.17.3",
+    "zod": "^3.25.67",
+    "zod-to-json-schema": "^3.24.6"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/providers/anthropic/src/llm.ts
```diff
@@ -30,6 +30,8 @@ import { ToolCallLLM } from "@llamaindex/core/llms";
 import { extractText } from "@llamaindex/core/utils";
 import { getEnv } from "@llamaindex/env";
 import { isDeepEqual } from "remeda";
+import { z } from "zod";
+import { zodToJsonSchema } from "zod-to-json-schema";
```
himself65, line 34: use zod v4, it's to json schema built-in, you can refer to my another PR https://github.com/anthropics/dxt/pull/39/files#diff-00e8b7e9ce9d95fa021ecce03324e02fe9b946eaeaf0c2f779c0c5d9278ef051

himself65 (changes requested):

himself65 (approved):

AstraBert:
@himself65 I feel like with zod/v4 the code is more elegant and requires less things:  think we could still bump the zod version (not sure how to do it properly) but from what I can see on npm there shouldn't have been big changes since our version (3.25.67) only differs for patches and not for minor/majors from the latest one (3.25.76)

packages/providers/anthropic/package.json
```diff
@@ -31,16 +31,17 @@
...
   },
   "peerDependencies": {
     "@llamaindex/core": "workspace:*",
     "@llamaindex/env": "workspace:*"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "0.37.0",
-    "remeda": "^2.17.3"
+    "remeda": "^2.17.3",
+    "zod": "^3.25.67"
```
marcusschiesser, line 45: note that zod must be a `peerDependency` as currently we're having it as dependency in the core package. So if we upgrade to zod/4 we must do it in core first. 
Doing so we probably should make it an optional peerDependency in core too as users usually bring their own

PR #2085, opened 2025-07-11
feat: add supportToolCall getter to DeepSeekLLM class

phields:


phields:
> should already have supportToolCall from the OpenAI class that is extended - which problem are you facing?

OpenAI class determines whether the model supports ToolCall by the model name. The DeepSeek model name is not in the supported list.
```
        agentLLM = new DeepSeekLLM({
          apiKey: process.env.DEEPSEEK_API_KEY,
          model: "deepseek-chat",
          temperature: 1,
        });

```
got error:
Error: Error: FunctionAgent requires an LLM that supports tool calls

Issue #1840, opened 2025-04-13
Stream each part of a tool call


r3rer3:
At the moment, it is only possible to know when an LLM has finished streaming a tool call, but it is not possible to know when it has started streaming a tool call (e.g. Vercel AI SDK has a `partial-call` state so the user know the tool call is in progress). Knowing it would be great to improve UI.

marcusschiesser:
Good idea. What UI SDK are you using to consume the events?

marcusschiesser:
If you'll send a snippet we can optimize the usage for your case

marcusschiesser:
Got it, let's divide this in two parts:
1. Updating Agent Workflows
2. Updating the LLM classes
Can you send a PR with your changes to `streamChat` from Anthropic? Seems like a good starting point for 2.

r3rer3:
I will send a PR tomorrow! :)

r3rer3:
Hi! Sorry, I haven't had the time yet to create the PR because of work :/ I will open as soon as possible. Sorry for the delay

PR #2100, opened 2025-07-14
feat(anthropic): stream partial tool calls

r3rer3:
Stream partial calls. For context, this is already implemented for the openai provider.

See #1840


marcusschiesser (approved):
Looking good, did you already test this?

himself65 (approved):

PR #2109, opened 2025-07-17
Update SupabaseVectorStore.ts to fix score calculating error

jerrycheng13:
The match_documents() function already returns cosine similarity score if embeddings and query_embedding are normalized https://ts.llamaindex.ai/docs/llamaindex/modules/data/stores/vector_stores/supabase

While the current code returns 1 - item.similarity || 0 which is wrong


marcusschiesser (approved):
Thanks @jerrycheng13

PR #2114, opened 2025-07-18
fix: ensure correct message content in agent workflow

thucpn:


packages/workflow/src/agent/agent-workflow.ts
```diff
@@ -502,15 +506,16 @@ export class AgentWorkflow implements Workflow {
     if (directResult) {
       const isHandoff = directResult.toolName === "handOff";
 
-      const output =
-        typeof directResult.toolOutput.result === "string"
-          ? directResult.toolOutput.result
-          : JSON.stringify(directResult.toolOutput.result);
+      // @ts-expect-error - raw is not typed in ToolResult (should add)
```
marcusschiesser, line 509: What if the tool output is a string? Messagecontent is also just a string

marcusschiesser (approved):

PR #2115, opened 2025-07-18
feat: adding responseFormat to llm.exec

AstraBert:
As per title, but it is untested.

### DIFF
--- a/packages/core/src/llms/base.ts
+++ b/packages/core/src/llms/base.ts
@@ -1,3 +1,5 @@
+import { z } from "zod";
+import { tool } from "../tools/";
 import { extractText } from "../utils/llms";
 import { streamConverter } from "../utils/stream";
 import { callTool, getToolCallsFromResponse } from "./tool-call";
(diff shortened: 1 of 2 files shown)


packages/core/src/llms/base.ts
```diff
@@ -96,6 +98,31 @@ export abstract class BaseLLM<
     | ExecResponse<AdditionalMessageOptions>
     | ExecStreamResponse<AdditionalMessageOptions>
   > {
+    const responseFormat = params.responseFormat;
+    if (responseFormat && isZodSchema(responseFormat)) {
```
marcusschiesser, line 102: this is just adding the tool to generate a JSON object - how do we ensure that the `content` field of the returned message is the JSON object?

`content` is of type `string | MessageContentDetail[]`, so as a user i would assume `content` will be a `stringified` JSON with the `responseFormat` structure. 

later (another PR) we need to use the LLM's native JSON output if available

packages/core/src/llms/base.ts
```diff
@@ -1,3 +1,4 @@
+import { tool } from "../tools/";
```
marcusschiesser, line 1: add example to use in `/examples`

marcusschiesser: Thanks example looks great

Issue #2124, opened 2025-07-21
Documentation for Bedrock is outdated


tersimdev:
https://ts.llamaindex.ai/docs/llamaindex/modules/models/llms/bedrock
The documentation is outdated here, did not realise until recently that there was @llamaindex/aws available.

marcusschiesser:
@tersimdev greet catch, want to send a PR?

tersimdev:
sure, can i check if these comments are still true for the meta models? @marcusschiesser 

<img width="1180" height="200" alt="Image" src="https://github.com/user-attachments/assets/879dd640-d6e6-4613-8cbd-6e527089b17a" />

PR #2129, opened 2025-07-22
docs: updated bedrock import and supported models

tersimdev:
Changed @llamaindex/community to @llamaindex/aws, and updated model and inference models list. For #2124

**Question**: are the comments for the meta models and claude 3 opus still valid? Not sure what "only available via inference endpoints (see below)" means.


logan-markewich:
Thanks for tackling this @tersimdev -- someone raised this issue in discord  today and youve already fixed it 🫡

Issue #2071, opened 2025-07-08
Remove: Hardcoded console.log


marcusschiesser:
### Discussed in https://github.com/run-llama/LlamaIndexTS/discussions/2070

<div type='discussions-op-text'>

<sup>Originally posted by **Falco-Boehnke** July  8, 2025</sup>
Hi all! We noticed this hardcoded console.log and were wondering: What is the thought behind it? Why do we need it to be hardcoded here? And especially, could this not be removed, or otherwise be made loggable, so it's not a forced log?

https://github.com/run-llama/LlamaIndexTS/blob/f631d4f7d68402e7acfaa74851d341767f4532b5/packages/workflow/src/agent/agent-workflow.ts#L525</div>

marcusschiesser:
@Falco-Boehnke - yes that shouldn't be hard coded. We actually have a type for this: https://github.com/run-llama/LlamaIndexTS/blob/main/packages/env/src/logger/index.ts#L1

Would you like to send a PR that uses this type?

tersimdev:
sure, i can add another pr for that

PR #2130, opened 2025-07-22
chore: add logger as param to AgentWorkflow constructor

tersimdev:
Changes as suggested in #2071 

Additionally, simplified the intialization of the logger object. The logic should be equivalent since `this.verbose` is set in a line above.

```ts
//from
this.logger =
      verbose === false
        ? emptyLogger
        : verbose || this.verbose
          ? consoleLogger
          : emptyLogger;

//to
this.logger = logger ?? (this.verbose ? consoleLogger : emptyLogger);
```

### DIFF
--- a/.changeset/brave-snakes-grin.md
+++ b/.changeset/brave-snakes-grin.md
@@ -1,0 +1,5 @@
+---
+"@llamaindex/workflow": patch
+---
+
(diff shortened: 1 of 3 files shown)


marcusschiesser (approved):
thanks @tersimdev

PR #2134, opened 2025-07-23
chore: Deprecate community package and link to AWS package

marcusschiesser:


logan-markewich (commented):
hehe, are you adding these files back just so we can deprecate? (I had a feeling that was the only way)

marcusschiesser:
@logan-markewich hehe, yes didn't have a better idea

PR #2142, opened 2025-07-30
chore: deprecate @llamaindex/cloud in favor of llama-cloud-services

AstraBert:
With this PR we deprecate @llamaindex/cloud and we migrate the exports of the package to llama-cloud-services.

(This PR will be able to merge only once `llama-cloud-services` has been merged into main from https://github.com/run-llama/llama_cloud_services/pull/811)


packages/cloud/package.json
```diff
@@ -78,20 +78,7 @@
...
-  },
-  "peerDependencies": {
-    "@llama-flow/core": "^0.4.1",
-    "@llamaindex/core": "workspace:*",
-    "@llamaindex/env": "workspace:*"
-  },
   "dependencies": {
-    "p-retry": "^6.2.1",
-    "zod": "^3.25.76"
+    "llama-cloud-services": "^0.1.0"
```
logan-markewich, line 82: I think this can just be "*" right?

AstraBert: Makes sense!
Will do tomorrow ;)

PR #2167, opened 2025-08-12
feat: add responseFormat to llm.exec

AstraBert:
With this PR we add `responseFormat` to `llm.exec` to have a way of generating structured output responses with a tool call


packages/core/src/tools/function-tool.ts
```diff
@@ -94,7 +95,7 @@ export class FunctionTool<
     ) {
       const { execute, parameters, ...restConfig } = fnOrConfig;
 
-      if (parameters instanceof z.ZodSchema) {
+      if (isZodSchema(parameters)) {
```
marcusschiesser, line 98: @AstraBert great catch!

marcusschiesser (approved):
great job @AstraBert

Issue #2171, opened 2025-08-16
Bedrock LLM complete is using the wrong region


ducduyn31:
**Describe the bug**
When using llm.complete with a Bedrock model, it is using the wrong region. When specify region using `AWS_REGION=us-east-1`, it didn't pick up, and produce the follow error:

```
ValidationException: Invocation of model ID anthropic.claude-sonnet-4-20250514-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.
```

**To Reproduce**
Code to reproduce the behavior:

```ts
import { Bedrock, INFERENCE_BEDROCK_MODELS } from "@llamaindex/aws";

const llm = new Bedrock({
  model: INFERENCE_BEDROCK_MODELS.US_ANTHROPIC_CLAUDE_4_SONNET,
});

console.log(process.env.AWS_REGION); // us-east-1

const response = await llm.complete({
  prompt: "Hello, how are you?"
});

console.log(response.text);
```

**Expected behavior**
Response should be printed without any errors as with using llm.chat

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. macOS, Linux]
- JS Runtime / Framework / Bundler (select all applicable)
- [x] Node.js
- [ ] Deno
- [x] Bun
- [ ] Next.js
- [ ] ESBuild
- [ ] Rollup
- [ ] Webpack
- [ ] Turbopack
- [ ] Vite
- [ ] Waku
- [ ] Edge Runtime
- [ ] AWS Lambda
- [ ] Cloudflare Worker
- [ ] Others (please elaborate on this)
- Version [e.g. 22] 22

**Additional context**

PR #2172, opened 2025-08-16
fix: bedrock complete using actual modelId

ducduyn31:
Fix issue #2171 by adding the `actualModel` as an argument to the AWS command. A similar implementation exists in the `streamChat` and `nonStreamChat` methods, but it was missing in `complete` method. This resulted in the following exception:
```
ValidationException: Invocation of model ID anthropic.claude-sonnet-4-20250514-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.
```

After applying this fix, it now works as expected.

Here is the code I used to test, given that we setup necessary environment variables.
```ts
import { Bedrock, INFERENCE_BEDROCK_MODELS } from "@llamaindex/aws";

const llm = new Bedrock({
  model: INFERENCE_BEDROCK_MODELS.US_ANTHROPIC_CLAUDE_4_SONNET,
});

console.log(process.env.AWS_REGION); // us-east-1

const response = await llm.complete({
  prompt: "Hello, how are you?"
});

console.log(response.text);
```

Here is the result:
```
❯ pnpm bun bedrock-demo.ts
us-east-1
Hello! I'm doing well, thank you for asking. I'm here and ready to help with whatever you'd like to discuss or work on. How are you doing today?
```

### DIFF
--- a/.changeset/cyan-eagles-lie.md
+++ b/.changeset/cyan-eagles-lie.md
@@ -1,5 +1,5 @@
 ---
-"@llamaindex/community": minor
+"@llamaindex/community": patch
 ---
 
 fix: invoke complete command using the actual modelId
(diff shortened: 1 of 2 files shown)


marcusschiesser:
Thanks @ducduyn31

Issue #2020, opened 2025-06-14
progress callback


jeremybmerrill:
Would you be interested in a PR that would modify the [existing logProgress option](https://github.com/run-llama/LlamaIndexTS/blob/8122c7245e80f9a0f6f635149345ab8ff9823abb/packages/core/src/embeddings/base.ts#L18-L20) on `BaseEmbedding` to take a callback that gets called with the progress so far and the queue length.

This is useful for desktop apps where users might be shown a progress bar in the UI.

I see two ways to do it:
 
1. new progressCallback arg, deprecate logProgress: maintain logProgress (for now) its functionality (since the `logProgress: true` option would be superseded by `progressCallback: (i, queue_length) => console.log(`getting embedding progress: ${i} / ${queue.length}`);`
2. change the type of logProgress to be a boolean or a function; if it's `true`, use the existing console.log.

I've used #1 (minus the deprecation of logProgress) here: https://github.com/jeremybmerrill/meaningfully/blob/feature/progress-bar/src/main/services/progressOpenAIEmbedding.ts

let me know if you want this, and I will send a PR (with tests).

marcusschiesser:
great idea. we have in env package the `Logger` type: https://github.com/run-llama/LlamaIndexTS/blob/lee%2Fagent-step-handler/packages/env/src/logger/index.ts - maybe you use that?

PR #2187, opened 2025-09-04
vectorStoreIndex has new option progressCallback

jeremybmerrill:
In #2098, I added a `progressCallback` option to `BaseEmbedding`. Unfortunately this doesn't work as I expected. The transform function defined in BaseEmbedding's constructor is defined when the inheriting Embedding object is created, and there's no way to access it after the fact -- nor is there a way to provide the `progressCallback` at creation time. 

The purpose of this is to allow GUI clients of LlamaIndexTS to create progress bars.

For me to get this to work, I have modified VectorStoreIndex to pass that option through (as `options.progressCallback`) from `init()` to `buildIndexFromNodes()` to `insertNodes()` to `getNodeEmbeddingResults()` to the `embedModel()` call of the embedding model's transform function.

I've included a passing test.

see #2098 (and #2020):

### DIFF
--- a/.changeset/brown-cherries-brake.md
+++ b/.changeset/brown-cherries-brake.md
@@ -1,0 +1,6 @@
+---
+"llamaindex": patch
+"@llamaindex/llamaindex-test": patch
+---
+
+feat: vectorStoreIndex has new option progressCallback


marcusschiesser:
Thanks @jeremybmerrill looks good!

Issue #2182, opened 2025-08-27
Cannot set anthropic temperature to 0


atevington-centralreach:
**Describe the bug**
Setting temp to 0 actually sets temp to default of 1

https://github.com/run-llama/LlamaIndexTS/blob/main/packages/providers/anthropic/src/llm.ts#L182

**Expected behavior**
Temp should be 0

naaa760:
HI @atevington-centralreach @marcusschiesser 
 temperature bug has been resolved - setting `temperature: 0` now properly preserves the zero value instead of defaulting to 1.

PR #2190, opened 2025-09-07
fix: anthropic temperature parameter not respecting value 0

naaa760:
### PR Description:

fixes:   #2182

**Problem:**
Setting `temperature: 0` in Anthropic LLM constructor was incorrectly defaulting to `1` instead of preserving the `0` value.

**Root Cause:**
The nullish coalescing operator (`??`) in the constructor was not properly handling the case where temperature is explicitly set to `0`.

**Solution:**
Changed the temperature initialization logic from:
```typescript
this.temperature = init?.temperature ?? 1;
```

To:
```typescript
this.temperature = init?.temperature != null ? init.temperature : 1;
```

**Impact:**
-  Temperature `0` now works correctly for deterministic responses
- All other valid temperature values (0.1, 0.5, etc.) are preserved
- Invalid/null values still default to `1`
-  No breaking changes to existing functionality

### DIFF
--- a/.changeset/light-spiders-pump.md
+++ b/.changeset/light-spiders-pump.md
@@ -1,0 +1,5 @@
+---
+"@llamaindex/anthropic": patch
+---
+
+fix: anthropic temperature parameter not respecting value 0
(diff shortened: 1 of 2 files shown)


marcusschiesser:
Thanks @naaa760 !

PR #2191, opened 2025-09-09
chore: add latest google models

solhosty:
Adds latest stabel 2.5 models from: https://ai.google.dev/gemini-api/docs/models


marcusschiesser:
Thanks @solhosty

PR #2212, opened 2025-10-02
fix: message delta visibility for anthropic fixed

Kannav02:
This PR aims to fix the message_delta not taken into consideration for Anthropic's LLMs

It solves the following issues 
- Anthropic's `message_delta` is emiting now, so the usage meta data isn't dropped and is actually reaching the user
- The final streamed chunk was always `undefined`, this is fixed as the lastChunk is being assigned the value of the chunk that is read last in the loop, even though it might have empty content, their is usage data which can be passed on.
- Included `raw` as one of the fields in AgentWorkflow, so callers of `agent.run` recieve the usage when it is sent by the `takeStep` function

The following tests were ran and they were passing for the same
- `pnpm --filter @llamaindex/workflow test`
- `pnpm --filter @llamaindex/anthropic test`

### DIFF
--- a/packages/workflow/src/agent/function-agent.ts
+++ b/packages/workflow/src/agent/function-agent.ts
@@ -197,6 +197,7 @@ export class FunctionAgent implements BaseWorkflowAgent {
     let lastChunk: ChatResponseChunk | undefined;
     const toolCalls: Map<string, AgentToolCall> = new Map();
     for await (const chunk of responseStream) {
+      lastChunk = chunk;
       response += chunk.delta;
       ctx.sendEvent(
         agentStreamEvent.with({
--- a/packages/providers/anthropic/src/llm.ts
+++ b/packages/providers/anthropic/src/llm.ts
@@ -604,6 +604,15 @@ export class Anthropic extends ToolCallLLM<
           ? part.delta.signature
           : undefined;
 
+      if (part.type === "message_delta") {
+        yield {
+          raw: part,
+          delta: "",
+          options: {},
+        };
+        continue;
+      }
+
       if (
         part.type === "content_block_start" &&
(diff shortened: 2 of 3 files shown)


packages/workflow/src/agent/function-agent.ts
```diff
@@ -197,6 +197,7 @@ export class FunctionAgent implements BaseWorkflowAgent {
     let lastChunk: ChatResponseChunk | undefined;
     const toolCalls: Map<string, AgentToolCall> = new Map();
     for await (const chunk of responseStream) {
+      lastChunk = chunk;
```
marcusschiesser, line 200: Why is this needed?

.changeset/wild-bananas-sip.md
```diff
@@ -0,0 +1,6 @@
+---
+"@llamaindex/workflow": patch
+"@llamaindex/anthropic": patch
+---
+
+fix: message delta visibility for anthropic
```
marcusschiesser, line 6: We need a change set what we are fixing for workflow in this PR

packages/workflow/src/agent/function-agent.ts
Kannav02: because in the following code snippet 
```
    return {
      response: message,
      toolCalls: Array.from(toolCalls.values()),
      raw: lastChunk?.raw,
      currentAgentName: this.name,
    };
```

the raw chunk that is being returned takes the value from lastChunk.raw, in the earlier code, lastChunk was beind defined, but never assigned a value, thus lastChunk always had the value as `undefined`, so thats why i assigned it tha lastChunk that we see in the loop, sometime it could even be empty text, but even if its empty, it still sends metadata

PR #2234, opened 2025-11-04
fix: first tool call chunk without arguments

fengkx:
Clipboard_Screenshot_1762257311
In some openai compatiable model, The first tool call chunk that contain an id field may not have a arguments field (in next chunk). 

In this case, `currentToolCall.input` is `undefined`, which make the string concated into `undefined{....`. It will break JSON parse.

We should check for this before concat the arguments string


marcusschiesser:
Thanks @fengkx

PR #2237, opened 2025-11-06
feat: GeminiEmbedding rate-limit handling

jeremybmerrill:
Google Gemini has rate-limits on embeddings -- 3000 vectors per minute. Right now, if we're generating embeddings via LlamaIndex (e.g. with `VectorStoreIndex.init` or similar methods), if we hit the rate limits, it just errors out, with no ability to wait or restart.

This proposed addition would wait 5s and retry, up to 20 times any embed call that fails with a rate limit error. That up-to-100s wait gets you out of the per-minute limit -- so that the requests-per-minute limit is seamlessly handled by applications using LlamaIndex.

I've added an example file that fails with the existing main branch of llamaindexts, but succeeds with this patch. You can run it with `ts-node examples/models/gemini/embedding_ratelimits.ts`. (I didn't really know how write a proper jest test for this, without actually hitting the Gemini API and without faking the way that the Google AI library throws errors. Rather than tightly couple the test to the current behavior of the Google AI library, I wrote an example that does hit the Gemini API.)


jeremybmerrill:
> This PR should not include the postgres change

Aw crap, how'd I do that.

I've re-pushed a version of this branch that doesn't include the Postgres change.

